PR #6, opened 2025-02-04
Proof of concept based on Rotonda.

ximon18:
This draft PR contains a proof-of-concept "bump-in-the-wire" DNSSEC signing server that can receive and publish zones via XFR.

It is based on a modified branch of `domain` to address some issues found with `Service` trait bound matching and to address a short coming in the `ZoneUpdater`.

It is not meant to be fast or memory efficient or even a correct DNS server (as domain lacks various things needed for that such as setting the AA flag correctly on responses, support for various record types is missing, and support for DNSSEC signed responses. It is also not meant to be clean code without bugs.

Known issues include blocking the HTTP API while signing, no current hot-reconfiguration support (though the underlying Rotonda base supports it), no Key Manager component yet, no proper authentication on APIs, no metrics (though the underlying Rotonda base supports it), no use of the Rotonda base "Status Reporter" system, and most importantly: NO SUPPORT FOR INCREMENTAL SIGNING, and DUMB SUPPORT FOR SIGNING OF `Zone`s VIA COLLECTION FIRST INTO A VEC, limited docs (especially around the signer configuration), confusing handling of misspelled config file entries (by silently ignoring them), poor config file syntax, some things in config maybe should never be in config as they are dynamic, perhaps it would make sense to support catalog zones, no command line interface, no web UI, no JSON API outputs (instead they currently output poor mans plain text), no key life cycle for DNSSEC (due to the missing Key Manager component), etc.

Instead it is meant as a quickly put together demonstration of the design ideas that were discussed internally about a component based model with events flowing to a Central Command and commands in turn being issued to components. It doesn't strictly implement the async fn event and command model that was discussed, as only events are communicated via async fn calls while commands are issued via multiple-producer single-consumer channel.

I probably forgot a lot of things... ;-)

See `README.md` and `nameshed.conf` for more information.


ximon18:
Agreed in the team to merge this to `main`. This is not a reflection of readiness or quality, it is simpler to make it easier for further work to be done by the team.

PR #10, opened 2025-06-04
Add key manager skeleton

ximon18:


src/units/zone_signer.rs
```diff
@@ -1579,3 +1528,147 @@ impl domain::dnssec::sign::records::Sorter for MultiThreadedSorter {
...
+    Ok((key_id, algorithm, flags, host_and_port))
+}
+
+// Like domain::crypto::sign::KeyPair but omits Ring which is not thread safe.
+#[derive(Debug)]
+enum ThreadSafeKeyPair {
+    OpenSSL(domain::crypto::openssl::sign::KeyPair),
+
+    Kmip(domain::crypto::kmip::sign::KeyPair),
+}
```
bal-e, line 1651: It appears that Ring support can be restored quite easily, by making small changes in `domain`.  I'll add a PR for that; can we add a TODO comment here to note that we will be able to remove this type soon?

bal-e (approved):
All the logic here seems sound.

PR #13, opened 2025-07-09
Set up a new configuration system

bal-e:
This implements the first step in the UI for Nameshed, the configuration mechanisms.  It does not yet account for zone policy files; they will be implemented in a similar style later.

- A new schema is implemented, based on internal discussions of Nameshed's UI.  It's fairly incomplete, but gets the idea across.

- This design separates Nameshed's internal view of its configuration from any of the methods to configure it (CLI, env vars, or config file).  This ensures that the internal structures can evolve without changing any external interface.

- All of Nameshed's configuration is stored in a single location, instead of spread out among the various components.  This makes it easier to see the overall configuration state.

@ximon18, do you want to merge this to `main` before or after integrating all the components with it?


config.toml
```diff
@@ -0,0 +1,84 @@
...
+# Configuring how zones are loaded.
+[loader]
+# Where to listen for zone update notifications.
+#
+# A DNS server will be bound to these addresses.  If a DNS NOTIFY message for a
+# known zone is received there, the zone will be reloaded appropriately.
+#
+# Unless explicitly specified (e.g. 'udp://localhost:0000'), each address will
+# be served over UDP and TCP.  An empty array will disable listening entirely.
+notif-listeners = ["127.0.0.1:8050"] # TODO: Pick a good default port
```
ximon18, line 42: If you're going to use 5 out of the 6 letters of notify I would use the entire word.

config.toml
```diff
@@ -0,0 +1,84 @@
...
+# The location the daemon writes logs to.
+#
+# Logs are appended line-by-line to the specified file path.  It can be set to
+# '/dev/stdout' or '/dev/stderr' for standard output and error, respectively.
+# If it is a terminal, ANSI escape codes may be used to style the output.
+log-file = "/dev/stdout"
+
+# Configuring how zones are loaded.
+[loader]
+# Where to listen for zone update notifications.
```
ximon18, line 35: The term "update" might be confusing here as this is not about DNS UPDATE messages.

src/config/env.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    /// The file to write logs to.
+    pub log_file: Option<Box<Utf8Path>>,
+}
+
+impl EnvSpec {
+    /// Process environment variables.
+    pub fn process() -> Result<Self, EnvError> {
+        Ok(Self {
+            config: match std::env::var("NAMESHED_CONFIG") {
```
ximon18, line 28: Uou use _FILE below in NAMESHED_LOG_FILE so perhaps this should also have a _FILE suffix, or both should be called _PATH?

src/config/mod.rs
```diff
@@ -0,0 +1,351 @@
...
+    }
+}
+
+impl<T: Eq> Eq for Setting<T> {}
+
+//----------- SettingSource ----------------------------------------------------
+
+/// The source of a configured setting.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum SettingSource {
```
ximon18, line 286: Maybe note that the order matters here, i.e. that you compare these elsewhere and the "higher" one wins, which I assume is higher by automatically assigned unsigned integer enum variant ID?

src/config/mod.rs
```diff
@@ -0,0 +1,351 @@
...
+    pub hsm_stores: HashMap<Box<str>, HsmStoreConfig>,
+}
+
+//----------- HsmStoreConfig ---------------------------------------------------
+
+/// Configuration for an HSM store.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum HsmStoreConfig {
+    /// A PKCS#11 store.
+    PKCS11 {
```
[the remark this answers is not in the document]
ximon18: Regarding the PIN, or any secrets, @Philip-NLnetLabs noted that they should not be stored directly in the config file, but if in config then in a separate config file.

config.toml
```diff
@@ -0,0 +1,84 @@
...
+
+# The configuration file version.
+# - 'v1': This format.
+version = "v1"
+
+# Configuring the Nameshed daemon.
+#
+# This controls Nameshed's behaviour as a daemon, i.e. how it interacts with the
+# external system.
+[daemon]
```
ximon18, line 12: Syslog support seems to be missing.

src/config/file/v1.rs
```diff
@@ -0,0 +1,449 @@
+//! Version 1 of the configuration file.
```
ximon18, line 1: Is this version included in the config file somewhere? I didn't see it but it would be good if it was so that you can detect the version of the config file.

src/config/mod.rs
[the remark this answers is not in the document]
bal-e: I see in `KmipServerConnectionSettings` that it's possible to specify client certificates as PEM-format key/cert file paths, but also as a combined PKCS#12 key/cert file path.  Are those two options mutually exclusive?

src/config/mod.rs
bal-e: Yep, will document that.

config.toml
bal-e: Yeah, I'm going to change the `log-file` option to more generally support URIs, which we could use to specify a syslog target conveniently.

src/config/mod.rs
```diff
@@ -0,0 +1,351 @@
...
+        // Return the prepared configuration.
+        Ok(this)
+    }
+}
+
+//----------- DaemonConfig -----------------------------------------------------
+
+/// Daemon-related configuration for Nameshed.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct DaemonConfig {
```
ximon18, line 97: So perhaps deliberately at this point, but I note that daemon specific settings are missing here such as user name/group name/id to drop to, chroot jail, etc. I'm fine if we exclude those at this point and continue to evolve this as we go. We need to resolve what we are going to do about daemonbase `daemonizing` capabilities as that overlaps with this, but that doesn't have to be in this PR.

src/config/mod.rs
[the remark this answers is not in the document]
ximon18: Yes.

src/config/env.rs
ximon18: I think I slightly prefer PATH as FILE is never clear to me if it includes the path to the containing directory or if that is defined elsewhere.

src/config/mod.rs
bal-e: I will add those settings here shortly.

src/config/mod.rs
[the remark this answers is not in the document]
bal-e: As per internal discussions regarding HSM configuration and security, we decided to move the HSM stuff to internal Nameshed state (so it can only be manipulated via the CLI/API).

PR #15, opened 2025-08-11
Introduce a new zone loader

bal-e:
This lays out a brand new zone loader, based on the existing zonemaintainer code.  It builds in proper support for loading zonefiles, and uses new-base very prominently, but more importantly, demonstrates a new state management paradigm that I hope to use throughout Nameshed.  It makes it much easier to examine and consistently snapshot the program's state.  The difficulty is that I've never seen any other uses of this paradigm before, which means that I have to invent some things (e.g. scheduling events) from scratch.  But I'm very confident in its reliability.

Tasks:
- [x] Extract AXFR and IXFR loading from the zonemaintainer
- [x] Implement a zonefile loader around new-zonefile
- [x] Define the zone loader's state under the new paradigm
- [x] Implement automatic zone refresh scheduling
- [ ] Implement watching zonefiles for changes
  - [x] Implement a simple path monitor
- [x] Implement thorough logging (esp. trace)
- [ ] Reconsider `ZoneContents` wrt signer-created zone versions
- [ ] Implement zone addition and removal
  - [ ] Design resilient, atomic state changes
- [x] Add a way for users to request a refresh/reload of a zone
- [x] Integrate the new zone loader with the rest of Nameshed

@ximon18, I'd appreciate if you could test this out in your local sandbox.  The new-base zone loader doesn't support quite as many record types yet, but I'm willing to address those now.

### DIFF
--- a/src/tsig/file/mod.rs
+++ b/src/tsig/file/mod.rs
@@ -1,0 +1,63 @@
+//! The TSIG keys file.
+
+use std::{
+    fs::{self, File},
+    io::{self, BufReader, BufWriter},
+};
+
+use camino::Utf8Path;
+use serde::{Deserialize, Serialize};
+
+use crate::tsig::TsigStore;
+
+pub mod v1;
+
+//----------- Spec -------------------------------------------------------------
+
+/// A TSIG keys file.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case", tag = "version")]
+pub enum Spec {
+    /// The version 1 format.
+    V1(v1::Spec),
+}
+
+//--- Conversion
+
+impl Spec {
+    /// Parse from this specification.
+    pub fn parse(self, store: &mut TsigStore) {
+        match self {
+            Self::V1(spec) => spec.parse(store),
+        }
+    }
+
+    /// Build into this specification.
+    pub fn build(store: &TsigStore) -> Self {
+        Self::V1(v1::Spec::build(store))
+    }
+}
+
+//--- Loading / saving
+
+impl Spec {
+    /// Load and parse this specification from a file.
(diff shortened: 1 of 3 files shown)


src/common/file_io.rs
tertsdiepraam, line 1: The rustfmt config change definitely doesn't make this easy to review.

src/loader/mod.rs
tertsdiepraam, line 1: Ah there's the meat (thanks GitHub for hiding the diff)

tertsdiepraam (commented):
I might be dense, but I can't find where you remove the functionality you're replacing? Which of these files contains that?

src/common/file_io.rs
bal-e: Huh, I thought I had re-formatted the same code directly on `main` so it wouldn't show up in the diff.

src/lib.rs
```diff
@@ -1,6 +1,7 @@
 //! Rotonda
 #![allow(renamed_and_removed_lints)]
 #![allow(clippy::unknown_clippy_lints)]
+#![allow(clippy::uninlined_format_args)]
```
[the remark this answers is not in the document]
bal-e: Yes, but these pop up in all the code unrelated to mine and make it harder for me to use Clippy.  The right solution might be to fix all that code separately, but I think this particular lint represents a style choice and we may not want to enforce it.

bal-e:
> I can't find where you remove the functionality you're replacing?

Ah yeah, this PR supersedes all the zone maintainer code, and `zone_loader.rs`.  I'll remove them.

ximon18:
Would it be in scope to fix https://github.com/NLnetLabs/nameshed/issues/22 as part of this work?

ximon18:
Note: We decided to prioritize other work first and come back to this later.

PR #16, opened 2025-08-13
Delete http.rs and all code using it

mozzieongit:
Arya said to just delete http.rs and replace it with something new ^^
This separate PR is to make the actual HTTP API and CLI easier reviewable.

I've just deleted http.rs and then resolved all compilation errors and did not pay much attention to what I deleted, I might look at this later.


src/log.rs
```diff
@@ -374,7 +374,6 @@ impl LogConfig {
             // log level is set to debug or trace.
             res = res
                 .level_for("tokio_reactor", LevelFilter::Info)
-                .level_for("hyper", LevelFilter::Info)
                 .level_for("reqwest", LevelFilter::Info)
                 .level_for("h2", LevelFilter::Info)
```
bal-e, line 378: These probably need to be removed too.

bal-e (commented):
Looks good!

mozzieongit:
Turns out. It's not good. The http stuff is required for the approval hooks

PR #18, opened 2025-08-22
Rewrite the `log` module

bal-e:
- Implement `log::Log` manually, with an `RwLock`, to support changing the logging strategy after initialization.  This eliminates the need for `log-reroute` while providing more flexibility.

- Provide a safe prepare-apply scheme for changing the log config. This should compose nicely with larger configuration changes.

- Provide a fallback stderr logger, which works even if the logger is poisoned (i.e. a panic occurs while a write lock is held).

- Set up functionality for limiting trace-level logging to a user-specified list of targets.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -638,19 +638,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "env_logger"
-version = "0.10.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4cd405aab171cb85d6735e5c8d9db038c17d3ca007a4d2c25f337935c3d90580"
-dependencies = [
- "humantime",
- "is-terminal",
- "log",
- "regex",
- "termcolor",
-]
-
-[[package]]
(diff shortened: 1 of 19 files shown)


ximon18 (approved):
LGTM, did some manual testing of file, stderr and syslog logging and seemed to work. 

As discussed there is some potential confusion between the different meaning of TARGET in NAMESHED_LOG taking a LogTargetSpec and NAMESHED_LOG_TRACE_TARGETS, I guess this is resolvable through documentation. 

One oddity: `NAMESHED_LOG=file:/tmp/blah` works but `NAMESHED_LOG=file:-` does not do as I expected as it wrote to a file called `-` instead of logging to stdout. Also NAMESHED_LOG=stderr didn't work, should it?

bal-e:
I've made the possible inputs for the log target more obvious, both in `env.rs` and `args.rs`.  I've also rebased onto `main` and updated the CLI command.

ximon18 (approved):

PR #23, opened 2025-08-27
Send NOTIFY on new zone publication.

ximon18:
Also adds config support via new env var PS_XFR_OUT.

### DIFF
--- a/src/common/xfr.rs
+++ b/src/common/xfr.rs
@@ -8,13 +8,21 @@ use crate::common::tsig::TsigKeyStore;
 use crate::log::ExitError;
 use crate::zonemaintenance::types::{NotifyConfig, XfrConfig};
 
+/// Parse a given ACL string of the form "<addr>[:<port>][ KEY <TSIG key name>]"
+/// filling in the given [`XfrConfig`] and [`NotifyConfig`] objects with TSIG
+/// details and returning the determined SocketAddr (which will have port 0 if
+/// no port was specified).
+///
+/// A returned SocketAddr without port permits XFR only, while with a port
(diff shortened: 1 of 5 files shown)


etc/policy.template.toml
```diff
@@ -376,4 +376,18 @@ required = false
...
 # How published zones are served.
-[server]
+[server.outbound]
+
+# The set of nameservers from which SOA and XFR requests may be received.
+#
+# If empty, any nameserver may request XFR from us.
+#
+# A collection of IP addresses.
+accept-xfr-requests-from = []
```
mozzieongit, line 386: As this isn't enforced yet, maybe it shouldn't be in the template, yet. So that users don't get surprised this setting doesn't work

src/policy/file/v1.rs
```diff
@@ -674,7 +675,7 @@ impl SignerSerialPolicySpec {
 
 /// Spec for generating denial-of-existence records.
 #[derive(Clone, Debug, Default, Deserialize, Serialize)]
-#[serde(deny_unknown_fields, tag = "type")]
+#[serde(rename_all = "kebab-case", deny_unknown_fields, tag = "type")]
```
mozzieongit, line 678: Why kebab-case? we don't want `NSec` to turn into `n-sec` for example. that's why we're using `rename` on the items below.

ximon18: So this isn't really part of the PR, I just noticed it while nearby in the code and was confused by `opt_out` in my policy file working when it shouldn't. I just pushed a more targeted fix for this to this PR, is that better? We can also remove this change entirely.

PR #26, opened 2025-08-27
Actually cause the Zone Loader to add the new zone.

ximon18:
Adds a new `ApplicationCommand::RegisterZone` and extends `ZoneLoader` to honour it.


ximon18:
Note: this is missing various access control and zone signing settings, but more importantly it should also trigger KeyManager to invoke `dnst keyset create` and `dnst keyset init`, so maybe we should send the register command to "KS" as well.

PR #29, opened 2025-08-29
Use unbounded MPSC channels everywhere

bal-e:
This helps avoid infecting everything with 'async'.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -81,7 +81,7 @@ fn main() -> ExitCode {
         };
 
         // Shut down Nameshed.
-        manager.terminate();
+        manager.terminate().await;
         result
     })
 }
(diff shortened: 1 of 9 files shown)


ximon18 (approved):
I haven't tried compiling or running this, I assume you have.

I presume this just swaps one kind of fire-and-forget for another, but one difference I can think of is that the removed .await points are no longer opportunities for the Tokio task scheduler to switch to another task.

As long as you think this is not a problem I say go ahead, we'll find out easily enough anyway in the testing we are and will be doing.

bal-e:
The idea of bounded channels (as we were using previously) was to provide some backpressure in the system, but 1) it's a very limited and inflexible form of backpressure, and 2) nothing in our system is so performance-intensive, or expecting so many messages, for backpressure to be necessary.  I've checked the code compiles and I don't expect any issues.

PR #30, opened 2025-08-29
Forward NOTIFYs from the PS to the ZL

bal-e:
This removes the zone loader's NOTIFY support in favor of listening through the publication server.  This means users can use the same port for getting published zones and pinging the server for updated zones.

At the moment, SOA serials are not checked, and requests for arbitrary zones are forwarded.  At some point, we should update the NOTIFY middleware service to pass SOA serials forward, and should only forward zones known to the zone loader.  The latter point can be a bit difficult because we can't check the `ZoneTree`; it only contains zones that have been loaded already, and we might want to allow NOTIFY-ing for a zone that was not successfully loaded yet.  The upcoming state work will make this possible.

### DIFF
--- a/src/comms.rs
+++ b/src/comms.rs
@@ -67,7 +67,10 @@
 
 use domain::base::Serial;
 use domain::zonetree::StoredName;
-use std::fmt::{self, Debug};
+use std::{
+    fmt::{self, Debug},
+    net::IpAddr,
+};
 
 use crate::api::ZoneAdd;
 
@@ -134,6 +137,26 @@ pub enum ApplicationCommand {
         zone_name: StoredName,
         zone_serial: Serial,
     },
+
+    /// Refresh a zone.
+    ///
+    /// The zone loader will initiate a refresh for the zone, and query the
+    /// zone's source to look for a newer version of the zone.  This command
+    /// can be used in response to a user request or a NOTIFY message.
+    RefreshZone {
+        /// The name of the zone to refresh.
+        zone_name: StoredName,
+
+        /// The source address of the NOTIFY message.
+        source: Option<IpAddr>,
+
+        /// The expected new SOA serial for the zone.
+        ///
+        /// If this is set, and the zone's SOA serial is greater than or equal
+        /// to this value, the refresh can be ignored.
+        serial: Option<Serial>,
+    },
+
     SignZone {
         zone_name: StoredName,
         zone_serial: Option<Serial>,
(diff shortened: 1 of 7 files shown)


src/comms.rs
```diff
@@ -134,6 +137,22 @@ pub enum ApplicationCommand {
...
+    /// zone's source to look for a newer version of the zone.  This command
+    /// can be used in response to a user request or a NOTIFY message.
+    RefreshZone {
+        /// The name of the zone to refresh.
+        zone_name: StoredName,
+
+        /// The SOA serial and source address from the NOIFY message.
+        //
+        // NOTE: The serial number is currently ignored.
+        notify_info: Option<(Serial, IpAddr)>,
```
ximon18, line 153: Should the IpAddr be optional, in the case of an internally raised event for example?

ximon18: Ah ignore me, the whole notify_info is optional I see.

src/units/zone_loader.rs
```diff
@@ -86,13 +86,10 @@ use crate::zonemaintenance::types::{
...
 #[derive(Debug)]
 pub struct ZoneLoader {
-    /// Addresses and protocols to listen on.
-    pub listen: Vec<ListenAddr>,
-
     /// The zone names and (if primary) corresponding zone file paths to load.
     pub zones: Arc<HashMap<String, String>>,
 
-    /// XFR in per secondary zone: Allow NOTIFY from, and when with a port also request XFR from.
+    /// XFR in per secondary zone.
```
ximon18, line 92: I think the old comment was still relevant, the ZM will ignore a NOTIFY that is not permitted IIRC.

src/units/zone_loader.rs
```diff
@@ -208,59 +185,68 @@ impl ZoneLoader {
...
+                            notify_info: None,
+                        }) => {
+                            zone_maintainer.force_zone_refresh(&zone_name, Class::IN).await;
                         }
+
+                        Some(ApplicationCommand::RefreshZone {
+                            zone_name,
+                            notify_info: Some((serial, source)),
+                        }) => {
+                            // TODO: Support checking the serial number.
```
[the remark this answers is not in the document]
bal-e: The NOTIFY RFC specifies that the NOTIFY message includes a serial number, and that the client should ignore the NOTIFY message (without even performing a SOA query) if that number is older or equal.

ximon18 (approved):

PR #32, opened 2025-08-29
Update env var handling.

ximon18:
Remove obsoleted env var handling, and fix XFR/NOTIFY configuration.

- Use zone names as HashMap keys because they will match both "example.com" and "example.com." while string matching will not, and because the only way to add zones now is from the CLI which sends a validated zone name.
- If XFR/NOTIFY settings exist for the zone being added _for the same source_, use them, but if the source is different then use default settings.
- Make logging for primary and secondary configuration consistent.
- Change the default mode of XFR from None to IXFR with AXFR fallback, as if a primary is added we probably want to serve XFR by default to secondaries.

With this change a zone is not always added, you must now use the CLI to add a zone. You can still however preconfigure via env vars XFR / NOTIFY settings, because the CLI doesn't support that yet. Preconfigured settings will be used if they match the zone being added by zone name and source address, otherwise defaults will be used.

### DIFF
--- a/src/units/zone_loader.rs
+++ b/src/units/zone_loader.rs
@@ -189,28 +189,28 @@ impl ZoneLoader {
 
                         Some(ApplicationCommand::RegisterZone { register }) => {
                             let res = match register.source {
-                                        ZoneSource::Zonefile { path /* Lacks XFR out settings */ } => {
-                                            Self::register_primary_zone(
-                                                register.name.clone(),
-                                                &path.to_string(),
-                                                component.tsig_key_store(),
-                                                None,
-                                                &self.xfr_out,
-                                                &zone_updated_tx,
-                                              ).await
-                                        }
-                                        ZoneSource::Server { addr /* Lacks TSIG key name */ } => {
-                                            // Use any existing XFR inbound
-                                            // ACL that has been defined for
-                                            // this zone from this source.
-                                            Self::register_secondary_zone(
-                                                register.name.clone(),
-                                                component.tsig_key_store(),
-                                                Some(addr),
-                                                &self.xfr_in,
-                                                zone_updated_tx.clone(),
-                                            )
-                                        },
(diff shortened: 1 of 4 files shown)


src/units/zone_loader.rs
```diff
@@ -226,16 +221,25 @@ impl ZoneLoader {
...
                                         ZoneSource::Server { addr /* Lacks TSIG key name */ } => {
-                                            let xfr_in = format!("{addr}");
+                                            // Use any existing XFR inbound
+                                            // ACL that has been defined for
+                                            // this zone from this source.
+                                            let xfr_in = self.xfr_in
+                                                .get(&register.name)
+                                                .map(|v| v.to_string())
+                                                .unwrap_or_default();
+                                            error!("XIMON: {xfr_in}");
```
bal-e, line 237: I don't think we want to merge that :p

src/units/zone_loader.rs
```diff
@@ -128,33 +128,28 @@ impl ZoneLoader {
...
-        let no_xfr_settings = "".to_string();
-
         // Load primary zones.
         // Create secondary zones.
         for (zone_name, zone_path) in self.zones.iter() {
-            let Ok(apex_name) = Name::from_str(zone_name) else {
-                error!("[ZL]: Error: Zone name '{zone_name}' is invalid. Skipping zone.");
-                continue;
-            };
+            error!("[ZL]: Error: Zone name '{zone_name}' is invalid. Skipping zone.");
```
bal-e, line 134: I think this line was meant to be removed entirely.  The error is no longer possible.

bal-e (commented):
I plan to simplify the zone loader code wrt configuring primary/secondary zones soon, as part of integrating it with the new state stuff (which will also provide the XFR and TSIG info differently).  But we can still merge these changes in the meantime.

ximon18:
Discussed with @bal-e, will merge.

PR #34, opened 2025-08-29
Rename to cascade and add packaging

mozzieongit:
Left over TODOs:
- [x] Create cascade email alias
- [x] Create description for the project (and repo)
  - Unless "A batteries-included DNSSEC signing solution" is good
- Create packages for the dependencies (dnst and kmip2pkcs11) or disable package testing (in progress)


mozzieongit:
Packaging basically works. 
the pkg-test job fails because it's missing dnst and ~nameshed-hsm-relay~ cascade-hsm-relay in the repositories.

Cargo.toml
```diff
@@ -100,3 +104,69 @@ syslog      = "7.0"
 [features]
 default = []
 tls = ["tokio-rustls", "rustls-pemfile"]
+
+# TODO: Do we want this?
```
ximon18, line 108: So Routinator and Krill both have this set to `panic = "abort"`. The rationale given in the Krill `cargo.toml` is "# Make sure that Krill crashes on panics, rather than losing threads and limping on in a bad state". I.e. a panic in a particular thread could leave the application in a bad state so it is better to exit completely. That seems sensible, and clearly as we already do it for at least two projects we've discussed this before and concluded this was the best approach. Does anyone have a counter argument here? The only one I can think of is not based on fact, which is I'm wondering does panic without abort allow Drop handlers to run which if not run could cause even more mess, such as leaving persisted data in an inconsistent state?

Cargo.toml
```diff
@@ -100,3 +104,69 @@ syslog      = "7.0"
...
+
+[package.metadata.deb]
+assets = [
+	[
+		"target/release/cascade",
+		"usr/bin/",
+		"755",
+	],
+	[
+		"target/release/cascadec",
```
ximon18, line 120: While not caused by this PR, we should decide what we want the client binary to be called. `+c` is an approach we used for Krill, but NSD uses `nsd-control` and OpenDNSSEC uses `ods-control`, `ods-signer` and `ods-enforcer` with the daemons having `+d`, which we may want to do as well?

Cargo.toml
Philip-NLnetLabs: I think abort is best. And cascade should be able to handle crashes. The computer/VM may also crash. That should not affect the consistency of state. Mimir seems to be an exception where continuing works fine.

Cargo.toml
Philip-NLnetLabs: Maybe we use cascaded for the daemon, then we can use cascade for the cli. cascade-control seems a bit long.

Cargo.toml
```diff
@@ -100,3 +104,69 @@ syslog      = "7.0"
...
+	#{ source = "doc/cascadec.1", dest = "/usr/share/man/man1/cascadec.1", mode = "644", doc = true },
+	{ source = "config.toml", dest = "/etc/cascade/config.toml", mode = "644", config = true },
+	{ source = "pkg/common/service.preset", dest = "/usr/lib/systemd/system-preset/50-cascade.preset", mode = "644" },
+]
+
+# ensure that the useradd tools are present by installing their respective packages
+[package.metadata.generate-rpm.requires]
+shadow-utils = "*"
+dnst = "*"
+nameshed-hsm-relay = "*"
```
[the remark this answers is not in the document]
mozzieongit: True. I now realize that it's just the name to the not-yet existing package.

Cargo.toml
ximon18: `+d` is quite a normal thing in the Linux world though, I suppose a direct comparison like `knotc` and `knotd` fails because we don't say `cascaded` and `cascadec`, but the single character difference between c and d is both easy to mistype and prevents auto-complete so I prefer `cascade` without the `c` for that reason. This is also a reason against `xxx-control` because someone can accidentally invoke the daemon via `xxx` when they meant the client, so the client having a shorter name than the daemon reduces the chance of that foot gun.

ximon18:
> Packaging basically works. the pkg-test job fails because it's missing dnst and ~nameshed-hsm-relay~ cascade-hsm-relay in the repositories.

Needs changing again, now to `kmip2pkcs11`.

src/manager.rs
```diff
@@ -440,8 +440,8 @@ impl Manager {
                         server_addr,
                         server_port,
                         server_insecure: true,
-                        server_username: std::env::var("NAMESHED_HSM_RELAY_USERNAME").ok(),
-                        server_password: std::env::var("NAMESHED_HSM_RELAY_PASSWORD").ok(),
+                        server_username: std::env::var("CASCADE_HSM_RELAY_USERNAME").ok(),
```
ximon18, line 443:
```suggestion
                        server_username: std::env::var("KMIP2PKCS11_USERNAME").ok(),
```

ximon18 (approved):

tertsdiepraam (approved):

PR #35, opened 2025-08-29
Prepare components for the new state system

bal-e:
The following components are introduced:

- [x] Zone policies.
- [x] Per-zone states.
- [x] A new TSIG key store.

These components will fit together as part of the new state system.
All of them support loading/saving and are very transparent.

### DIFF
--- a/src/config/file/v1.rs
+++ b/src/config/file/v1.rs
@@ -87,12 +87,12 @@ impl Spec {
 
     /// The default value for `zone_state_dir`.
     fn zone_state_dir_default() -> Box<Utf8Path> {
-        "/var/db/cascade/zone-state".into()
+        "/var/db/cascade/zone-state.db".into()
     }
 
     /// The default value for `tsig_store_path`.
     fn tsig_store_path_default() -> Box<Utf8Path> {
-        "/var/db/cascade/tsig-keys".into()
+        "/var/db/cascade/tsig-keys.db".into()
     }
 }
 
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -52,8 +52,8 @@ impl Default for Config {
     fn default() -> Self {
         Self {
             policy_dir: "/etc/cascade/policies".into(),
-            zone_state_dir: "/var/db/cascade/zone-state".into(),
-            tsig_store_path: "/var/db/cascade/tsig-keys".into(),
+            zone_state_dir: "/var/db/cascade/zone-state.db".into(),
+            tsig_store_path: "/var/db/cascade/tsig-keys.db".into(),
             daemon: Default::default(),
             loader: Default::default(),
             signer: Default::default(),
@@ -336,6 +336,13 @@ impl<T> Setting<T> {
 
     /// The current value.
     pub const fn value(&self) -> &T {
+        // This is a 'const' implementation of:
+        //
+        // self.args.as_ref()
+        //     .or(self.env.as_ref())
+        //     .or(self.file.as_ref())
+        //     .unwrap_or(&self.default)
+
         match self {
             Self {
                 args: Some(value), ..
--- a/src/policy/file/v1.rs
+++ b/src/policy/file/v1.rs
@@ -126,7 +126,6 @@ pub struct SignerSpec {
     //
     // TODO:
     // - Signing policy (disabled, pass-through?, enabled)
-    // - Serial number policy
 }
 
 //--- Conversion
--- a/src/policy/mod.rs
+++ b/src/policy/mod.rs
@@ -59,14 +59,28 @@ pub fn reload_all(
         // Filter for UTF-8 paths.
         let Ok(path) = Utf8PathBuf::from_path_buf(entry.path()) else {
             log::warn!(
-                "Ignoring potential policy file '{}' as the path is non-UTF-8",
+                "Ignoring potential policy '{}' as the path is non-UTF-8",
                 entry.path().display()
             );
             continue;
         };
 
+        // Filter hidden files.
+        if path
+            .file_name()
+            .expect("this path has a known parent directory")
+            .starts_with('.')
+        {
+            log::debug!("Ignoring hidden file '{path}' among policies");
+            continue;
+        }
+
         // Filter for '.toml' files.
-        if path.extension() != Some("toml") {
+        if path
+            .extension()
+            .is_none_or(|e| !e.eq_ignore_ascii_case("toml"))
+        {
+            log::warn!("Ignoring potential policy '{path}'; policies must end in '.toml'");
             continue;
         }
 
@@ -77,7 +91,10 @@ pub fn reload_all(
         let spec = match file::Spec::load(&path) {
             Ok(spec) => spec,
             // Ignore a directory ending in '.toml'.
-            Err(err) if err.kind() == io::ErrorKind::IsADirectory => continue,
+            Err(err) if err.kind() == io::ErrorKind::IsADirectory => {
+                log::warn!("Ignoring potential policy '{path}'; policies must be files");
+                continue;
+            }
             Err(err) => return Err(err),
         };
 
--- a/src/zone/state/mod.rs
+++ b/src/zone/state/mod.rs
@@ -3,7 +3,7 @@
 use std::{
     collections::hash_map,
     fs,
-    io::{self, BufReader, BufWriter},
+    io::{self, BufReader},
     sync::Arc,
 };
 
@@ -106,19 +106,7 @@ impl Spec {
 
     /// Build and save this specification to a file.
     pub fn save(&self, path: &Utf8Path) -> io::Result<()> {
-        fs::create_dir_all(
-            path.parent()
-                .expect("'path' must be a file, so it must have a parent"),
-        )?;
-
-        let file = BufWriter::new(
-            fs::OpenOptions::new()
-                .create(true)
-                .write(true)
-                .truncate(true)
-                .open(path)?,
-        );
-        serde_json::to_writer(file, self)?;
-        Ok(())
+        let text = serde_json::to_string(self)?;
+        crate::util::write_file(path, text.as_bytes())
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -353,6 +353,7 @@ dependencies = [
  "serde_with",
  "smallvec",
  "syslog",
+ "tempfile",
  "tokio",
  "tokio-metrics",
  "tokio-rustls 0.23.4",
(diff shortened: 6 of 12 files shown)


src/config/mod.rs
```diff
@@ -287,54 +308,70 @@ pub enum LogTarget {
...
+            Self {
+                args: Some(value), ..
+            }
+            | Self {
+                env: Some(value), ..
+            }
+            | Self {
+                file: Some(value), ..
+            }
+            | Self { default: value, .. } => value,
```
tertsdiepraam, line 349: I wonder if
```rust
self.args
    .or(self.env)
    .or(self.file)
    .unwrap_or(self.default)
```
is slightly more explicit about the ordering (even though it's functionally the same).

tertsdiepraam: I love this system btw, to be able to see all of these values and tell the user where the things come from.

src/policy/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+            log::warn!(
+                "Ignoring potential policy file '{}' as the path is non-UTF-8",
+                entry.path().display()
+            );
+            continue;
+        };
+
+        // Filter for '.toml' files.
+        if path.extension() != Some("toml") {
+            continue;
```
tertsdiepraam, line 70: Should we warn here?
```
Only TOML is supported for policy definitions
```
or something like that.

src/config/env.rs
```diff
@@ -33,7 +33,7 @@ impl EnvSpec {
                 .transpose()
         }
 
-        let config_path =
+        let config =
             var("NAMESHED_CONFIG_PATH")?.map(|path| Utf8PathBuf::from(path).into_boxed_path());
 
         let log_level = var("NAMESHED_LOG_LEVEL")?
```
ximon18, line 39:
```suggestion
        let log_level = var("CASCADE_LOG_LEVEL")?
```

src/config/file/v1.rs
```diff
@@ -7,16 +7,27 @@ use serde::Deserialize;
...
+
+    /// The directory storing per-zone state files.
+    #[serde(default = "Spec::zone_state_dir_default")]
+    pub zone_state_dir: Box<Utf8Path>,
+
+    /// The file storing TSIG keys.
+    #[serde(default = "Spec::tsig_store_path_default")]
+    pub tsig_store_path: Box<Utf8Path>,
+
     /// Configuring the Nameshed daemon.
```
ximon18, line 31: Maybe we also need a file for storing KMIP credentials?

Currently the KMIP support in `dnst keyset` includes a subcommand `dnst keyset kmip add-server` which takes a credentials filename as an argument, along with optional username and password arguments, and creates the credentials file if it doesn't exist, otherwise uses the credentials from it.

When adding a zone to Cascade we'll need to provide these credentials, and perhaps to streamline things we should just store them all the KMIP credentials in a single KMIP credentials file and have them passed through from the command line.

Whatever we do maybe it is desirable to be consistent in handling of secrets and thus KMIP credentials as well as TSIG keys,

src/config/file/v1.rs
```diff
@@ -48,6 +62,40 @@ impl Spec {
...
+    }
+
+    /// The default value for `zone_state_dir`.
+    fn zone_state_dir_default() -> Box<Utf8Path> {
+        "/var/db/nameshed/zone-state".into()
+    }
+
+    /// The default value for `tsig_store_path`.
+    fn tsig_store_path_default() -> Box<Utf8Path> {
+        "/var/db/nameshed/tsig-keys".into()
```
ximon18, line 95:
```suggestion
        "/var/db/cascade/tsig-keys".into()
```

ximon18, line 95: "db" ? This could cause some confusion / misinterpretation as there's no "database" here, though one can argue about the definition of database of course...

src/config/file/v1.rs
```diff
@@ -79,42 +127,30 @@ impl DaemonSpec {
...
-                    value: LogTarget::File("/var/log/nameshed.log".into()),
-                }),
+            level: Setting {
+                default: LogLevel::Info,
+                file: self.log_level.map(|v| v.build()),
+                env: None,
+                args: None,
+            },
+            target: Setting {
+                default: LogTarget::File("/var/log/nameshed.log".into()),
```
ximon18, line 137:
```suggestion
                default: LogTarget::File("/var/log/cascade.log".into()),
```

src/config/mod.rs
```diff
@@ -38,52 +46,42 @@ pub struct Config {
...
 
 impl Config {
     /// Set up a [`clap::Command`] with config-related arguments.
     pub fn setup_cli(cmd: clap::Command) -> clap::Command {
         args::ArgsSpec::setup(cmd)
     }
 
-    /// Process all configuration sources.
-    pub fn process(cli_matches: &clap::ArgMatches) -> Result<Self, ConfigError> {
+    /// Initialize Nameshed's configuration.
```
ximon18, line 74:
```suggestion
    /// Initialize Cascade's configuration.
```

src/policy/file/v1.rs
```diff
@@ -0,0 +1,346 @@
...
+
+impl KeyManagerSpec {
+    /// Parse from this specification.
+    pub fn parse(self) -> KeyManagerPolicy {
+        KeyManagerPolicy {}
+    }
+
+    /// Build into this specification.
+    pub fn build(policy: &KeyManagerPolicy) -> Self {
+        let KeyManagerPolicy {} = policy;
```
ximon18, line 101: ? Is this a placeholder for future behaviour?

src/policy/file/v1.rs
```diff
@@ -0,0 +1,346 @@
...
+
+    /// How denial-of-existence records are generated.
+    pub denial: SignerDenialSpec,
+
+    /// Reviewing signed zones.
+    pub review: ReviewSpec,
+    //
+    // TODO:
+    // - Signing policy (disabled, pass-through?, enabled)
+    // - Serial number policy
```
ximon18, line 129: You have serial number policy above.

src/policy/file/v1.rs
```diff
@@ -0,0 +1,346 @@
...
+pub enum SignerDenialSpec {
+    /// Generate NSEC records.
+    NSec,
+
+    /// Generate NSEC3 records.
+    NSec3 {
+        /// Whether and how to enable NSEC3 Opt-Out.
+        opt_out: Nsec3OptOutSpec,
+    },
+}
```
ximon18, line 218: I wonder how we want to handle NSEC <-> NSEC3 migration...

src/policy/file/v1.rs
```diff
@@ -0,0 +1,346 @@
...
+    }
+}
+
+//----------- ReviewSpec -------------------------------------------------------
+
+/// Policy for reviewing loaded/signed zones.
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case", deny_unknown_fields, default)]
+pub struct ReviewSpec {
+    /// Whether review is required.
```
ximon18, line 299: Required is a bit vague to me. Does it mean the hook won't be invoked, or will be invoked but the result will be ignored?

src/policy/file/v1.rs
```diff
@@ -0,0 +1,346 @@
...
+    pub required: bool,
+
+    /// A command hook for reviewing a new version of the zone.
+    pub cmd_hook: Option<String>,
```
ximon18, line 303: Should this be a PathBuf rather than a String?

src/policy/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+
+    /// Whether the policy is being deleted.
+    ///
+    /// This is an intermediate state used to prevent race conditions while the
+    /// policy is being removed.  In this state, new zones cannot be attached to
+    /// this policy.
+    pub mid_deletion: bool,
+
+    /// The zones using this policy.
+    pub zones: foldhash::HashSet<Name<Bytes>>,
```
ximon18, line 29: Is zone name enough? I'm wondering if we need to know that an already signed zone used this policy when we (incrementally) resign it, while if the policy is changed for the same zone would the next unsigned version of the zone used the new policy while the previously signed version refers to the old policy?

src/policy/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+        // Record the new policy.
+        let prev = new_policies.insert(name.into(), policy);
+        assert!(prev.is_none(), "there is at most one policy per path");
+    }
+
+    // Traverse policies whose files were not found.
+    for (name, policy) in policies.drain() {
+        // If any zones are using this policy, keep it.
+        if !policy.zones.is_empty() {
+            log::error!("The file backing policy '{name}' has been removed, but some zones are still using it; Nameshed will preserve its internal copy");
```
ximon18, line 106:
```suggestion
            log::error!("The file backing policy '{name}' has been removed, but some zones are still using it; Cascade will preserve its internal copy");
```

src/zone/state/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /// Build and save this specification to a file.
+    pub fn save(&self, path: &Utf8Path) -> io::Result<()> {
+        fs::create_dir_all(
+            path.parent()
+                .expect("'path' must be a file, so it must have a parent"),
+        )?;
+
+        let file = BufWriter::new(
```
ximon18, line 114: Is this atomic, or should we write then rename?

ximon18 (approved):
Looks like a nice start with some good intentions, a bit hard to take in in one quick pass through but that's on me.

src/config/file/v1.rs
bal-e: I agree!  One day, we may want to have a single credentials file storing TSIG, KMIP, maybe even TLS secrets.  I don't know whether we want to implement that now (as that would involve changing `dnst keyset`) or to just support both TSIG and KMIP separately.

src/config/file/v1.rs
bal-e: Yeah, I wasn't sure what extension to add.  `.db` feels generic enough, and might guide users to treat it as an opaque file.  Will do.

src/policy/file/v1.rs
bal-e: Yes, we can fill in the `KeyManagerPolicy` struct in the future.  I wasn't clear on all the details for it yet, so I just left it empty.

src/policy/file/v1.rs
bal-e: I did think about that a bit.  That information needs to be stored in the state of each zone, which the signer will have to query.  For the PoC, I don't think we have the time to support such migrations, but I'm confident it can be implemented (relatively) easily afterwards.

src/zone/state/mod.rs
bal-e: :facepalm: I'll update it to be more careful.

PR #37, opened 2025-09-01
Scaffold docs

tertsdiepraam:
The most controversial thing in here is probably to use `uv` locally and export `requirements.txt` from there. I could remove that if that's preferred. The commands are:
```
uv run make html
uv export --format requirements.txt > requirements.txt
```


ximon18:
> The most controversial thing in here is probably to use uv locally and export requirements.txt from there. I could remove that if that's preferred

What was the rationale for doing so? Did Routinator also do that?

tertsdiepraam:
No, routinator doesn't have it. I added it so that I don't need to think about venvs, installing the right python version and locking dependencies.

doc/manual/pyproject.toml
```diff
@@ -0,0 +1,16 @@
+[project]
+name = "manual"
+version = "0.1.0"
+description = "Add your description here"
```
ximon18, line 4: Should we populate this? I see it is not set for the roto project version either.

ximon18 (approved):
I'm not an expert on this but assuming it technically works, the settings seem sane aside from a couple of nits I point out in comments.

doc/manual/pyproject.toml
tertsdiepraam: It's not really an issue because we don't publish this Python project, but I added a little description anyway.

PR #38, opened 2025-09-01
Extract code inside `tokio::select!()` to allow it to be formatted.

ximon18:
As for some reason cargo fmt/rustfmt does nothing inside macro! invocation bodies.

This PR does NOT contain any logic changes, it only moves code into helper functions which can then be formatted as usual. It also reduces the size of the long match handle blocks making the code more readable.

This PR may conflict / overlap with other PRs, I'll take the merge hit, I assume there'll still be some beneficial code changes  left in this PR even if other bits get obsoleted by other PRs.


bal-e (approved):
This is a great step towards making this event-driven architecture easier to maintain.

PR #41, opened 2025-09-01
installation and building scaffolding

AlexanderBand:


doc/manual/source/installation.rst
```diff
@@ -0,0 +1,412 @@
+Installation
+============
+
+System Requirements
+-------------------
+
+When choosing a system for Cascade, Make sure you have [TODO] of available
+memory and [TODO] of disk space for the application.
```
tertsdiepraam, line 8: Won't this depend heavily on the zones? Might be difficult to create a general guideline.

mozzieongit:
2 Options:
Remove cargo install instructions, or
Update cargo install instruction to
- dependencies,
- creating directories,
- config file

doc/manual/source/building.rst
```diff
@@ -101,6 +146,8 @@ use the ``--force`` option to overwrite an existing version:
 
     cargo install --locked --force cascade --version 0.1.0-rc1
 
+Make sure to install a compatible version of ``dnst``.
```
mozzieongit, line 149: Will we keep track of compatible dnst version in the Cascade readme?

mozzieongit:
> The secondaries only need SOA and XFR query support from Cascade.
> 
> We should document this expected usage quite prominently.

Indeed. And mention that only those queries are supported, and that dnssec queries can't be done, not even for testing/inspection of the zone.

mozzieongit:
TODO for me:
- Add Quick Start with simple setup
- Do KMIP setup for myself and document how it works

PR #46, opened 2025-09-02
Add basic CI

tertsdiepraam:
Let's see whether this works! I'm expecting failures!


bal-e (approved):
Looks good!  And CI would be lovely.

PR #47, opened 2025-09-03
Introduce global state

bal-e:
This adds `Center` and `State` to represent Cascade's entire state, and integrates it with the current codebase.  `Center` absorbs the previous `Component` type and includes `Mutex<State>`, providing the whole system with access to global state (config, zones, etc).  This will be followed up by a PR to implement saving/loading for `State`, i.e. global state persistence.  We can then begin using `State` in each unit.


src/targets/mod.rs
tertsdiepraam, line 1: The `Target` seems to be an outdated concept that's not applicable to Cascade anymore. Unless I'm missing something important about it (@ximon18?). Maybe we can remove it and make the CentralCommand a unit?

tertsdiepraam (approved):
I love this! But I'd also like to hear what @ximon18 thinks. (And you have some lint errors)

tertsdiepraam:
Ah I just introduced some conflicts. Apologies

tertsdiepraam:
Wrong button dammit

bal-e:
Running out of work time, so I'm just going to merge this and keep going.

PR #49, opened 2025-09-03
Start integrating policy with the signer

tertsdiepraam:


src/units/zone_signer.rs
```diff
@@ -716,33 +695,19 @@ impl ZoneSigner {
...
-            ) => todo!(),
         };
 
-        let _add_used_dnskeys = true;
         let now = Timestamp::now().into_int();
-        let inception = now.sub(self.inception_offset_secs).into();
-        let expiration = now.add(self.expiration_offset).into();
-        SigningConfig::new(denial, inception, expiration)
+        let inception = now.wrapping_sub(policy.signer.sig_inception_offset.as_secs() as u32);
+        let expiration = now.wrapping_sub(policy.signer.sig_validity_time.as_secs() as u32);
```
bal-e, line 709: This should be a `wrapping_add`, right?

bal-e (changes requested):

bal-e (approved):
Looks good!

PR #50, opened 2025-09-03
Implement state persistence

bal-e:
This implements a large and crucial feature for Cascade: state persistence.

On startup, the state file will be loaded; if it is found, Cascade continues from there, otherwise a new state is built from scratch (and the configuration file is loaded to seed it).

The process of integrating `Center` with the HTTP API is begun.  Zone addition and removal now hook through methods on `Center`; these will eventually supersede the older implementations (the zone loader needs to be integrated better before that can happen).

All in all, this brings Cascade significantly closer to being a usable, resilient daemon.  The next step is to integrate every unit into this system better, in particular the zone loader (by keeping the source of each zone in `ZoneState`) and the zone signer (by using zone policies).


ximon18:
In a quick test I saw:
```
State file not found; starting from scratch
...
[2025-09-03T21:49:03.733Z] ERROR cascade::center: Could not save global state: Permission denied (os error 13)
```
Where was it trying to save it? It would be good if we can include the path with this error.

bal-e:
> why it is periodically saving the state when no change has occurred?

Yeah, I had left this out initially.  I'll implement save-only-on-change mechanics now.

tertsdiepraam:
Moved a commit from my PR to this because the `change_policy` call was relying on it.

PR #51, opened 2025-09-04
Get listen addrs for servers from config.toml

tertsdiepraam:
They don't reload when the config changes, but seems like a good start.

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -89,10 +89,6 @@ pub fn spawn(
     log::info!("Starting unit 'RS'");
     let unit = ZoneServerUnit {
         center: center.clone(),
-        listen: vec![
-            "tcp:127.0.0.1:8056".parse().unwrap(),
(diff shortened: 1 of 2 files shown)


ximon18:
> They don't reload when the config changes, but seems like a good start.

That would not be possible for ports that must be bound before privileges are dropped so we may not be able to make them runtime reconfigurable.

My work in https://github.com/NLnetLabs/cascade/pull/53 now overlaps with this PR.

In my PR I extended the code to accept both socket addresses to bind to AND pre-bound sockets received from the O/S. But perhaps we should make all socket binding happen before daemonization (so that binding to a low numbered port happens before dropping privileges, which, if we fork for daemonization, also has to happen before starting Tokio and thus before starting units) and pass only bound sockets/listeners into the units.

ximon18:
After thinking about this some more, perhaps we should merge my PR and this PR, and make a new PR after that which would do early port binding.

PR #56, opened 2025-09-08
Policy commands

tertsdiepraam:
Example output of `policy show`:
```
foo:
  zones: <none>
  loader:
    review:
      required: false
      cmd_hook: <none>
  key manager: <unimplemented>
  signer:
    serial policy: counter
    signature inception offset: 0 seconds
    signature validity offset: 0 seconds
    denial: NSEC3 (opt-out: disabled)
    review:
      required: false
      cmd_hook: <none>
  server: <unimplemented>
```

Example output of `policy reload`:
image

And `policy list` is simply:
```
policy1
policy2
policy3
```

### DIFF
--- a/src/cli/commands/policy.rs
+++ b/src/cli/commands/policy.rs
@@ -2,7 +2,10 @@ use futures::TryFutureExt;
 use log::error;
 
 use crate::{
-    api::{PolicyListResult, PolicyReloadResult},
+    api::{
+        Nsec3OptOutPolicyInfo, PolicyInfo, PolicyInfoError, PolicyListResult, PolicyReloadResult,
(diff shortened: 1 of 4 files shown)


tertsdiepraam:
This seems to be about as far as I can take this now. Happy to take suggestions on the formatting and phrasing of the output.

src/cli/commands/policy.rs
```diff
@@ -53,10 +72,18 @@ impl Policy {
                         error!("HTTP request failed: {e}");
                     })?;
 
-                println!("Policy info: {res:?}");
+                let p = match res {
+                    Ok(p) => p,
+                    Err(e) => {
+                        eprintln!("{e:?}");
```
ximon18, line 78: Should this be `error!()` instead of `eprintln!()` ?

PR #57, opened 2025-09-08
Add a service specific variant of the default config.

ximon18:


tertsdiepraam (approved):
Maybe it would be even better if we could have overrides in the daemon section or something. But this is probably a good idea for now.

PR #61, opened 2025-09-10
Keep track of the lowest signature expiration time.

Philip-NLnetLabs:


src/units/zone_signer.rs
```diff
@@ -624,6 +625,40 @@ impl ZoneSigner {
             .saturating_add(insert_start.elapsed())
             .as_secs();
 
+        let reader = zone.read();
+        let apex_name = zone_name.clone();
+        let min_expiration = Arc::new(MinTimestamp::new());
+        let saved_min_expiration = min_expiration.clone();
+        reader.walk(Box::new(move |name, rrset, _cut| {
```
ximon18, line 632: Instead of walking the zone another option could be to modify `fn rrsig_inserter()` which sees every RRSIG that gets added to the zone.

Unless there are cases where you need to check existing RRSIGs, and then the existing zone walk done by `fn collect_zone()` could also be modified.

The only reason to do so would be to avoid a slow walk over a large zone, if that's a concern.

PR #62, opened 2025-09-11
Introduce policy for the key manager and forward the policy to keyset.

Philip-NLnetLabs:


src/policy/file/v1.rs
```diff
@@ -84,22 +87,139 @@ impl LoaderSpec {
...
+    auto_ksk: AutoConfig,
+    /// Configuration variable for automatic ZSK rolls.
+    auto_zsk: AutoConfig,
+    /// Configuration variable for automatic CSK rolls.
+    auto_csk: AutoConfig,
+    /// Configuration variable for automatic algorithm rolls.
+    auto_algorithm: AutoConfig,
+
+    /// DNSKEY signature inception offset (positive values are subtracted
+    ///from the current time).
```
ximon18, line 116:
```suggestion
    /// from the current time).
```

src/policy/file/v1.rs
```diff
@@ -84,22 +87,139 @@ impl LoaderSpec {
...
-pub struct KeyManagerSpec {}
+pub struct KeyManagerSpec {
+    /// Whether to use a CSK (if true) or a KSK and a ZSK.
+    use_csk: bool,
+
+    /// Algorithm and other parameters for key generation.
+    algorithm: KeyParameters,
+
+    /// Validity of KSKs.
+    ksk_validity: Option<u64>,
```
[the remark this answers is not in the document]
Philip-NLnetLabs: Do you have a suggestion? Normally I use Duration but that's unpleasant to type by hand in a toml file.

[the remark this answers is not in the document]
Philip-NLnetLabs: The signer policy has this:
    /// The offset for record signature inceptions, in seconds.
    pub sig_inception_offset: u64,

I can update the comment.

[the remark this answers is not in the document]
ximon18: That would be enough I think.

src/policy/mod.rs
```diff
@@ -307,3 +361,72 @@ pub struct ReviewPolicy {
...
+    /// Whether to handle the done step automatically.
+    pub done: bool,
+}
+
+//----------- DsAlgorithm -----------------------------------------------------
+
+/// The hash algorithm to use for DS records.
+#[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
+pub enum DsAlgorithm {
+    /// Hash the public key using SHA-256.
```
ximon18, line 418: Should we have a reference to why this is the default?

src/state/v1.rs
```diff
@@ -758,20 +760,107 @@ impl LoaderPolicySpec {
...
+    cds_remain_time: u64,
+
+    /// The DS hash algorithm.
+    ds_algorithm: DsAlgorithm,
+
+    /// The TTL to use when creating DNSKEY/CDS/CDNSKEY records.
+    default_ttl: Ttl,
+
+    /// Automatically remove keys that are no long in use.
+    autoremove: bool,
```
ximon18, line 812: Is that immediate, or after a predictable or unpredictable delay, or ...? Knot has `delete-delay: TIME` (it is a minimum it seems, and is 0 by default) and BIND9 has `purge-keys <duration>` (defaults to 90 days, is added to the "pre-publication interval"). Should we enhance the comment here to indicate when removal will happen? I don't mean to suggest that we should add support for a delay period btw, but should we track it as a future TODO?

src/units/key_manager.rs
```diff
@@ -97,11 +99,53 @@ impl KeyManager {
 
                 log::info!("Running {cmd:?}");
 
-                if let Err(e) = cmd.output() {
+                let output = cmd.output().map_err(|e| {
                     error!("[KM]: Error creating keyset: {e}");
+                    Terminated
+                })?;
+                if !output.status.success() {
+                    error!("[KM]: Create command failed with: {}", output.status);
```
ximon18, line 107: Should log statements here note which zone is concerned?

src/units/key_manager.rs
```diff
@@ -297,3 +353,103 @@ fn get_keyset_info(state_path: impl AsRef<Path>) -> KeySetInfo {
         retries: 0,
     }
 }
+
+fn policy_to_commands(center: &Center, zone_name: &Name<Bytes>) -> Vec<Vec<String>> {
+    // Ensure that the Mutexes are locked only in this block;
```
ximon18, line 358:
```suggestion
    // Ensure that the mutexes are locked only in this block;
```

ximon18 (approved):

src/units/key_manager.rs
Philip-NLnetLabs: I'll add the zone name.

src/state/v1.rs
Philip-NLnetLabs: I add a comment where the default is set to true.

src/policy/mod.rs
```diff
@@ -307,3 +361,72 @@ pub struct ReviewPolicy {
...
+
+//----------- KeyParameters ---------------------------------------------------
+
+#[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
+pub enum KeyParameters {
+    /// The RSASHA256 algorithm with the key length in bits.
+    RsaSha256(usize),
+    /// The RSASHA512 w algorithmith the key length in bits.
+    RsaSha512(usize),
+    /// The ECDSAP256SHA256 algorithm.
```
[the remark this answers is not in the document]
Philip-NLnetLabs: It is explicitly listed as the one to use in Section 3.2. I'll add a comment.

PR #63, opened 2025-09-11
Add CLI support for KMIP configuration.

ximon18:
- Adds two new state paths:
  - kmip_credentials_store_path (def: /var/lib/cascade/kmip/credentials.db)
  - kmip_server_state_dir (def: /var/lib/cascade/kmip/)
- Adds new CLI commands `hsm add`, `hsm list` and `hsm show`.
- Adds new API endpoints POST /kmip, GET /kmip and GET /kmip/<server_id>.
- Extends the POST /zone/add endpoint handler to accept an optional JSON `kmip_server_id` field.
- Extends the CLI Clap configuration with a `--kmip-server-id` arg to the `zone add` command.
- Modifies KeyManager handling of the AddZone event to check for kmip_server_id and, if set, invoke 'dnst keyset kmip add-server' between calls to 'create' and 'init'.

Still to do:
- Modify KMIP server (updating all dnst states for zones using the server).
- Delete KMIP server (updating all dnst states for zones using the server).
- Disable KMIP server?
- Add support for certificates.

Open questions:
- Naming: kmip, hsm, use which term when?
- TLS cert and key files: currently this PR reads them from files and sends them to the API as bytes. This allows Cascade to save them somewhere it definitely has access to, will work if the original files are removed, and will work if the CLI is run on a different machine than Cascade is running on. But maybe others think this should just try and instruct Cascade to use files at user specified locations and leave it at that. (though that would be inconsistent with policy which saves a Cascade state copy of a user policy file rather than use the file directly).


ximon18:
We're going to merge this as-is and add backlog items to track the missing pieces and address any issues found.

PR #66, opened 2025-09-16
Implement zone list and status properly (restoring some lost commits)

tertsdiepraam:


ximon18:
I'll just approve this on the same basis as you mentioned on #54: _"We can do more but I suggest that the merge this and I release the mutex lock on working on this again."_

ximon18:
Wait, I see mention of zone source, does this conflict with the changes made by @bal-e in https://github.com/NLnetLabs/cascade/pull/67?

PR #67, opened 2025-09-16
Enhance the zone loader

bal-e:
- XFRs use the source of the zone instead of a hard-coded address.

- The zone source is now stored in state and can be changed.

- The loader will react to changes in zone sources.

- The loader loads known zones at startup.

  This means Cascade can (for the most part) continue normal operation after a restart.  The key manager still needs work.

- The loader will react to zone removals.

- Proper support for TSIG-authenticated loading has been lost.  This will require better integration with the new TSIG key store, and some thought on how TSIG keys should be restored when loading from state.


src/units/zone_loader.rs
```diff
@@ -161,50 +141,37 @@ impl ZoneLoader {
...
-                                register.name
-                            );
-                        }
+                    ZoneLoadSource::Server { addr, tsig_key: _ } => {
+                        Self::register_secondary_zone(name.clone(), addr, zone_updated_tx)?
                     }
-                }
+                };
+
+                // TODO: Handle (or iron out) potential errors here.
```
ximon18, line 164: While an error shouldn't happen here, as it can only happen if inserting a zone that already exists and you just removed the zone, maybe it's worth logging something here if it happens as it is really not expected, AND would mean potentially that the zone wasn't added after all?

src/zone/state/v1.rs
```diff
@@ -388,3 +392,61 @@ impl ServerPolicySpec {
         Self {}
     }
 }
+
+//----------- LoadSourceSpec ---------------------------------------------------
+
+/// Where to load a zone from.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+pub enum LoadSourceSpec {
```
ximon18, line 401: Nit: Should this be `ZoneLoadSourceSpec`?

src/zone/state/v1.rs
```diff
@@ -202,14 +208,13 @@ impl SignerPolicySpec {
...
-#[derive(Copy, Clone, Debug, Default, Deserialize, Serialize)]
+#[derive(Copy, Clone, Debug, Deserialize, Serialize)]
 #[serde(rename_all = "kebab-case", deny_unknown_fields)]
 pub enum SignerSerialPolicySpec {
     /// Use the same serial number as the unsigned zone.
     Keep,
 
     /// Increment the serial number on every change.
-    #[default]
     Counter,
```
ximon18, line 218: Is the lost default here and on `Nsec3OptOutPolicy` set somewhere else, or lost entirely?

src/zone/state/v1.rs
```diff
@@ -388,3 +392,61 @@ impl ServerPolicySpec {
...
+        path: Box<Utf8Path>,
+    },
+
+    /// Load from a DNS server via XFR.
+    Server {
+        /// The TCP/UDP address of the server.
+        addr: SocketAddr,
+
+        /// The TSIG key to use, if any.
+        tsig_key: Option<Name<Bytes>>,
```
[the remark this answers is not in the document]
bal-e: I've left most of the infrastructure necessary to get TSIG working again when we have the time -- removing it would make things harder later.

src/zone/mod.rs
```diff
@@ -270,6 +299,49 @@ pub fn change_policy(
...
+        .update_tx
+        .send(Update::Changed(Change::ZoneSourceChanged(
+            name.clone(),
+            state.source.clone(),
+        )))
+        .unwrap();
+
+    zone.mark_dirty(&mut state, center);
+
+    log::info!("Set source of zone '{name}' to '{:?}'", state.source);
```
ximon18, line 341: Maybe log the change from the old value to the new value?

src/manager.rs
```diff
@@ -60,20 +60,12 @@ pub fn spawn(
     let zone_name =
         StoredName::from_str(&std::env::var("ZL_IN_ZONE").unwrap_or("example.com.".to_string()))
             .unwrap();
-    let xfr_in = std::env::var("ZL_XFR_IN").unwrap_or("127.0.0.1:8055 KEY sec1-key".into());
     let xfr_out = std::env::var("PS_XFR_OUT").unwrap_or("127.0.0.1:8055 KEY sec1-key".into());
```
ximon18, line 63: We should get rid of all the env var input handling, that's just a left over from the state of being temporarily without a config system. But I guess not in this PR.

ximon18 (approved):

src/manager.rs
bal-e: Yep, I only removed the variables we no longer need.  Feel free to address the zone server (which is the only remaining thing IIRC) in a separate PR.

PR #71, opened 2025-09-18
Have sensible default for DNSSEC parameters and document what is miss…

Philip-NLnetLabs:
…ing.

### DIFF
--- a/src/policy/file/v1.rs
+++ b/src/policy/file/v1.rs
@@ -6,8 +6,8 @@ use domain::base::Ttl;
 use serde::{Deserialize, Serialize};
 
 use crate::policy::{
-    KeyManagerPolicy, LoaderPolicy, Nsec3OptOutPolicy, PolicyVersion, ReviewPolicy, ServerPolicy,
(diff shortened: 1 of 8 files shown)


src/policy/file/v1.rs
```diff
@@ -195,38 +195,52 @@ impl KeyManagerSpec {
...
-            dnskey_signature_lifetime: FOUR_WEEKS,
-            dnskey_remain_time: FOUR_WEEKS / 2,
-            cds_inception_offset: ONE_DAY,
-            cds_signature_lifetime: FOUR_WEEKS,
-            cds_remain_time: FOUR_WEEKS / 2,
+            // The following have roughly the same defaults as used for
+            // signing the zone.
+            dnskey_inception_offset: 24 * 3600,
+            dnskey_signature_lifetime: 14 * 24 * 3600,
+            dnskey_remain_time: 14 * 24 * 3600 / 2,
```
ximon18, line 224: Is this trying to say half of dnskey_signature_lifetime? That might be better expressed through a couple of related constants that can't accidentally diverge from one another.

PR #75, opened 2025-09-22
More zone status update CLI command output.

ximon18:
See: https://github.com/NLnetLabs/cascade/issues/69

Outputs something like the following for a local zonefile.

```
# cascade zone status example.com
Zone: example.com
Status: At the published pipeline stage 
Policy: dummy
Latest input:
  Serial: 1
  Loaded from the zonefile '/etc/cascade/medium_example.com'
  Received 29252 bytes at 2025-09-23T08:09:02+00:00 in 0s
  Configured for automatic review by '/etc/cascade/approve_or_deny.sh'
  Waiting for zone reload command to receive changes
  Unsigned zone available on 127.0.0.1:9051
Latest output:
  Serial: 2025092301
  Signed at 2025-09-23T08:09:10+00:00
  Signed 998 records in 1s
  Signed zone available on 127.0.0.1:9052
  Not configured for review
  Published zone available on 127.0.0.1:9053
  Re-signing scheduled at ? (in ?)
DNSSEC keys:
  KSK tagged 56940:
    Reference: file:///var/lib/cascade/keys/Kexample.com.+013+56940.key
    Actively used for signing
  ZSK tagged 61464:
    Reference: file:///var/lib/cascade/keys/Kexample.com.+013+61464.key
    Actively used for signing
  Details:
    AlgorithmRoll: Propagation1
    Check that the following RRset has propagated to all name servers:
    example.com. 3600 IN DNSKEY 257 3 13 aHA8V1KdCN2qCLLZirvBEe4AxjHlm0nUpbx5lIQb72FEhXQYTjpl+xxyVsM4t0oIEMyAnE73ABNGKNugOIgdbw==
    example.com. 3600 IN DNSKEY 256 3 13 oXIb2C1bLk7sUzkAz0RV7tZPV0z07APUne0F80G7TCOF32fxA2U9ASQsDrm5yuYqn49s1eZ/RST69S7SQXyMFA==
    example.com. 3600 IN RRSIG DNSKEY 13 2 3600 1759824542 1758528542 56940 example.com. 3W1XDzI+pwZvgdISZnLy8LrGm4brzDEcVv2ZB0fzw6iN9VNmkJ+F4CqhNM1vfrucnVnHf2Jwvo7Gsux3xq6pRQ==
    
    Check that all authoritative records in the zone have been signed with the following Key(s) and that all nameservers of the zone serve that version or later:
[...]

Issue #69, opened 2025-09-16
Improve CLI zone status feedback


tertsdiepraam:
An early proposed UX design contained this proposed feedback to the user:
```
Status: waiting for changes
Latest input:
  Serial 2025011654
  Loaded from the zonefile '/var/run/zone/example.org.zone'
  Loaded at 2025-01-16 13:45:37 UTC (2 minutes ago)
  Not configured for review
  Watching the zonefile '/var/run/zone/example.org.zone' for changes
Latest output:
  Serial 2025011655 (matching input serial 2025011654)
  Signed at 2025-01-16 13:45:51 UTC (2 minutes ago)
  Available on dns://127.0.0.1:7858/example.org.
  Not configured for review
  Re-signing scheduled at 2025-04-16 13:45:51 (in 3 months)
DNSSEC keys:
  ZSK tagged 34791:
    Stored on disk at '/var/key/example.org.34791.key'
    Actively used for signing the zone
      Next rollover scheduled at 2025-07-16 13:45:40 UTC (in 6 months)
  KSK tagged 21737:
      Stored on disk at '/var/key/example.org.21737.key'
      Actively used for signing the DNSKEY record
      Next rollover scheduled at 2026-01-16 13:45:38 UTC (in 1 year)
```

Other notes:
- Any issues loading/receiving XFR?
- What stage is the zone at? loading/receiving/pending unsigned review/rejected at unsigned review/pending signing/pending signed review/rejected at signed review/published? 
- When will the first signatures expire?
- Is it being slow or fast to load/sign?
- What will happen next and in the future?

Philip-NLnetLabs:
When the first signatures expire is now tracked, is that is part of the state. Together a with a remain time (I don't know if that is already in policy) the next resign time can be computed.

I propose that we include the dnst keyset status -v for now. It's not pretty but very detailed. And it will be easier to tweak then to recreate entirely in cascade.

ximon18:
Some quick thoughts I had about this, but still need to think about this more and do more research as to what others show.

Pipeline
	•	Last / shortest / longest / total last / total shortest / total longest durations for a given zone at each stage: receiving, approving, queueing to sign, signing, signed approving, duration from published to fetched

Per zone
Past
	•	Last notify timestamp and origin
	•	Last source update success/failure
	•	Source byte size
	•	Source receive duration
	•	Source receive type: file, AXFR or IXFR
	•	Last unsigned / signed / published zone non-XFR query origin and timestamp
[...]

PR #80, opened 2025-09-22
Add manual key rolls to cascade CLI

mozzieongit:
Closes: #79

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -244,3 +244,88 @@ pub enum PolicyChange {
     Updated,
     Unchanged,
 }
+
+//------------ KeySet API Types ----------------------------------------------
+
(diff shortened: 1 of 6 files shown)


ximon18:
> I just thought, should the output of `cache-expired1` (in particular, the new DS record) be forwarded to the cascade output so that a user doesn't need to call status afterward to get the new DS RR? However, the current implementation is that the propagation/cache transitions are done automatically, even if started manually by the CLI. There seems to be no way to extract the DS record using the CLI right now. Is that implemented somewhere else?

This was mentioned on:
- https://github.com/NLnetLabs/cascade/issues/69#issuecomment-3311124955
- https://github.com/NLnetLabs/cascade/pull/75

PR #83, opened 2025-09-24
Key imports

tertsdiepraam:
~THIS IS ENTIRELY UNTESTED~
THIS IS NOT ENTIRELY UNTESTED ANYMORE

See the first commit for the actual changes and the second for some refactoring in another function that should hopefully not change any functionality.

I think this mostly works. The error handling is bad (just logging errors and ignoring them). But if you give a CSK it seems to work. There is an issue that the key manager seems to run _after_ signing though, that might lead to some weird things. I think we should fix that elsewhere though.

### DIFF
--- a/src/cli/commands/zone.rs
+++ b/src/cli/commands/zone.rs
@@ -732,7 +732,7 @@ fn kmip_imports(key_type: KeyType, x: &[String]) -> Vec<KeyImport> {
         .into_iter()
         .map(|chunk| {
             let [server, public_id, private_id, algorithm, flags] = chunk else {
-                panic!()
+                unreachable!()
             };
             KeyImport::Kmip(KmipKeyImport {
                 key_type,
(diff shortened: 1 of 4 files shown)


src/cli/commands/zone.rs
```diff
@@ -29,6 +33,27 @@ pub enum ZoneCommand {
...
+        #[arg(long = "import-zsk-file")]
+        import_zsk_file: Vec<Utf8PathBuf>,
+
+        #[arg(long = "import-csk-file")]
+        import_csk_file: Vec<Utf8PathBuf>,
+
+        #[arg(long = "import-ksk-kmip", value_names = ["server", "public_id", "private_id", "algorithm", "flags"])]
+        import_ksk_kmip: Vec<String>,
+
+        #[arg(long = "import-zsk-kmip", value_names = ["server", "public_id", "private_id", "algorithm", "flags"])]
```
ximon18, line 52: Ooo, I learned a thing about Clap.

ximon18 (approved):
LGTM, just the one question about whether we can combine separate file and KMIP inputs.

PR #84, opened 2025-09-24
Add rudimentary zone reload command

mozzieongit:
For zone files, this copies the approach of the changed source command, in that it removed and re-adds the zone. 
For XFR is triggers the RefreshZone AppCommand, which only fetches a new version if the SOA serial is newer.
This approach of course doesn't consider active signing pipelines and whatnot, but that should be fine for the demo, right?


mozzieongit:
TODO: trigger RefreshZone instead of re-adding the zone for XFR (see comments on #85)

PR #87, opened 2025-09-26
Add template config/policy file generation command

mozzieongit:
How about this for a start?

Then if we change the policy creation experience we might reconsider having the policy template here, but as long as the user has to create those files themself, this would probably be a good way of providing the templates.

- [ ] Once there is an example policy file, this can be included as well.

### DIFF
--- a/src/cli/commands/mod.rs
+++ b/src/cli/commands/mod.rs
@@ -4,6 +4,7 @@ pub mod hsm;
 pub mod keyset;
 pub mod policy;
 pub mod status;
+pub mod template;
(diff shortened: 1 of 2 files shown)


bal-e (approved):
This looks great!  A really simple way to help users with templating.

Is there a good way to tell the user we're about to print the file to standard output?  Maybe say "Print" instead of "Generate" in the docs?

mozzieongit:
> Keep in mind the template paths are changing in #90.

Will wait for that to merge, then update this PR

PR #94, opened 2025-09-29
WIP docs (including Quick Start)

mozzieongit:
So that Ximon and I don't do duplicate work


doc/manual/source/index.rst
```diff
@@ -27,14 +27,18 @@ Open-source with professional support services
    agreement. Cascade is liberally licensed under the `BSD 3-Clause license
    <https://github.com/NLnetLabs/cascade/blob/main/LICENSE>`_.
 
+Cascade is ONLY a DNSSEC signing solution and not a complete primary name
+server. To read what that entails, read the :doc:`limitations <limitations>`
```
mozzieongit, line 31: This will instead reference Ximons `before-you-start` doc

doc/manual/source/quick-start.rst
```diff
@@ -0,0 +1,104 @@
...
+
+           sudo systemctl edit cascaded.socket
+
+        and insert the following config:
+
+        .. code-block:: text
+
+           [Socket]
+           ListenDatagram=<your-ip>:53
+           ListenStream=<your-ip>:53
```
mozzieongit, line 32: This assumes #93 merged

doc/manual/source/quick-start.rst
```diff
@@ -0,0 +1,106 @@
+Quick Start
+============
+
+After :doc:`installing <installation>` Cascade you can immediately start using
+it, if you only need it to listen on localhost. If you need to be able to
+fetch the signed zones from other servers, you need to configure its
+listening addresses.
+
+Configuration the listen addresses
```
ximon18, line 9:
```suggestion
Configuring the listen addresses
```

doc/manual/source/quick-start.rst
```diff
@@ -0,0 +1,106 @@
...
+
+   .. group-tab:: Without systemd
+
+        When using Cascade without systemd, you need to configure the listen
+        address in Cascade's ``config.toml`` in the ``[servers]`` section:
+
+        .. code-block:: text
+
+            [server]
+            servers = ["127.0.0.1:53", "<your-ip>:53"]
```
ximon18, line 74: Do we need localhost here?

```suggestion
            servers = ["<your-ip>:53"]
```

doc/manual/source/quick-start.rst
```diff
@@ -0,0 +1,106 @@
...
+available to your public primaries, you need to add the required IP addresses
+to Cascade's listen addresses.
+
+.. tabs::
+
+   .. group-tab:: Using systemd
+
+        On a system with systemd, you need to override the
+        ``cascaded.socket``, as Cascade will only listen on localhost, by
+        default. To add your own listeners, use the following command:
```
ximon18, line 22: Too many commas.
```suggestion
        On systems using systemd to run Cascade the systemd files included in our provided packages use the systemd socket activation feature to bind to listen addresses on behalf of Cascade. If you wish to change the addresses bound to you will need to edit the ``cascaded.socket`` systemd file. One way to do this is to use the `systemctl edit` command like so:
```

doc/manual/source/quick-start.rst
```diff
@@ -0,0 +1,106 @@
...
+        if you wish to disable the listen address for localhost, you'll need
+        to replace the top line (``[Socket]``) with the following (adding
+        ``ListenStream=`` resets all previously defined ``ListenStream`` and
+        ``ListenDatagram`` settings):
+
+        .. code-block:: text
+
+           [Socket]
+           #ListenStream=
+           # ... the other ListenDatagram/ListenStream settings
```
ximon18, line 45:
```suggestion
           [Socket]
           # Uncomment the next line if you wish to disable listening on localhost.
           #ListenStream=
           ListenDatagram=<your-ip>:53
           ListenStream=<your-ip>:53
```

doc/manual/source/quick-start.rst
```diff
@@ -0,0 +1,106 @@
...
+            servers = ["127.0.0.1:53", "<your-ip>:53"]
+
+        Then you can start Cascade with (replace the config and state path
+        with your appropriate values):
+
+        .. code-block:: bash
+
+            sudo cascade --config /etc/cascade/config.toml --state /var/lib/cascade/state.db
+
+Adding an unsigned zone to sign
```
ximon18, line 83: ?

```suggestion
Signing your first zone
```

doc/manual/source/quick-start.rst
```diff
@@ -0,0 +1,106 @@
...
+-------------------------------
+
+After configuring Cascade, you can begin adding zones. Cascade supports zones
+sourced from a local file or fetched from another name server using XFR.
+
+However, before adding a zone you need to create a zone policy to use for the
```
ximon18, line 89: This would be easier if we had a default policy supplied with the packages, then we can just say the zone you are adding will use the default policy.

doc/manual/source/quick-start.rst
```diff
@@ -0,0 +1,106 @@
...
+it, if you only need it to listen on localhost. If you need to be able to
+fetch the signed zones from other servers, you need to configure its
+listening addresses.
+
+Configuration the listen addresses
+----------------------------------
+
+Cascade only listens on localhost, by default. To make your signed zones
+available to your public primaries, you need to add the required IP addresses
+to Cascade's listen addresses.
```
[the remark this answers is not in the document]
mozzieongit: Maybe like ca6406bb and just remove all mention of anything else?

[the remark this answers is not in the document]
ximon18: I don't recall what happened with XFR ACL policy, should check.

doc/manual/source/quick-start.rst
```diff
@@ -0,0 +1,89 @@
...
+
+
+Signing your first zone
+-------------------------------
+
+After configuring Cascade, you can begin adding zones. Cascade supports zones
+sourced from a local file or fetched from another name server using XFR.
+
+However, before adding a zone you need to create a zone policy to use for the
+zone (you can choose your own name instead of ``default``):
```
ximon18, line 72: How about this?

```suggestion
Zones take a lot of their settings from policy.

Policies allow easy re-use of settings across multiple zones and control things like whether or not zones should be reviewed and how, what DNSSEC settings should be used to sign the zone, and more.

Adding a policy is done by creating a file. To make it easy to get started we provide a default policy template so we'll use that to create a policy for our zone to use.

The name of the policy is taken from the filename. The directory to save the policy file to is determined by the ``policy-dir`` setting in configured in ``/etc/cascade/config.toml``. The filename can be any valid filename and will be used as the name of the policy.

In the example below the `sudo tee` command is needed because the default policy directory is not writable by the current user.
```

doc/manual/source/quick-start.rst
```diff
@@ -0,0 +1,89 @@
...
+   cascade template policy | sudo tee /etc/cascade/policies/default.toml
+   cascade policy reload
+
+Then, to add a zone use:
+
+.. code-block:: bash
+
+   cascade zone add --source <file-path|ip-address> --policy default <zone-name>
+
+Now, your zone will be picked up by Cascade, keys prepared, and the signing
```
ximon18, line 85: I suggest at this point showing `cascade zone status <zone-name>`.

doc/manual/source/quick-start.rst
```diff
@@ -0,0 +1,89 @@
...
+
+After :doc:`installing <installation>` Cascade you can immediately start using
+it, unless you need to adjust the addresses it listens on or need to modify
+the settings relating to daemonization.
+
+Configuring the listen addresses
+----------------------------------
+
+By default, Cascade only listens on localhost. If you want Cascade to listen
+on other addresses too, you need to configure them as shown below:
```
ximon18, line 12: If we write the following, the next bit that says "After configuring Cascade" makes more sense. What do you think?

```suggestion
Configuring Cascade
---------------------

By default, Cascade only listens on the localhost address. If you want Cascade to listen
on other addresses too, you need to configure them.

The ``/etc/cascade/config.toml`` file controls listen addresses, which filesystem paths Cascade uses, daemonizatioan settings (running in the background, running aa a different user, and log settings.

If using systemd to run Cascade some of these settings should be ignored and systemd features used instead.
```

doc/manual/source/quick-start.rst
```diff
@@ -0,0 +1,137 @@
...
+
+In the example below the `sudo tee` command is needed because the default
+policy directory is not writable by the current user.
+
+.. code-block:: bash
+
+   cascade template policy | sudo tee /etc/cascade/policies/default.toml
+   cascade policy reload
+
+Then, to add a zone use:
```
ximon18, line 101: As the policy topic grew, perhaps:

```suggestion
Defining policy
---------------

After configuring Cascade, you can begin adding zones. Cascade supports zones
sourced from a local file or fetched from another name server using XFR.

Zones take a lot of their settings from policy.

Policies allow easy re-use of settings across multiple zones and control
things like whether or not zones should be reviewed and how, what DNSSEC
settings should be used to sign the zone, and more.

Adding a policy is done by creating a file. To make it easy to get started we
provide a default policy template so we'll use that to create a policy for our
zone to use.

The name of the policy is taken from the filename. The directory to save the
policy file to is determined by the ``policy-dir`` setting as configured in
``/etc/cascade/config.toml``. The filename can be any valid filename and will
be used as the name of the policy.

In the example below the `sudo tee` command is needed because the default
policy directory is not writable by the current user.

.. code-block:: bash

   cascade template policy | sudo tee /etc/cascade/policies/default.toml
   cascade policy reload

Signing your first zone
-----------------------

Adding a zone to Cascade will cause Cascade to attempt to load, sign and publish it. If you configured review hooks they will be executed (and may intentionally prevent your zone reaching publication).

To add a zone use:
```

doc/manual/source/architecture.rst
```diff
@@ -20,25 +20,51 @@ stages:
    :align: center
    :alt: Cascade pipeline
 
+Robustness
+----------
+
+Cascade is written in the Rust programming language making it significantly
+less likely to crash or suffer from memory safety issues, and at the same time
+making it easier to leverage the higher core count of modern computers via
+Rusts' "fearless concurrency" when needed.
```
mozzieongit, line 29:
```suggestion
Rust's "fearless concurrency" when needed.
```

doc/manual/source/architecture.rst
```diff
@@ -20,25 +20,51 @@ stages:
...
+
+Cascade outsources PKCS#11 module loading to a separate :program:`kmip2pkcs11`
+daemon to avoid running untrusted 3rd party code inside the main Cascade
+process also thereby eliminating another source of potentially instability and
```
mozzieongit, line 33:
```suggestion
process also thereby eliminating another source of potential instability and
```

doc/manual/source/architecture.rst
```diff
@@ -20,25 +20,51 @@ stages:
    :align: center
    :alt: Cascade pipeline
 
+Robustness
+----------
+
+Cascade is written in the Rust programming language making it significantly
+less likely to crash or suffer from memory safety issues, and at the same time
```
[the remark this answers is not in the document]
ximon18: I don't consider a panic to be a crash, I'm referring more to SEGFAULT when I say crash. And yes the text is from a post production release point of view.

PR #97, opened 2025-09-30
Fix signing failure due to attempt to sign before `dnst keyset init` has been called.

ximon18:
- Do `dnst keyset create` ... `dnst keyset init` first, then only add the zone if that succeeds.
- Don't do too much in the HTTP server, logic should be done elsehwere, so moves sending of `ApplicationCommand::RegisterZone` into `fn add_zone()` (so that adding a zone from anywhere makes sure this gets done).
- Factor out code from `fn run_cmd()` to make it easier to see which cases `run_cmd()` handles and to make error conversion easier.
- Also removes an unnecessary state lock acquisition and unwrap().
- Handle file open error instead of unwrap().


src/units/key_manager.rs
```diff
@@ -295,24 +197,151 @@ impl KeyManager {
...
+                max_response_bytes,
+                key_label_prefix,
+                key_label_max_bytes,
+                has_credentials,
+            } = kmip_server;
+
+            let mut cmd = self.keyset_cmd(name.clone());
+
+            // TODO: This command should get issued _after_ keyset create
+            // but _before_ keyset init, both of which are issued above.
```
tertsdiepraam, line 285: This TODO is now fixed right?

src/units/key_manager.rs
```diff
@@ -295,24 +197,151 @@ impl KeyManager {
...
+                cmd.arg("--key-label-prefix").arg(key_label_prefix);
+            }
+
+            // TODO: --client-cert, --client-key, --server-cert and --ca-cert
+            cmd.output()
+                .map_err(|err| ZoneAddError::Other(err.reason))?;
         }
+
+        // Set config
+        let config_commands = policy_to_commands(&policy);
```
tertsdiepraam, line 322: This will conflict with the key imports. Let me know when I need to update that or if you want to merge key imports first.

src/center.rs
```diff
@@ -61,12 +61,14 @@ pub struct Center {
...
 
 /// Add a zone.
-pub fn add_zone(
+pub async fn add_zone(
     center: &Arc<Center>,
     name: Name<Bytes>,
     policy: Box<str>,
     source: api::ZoneSource,
 ) -> Result<(), ZoneAddError> {
+    register_zone(center, name.clone(), policy.clone()).await?;
```
tertsdiepraam, line 70: Hmm, yeah I guess this makes sense. I somehow didn't realize we could just make the keys before making the zone. Nice!

ximon18: I honestly don't know what the best order is, there just things that need to happen before a zone can be used, but I think we should also roll back partial changes if they don't all succeed and we lack that at present. There's also an overlap with the `zone enable` feature added by @bal-e in the `demo` branch yesterday that needs some consideration how to merge the two things.

ximon18:
> Code wise LGTM, haven't tested it

Lacks tests but ... yeah we know we need to do something about that, we'll have time soon I hope.

For me at least I have not seen the ordering issue since using this branch, but that's not proof of course.

PR #107, opened 2025-10-01
Overhaul how zones are reviewed

bal-e:
- Zone review no longer involves approval tokens.

- Added commands `cascade zone {approve,reject}` for manually approving
  or rejecting zones.

- Rely on the exit status of hook commands to judge approval/rejection.

### DIFF
--- a/src/targets/central_command.rs
+++ b/src/targets/central_command.rs
@@ -6,7 +6,7 @@ use domain::zonetree::StoredName;
 use log::info;
 use tokio::sync::mpsc;
 
-use crate::api::ZoneReviewStatus;
+use crate::api::{self, ZoneReviewStatus};
 use crate::center::{get_zone, halt_zone, Center, Change};
 use crate::comms::{ApplicationCommand, Terminated};
 use crate::manager::TargetCommand;
@@ -126,6 +126,25 @@ impl CentralCommand {
                     zone_name,
                     source,
                     serial,
+                },
+            ),
+
+            Update::ReviewZone {
+                name,
+                stage,
+                serial,
+                decision,
+            } => (
+                "Passing back zone review",
+                match stage {
+                    api::ZoneReviewStage::Unsigned => "RS",
+                    api::ZoneReviewStage::Signed => "RS2",
+                },
+                ApplicationCommand::ReviewZone {
+                    name,
+                    serial,
+                    decision,
+                    tx: tokio::sync::oneshot::channel().0,
                 },
             ),
 
--- a/src/units/zone_server.rs
+++ b/src/units/zone_server.rs
@@ -4,8 +4,6 @@ use std::collections::HashMap;
 use std::marker::Sync;
 use std::net::IpAddr;
 use std::pin::Pin;
-use std::process::Command;
-use std::str::FromStr;
 use std::sync::{Arc, Mutex};
 use std::time::Instant;
 
@@ -35,15 +33,13 @@ use domain::zonetree::types::EmptyZoneDiff;
 use domain::zonetree::Answer;
 use domain::zonetree::{StoredName, ZoneTree};
 use futures::Future;
-use log::warn;
 use log::{debug, error, info, trace};
 use serde::Deserialize;
-use tokio::sync::mpsc::{self, Sender};
-use tokio::sync::{oneshot, RwLock};
+use tokio::sync::{mpsc, oneshot, RwLock};
 #[cfg(feature = "tls")]
 use tokio_rustls::rustls::ServerConfig;
-use uuid::Uuid;
 
+use crate::api;
 use crate::center::{get_zone, Center};
 use crate::common::tsig::TsigKeyStore;
 use crate::comms::ApplicationCommand;
@@ -80,9 +76,6 @@ pub enum Source {
 #[derive(Debug)]
 pub struct ZoneServerUnit {
     pub center: Arc<Center>,
-
-    /// The relative path at which we should listen for HTTP query API requests
-    pub http_api_path: Arc<String>,
 
     /// XFR out per zone: Allow XFR to, and when with a port also send NOTIFY to.
     pub _xfr_out: HashMap<StoredName, String>,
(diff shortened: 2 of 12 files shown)


src/targets/central_command.rs
```diff
@@ -129,6 +129,25 @@ impl CentralCommand {
...
+                "Passing back zone review",
+                match stage {
+                    api::ZoneReviewStage::Unsigned => "RS",
+                    api::ZoneReviewStage::Signed => "RS2",
+                },
+                ApplicationCommand::ReviewZone {
+                    name,
+                    serial,
+                    decision,
+                    tx: tokio::sync::oneshot::channel().0,
```
ximon18, line 147: Why does immediately discard the receiver, won't that close the channel and invalidate the sender?

ximon18 (commented):
Looks good except for the confusing immediately dropped one shot rx, and there's code duplication that normally we would say should be cleaned up, but for this alpha version that will be okay for now. Let me try it out locally before I approve it.

ximon18:
Does `CASCADE_CONTROL` need removing from the policy template comments?

ximon18:
> In my policy I have:
> 
> ```toml
> [loader.review]
> #required = true
> required = false
> cmd-hook = "/etc/cascade/approve_or_deny.sh"
> ```
> 
> But `cascade policy show <policy name>` shows:
> 
> ```
> # cascade policy show dummy
> dummy:
>   zones: <none>
>   loader:
>     review:
>       required: false
>       cmd_hook: /etc/cascade/approve_or_deny.sh
>   key manager:
>     hsm server: <none>
>   signer:
>     serial policy: date counter
>     signature inception offset: 86400 seconds
>     signature validity offset: 1209600 seconds
>     denial: NSEC3 (opt-out: enabled)
>     review:
>       required: false
>       cmd_hook: <none>
>   server:
>     outbound:
>       accept XFR requests from:
>       send NOTIFY to:
> ```
> 
> Notice `cmd_hook` says "" while my policy file sets a path for the hook?

Ah, my bad, I see that's the signed review hook and the unsigned review hook is correctly showing the path.

mozzieongit (changes requested):

mozzieongit:
> Does CASCADE_CONTROL need removing from the policy template comments?

Yes, it should go. Users have no need to use the HTTP endpoint directly.

src/targets/central_command.rs
bal-e: This is used as a hack by the zone server to send itself a zone review decision.  There's nobody waiting on the result, so there need not be a receiver.  The sending side will not crash if the receiver is missing.

src/units/zone_server.rs
```diff
@@ -520,21 +502,15 @@ impl ZoneServer {
...
-            .entry((zone_name.clone(), zone_serial))
-            .and_modify(|e| e.push(approval_token))
-            .or_insert(vec![approval_token]);
+            .insert((zone_name.clone(), zone_serial));
 
         let Some(hook) = review.cmd_hook else {
             info!("[{unit_name}] No review hook set; waiting for manual review");
-            info!("[{unit_name}]: Confirm with HTTP GET {}approve/{approval_token}?zone={zone_name}&serial={zone_serial}", self.http_api_path);
-            info!("[{unit_name}]: Reject with HTTP GET {}reject/{approval_token}?zone={zone_name}&serial={zone_serial}", self.http_api_path);
+            info!("[{unit_name}]: Approve with HTTP POST /zone/{zone_name}/{zone_type}/{zone_serial}/approve");
```
ximon18, line 512: Is this guidance correct? Do we want people approving by using the otherwise undocumented and unofficial, internal, HTTP API? Don't we want to guide people to do `cascade zone approve` or whatever instead?

ximon18 (commented):
This PR should also update `approve_or_deny.sh` and `approve_or_deny_signed.sh` in this repository.

etc/policy.template.toml
```diff
@@ -61,19 +61,10 @@ required = false
...
-# To approve the zone contents, send an HTTP GET request to
-# `http://$CASCADE_CONTROL/approve/$CASCADE_TOKEN?zone=$CASCADE_ZONE&serial=$CASCADE_SERIAL`.
-# `approve` can be replaced by `reject` to reject the zone.  If the command
-# exits without approving or rejecting the zone, Cascade will wait for the user
-# to approve it manually.
-#
-# TODO: Use the exit code of the hook to infer approval / rejection?
+# accessible to Cascade (i.e. after it has dropped privileges).  Its exit code
+# will determine whether the zone is approved or not.
 #cmd-hook = "review-unsigned-zone.sh"
```
ximon18, line 68: I'm not realizing that `cmd-hook` as a name is a bit weird. It's a hook or a command but a bit redundant to call it both, while also "cmd" is a bit vague. Maybe we should rename this to `review-script-path`?

src/units/zone_server.rs
ximon18: I changed this to show the `cascade zone approve` and `cascade zone reject` commands that an operator would need to use.

src/units/zone_server.rs
```diff
@@ -545,34 +521,56 @@ impl ZoneServer {
...
-        match Command::new("sh")
+        match tokio::process::Command::new("sh")
             .args(["-c", &hook])
             .envs([
                 ("CASCADE_ZONE", &*zone_name.to_string()),
                 ("CASCADE_SERIAL", &*zone_serial.to_string()),
-                ("CASCADE_TOKEN", &*approval_token.to_string()),
                 ("CASCADE_SERVER", &*review_server.addr().to_string()),
                 (
                     "CASCADE_CONTROL",
```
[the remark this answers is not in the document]
ximon18: I removed the CASCADE_CONTROL env var as scripts now approve or reject by exit code.

etc/policy.template.toml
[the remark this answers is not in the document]
ximon18: Done.

ximon18:
> This PR should also update `approve_or_deny.sh` and `approve_or_deny_signed.sh` in this repository.

Done.

PR #113, opened 2025-10-02
Fix missing kmip state dir by pre-creating all state dirs on startup.

ximon18:


src/main.rs
```diff
@@ -73,16 +73,26 @@ fn main() -> ExitCode {
...
-                "Unable to create policy directory '{}': {e}",
-                &state.config.policy_dir
-            );
-            return ExitCode::FAILURE;
-        };
+        // Create required subdirectories (and their parents) if they don't
+        // exist. This is only needed for directories to which we write files
+        // without using util::write_file() as that function creates the
+        // directory (and parent directories) if missing. However, do it for
+        // all state directories now so that we don't discover only later that
```
mozzieongit, line 80: We could then also just do it for the keys directory too and not wait for keyset to create it

src/main.rs
```diff
@@ -73,16 +73,26 @@ fn main() -> ExitCode {
...
+        // we can't create the directory.
+        // TODO: Once we implement live config reloading, this should move
+        // somewhere else to also create the directories as specified in a the
+        // reloaded config.
+        for dir in [
+            &*state.config.kmip_credentials_store_path,
```
mozzieongit, line 86: This will create a directory called `/var/lib/cascade/kmip/credentials.db` (or whatever file name configured). This should be the parent directory to `state.config.kmip_credentials_store_path`.

mozzieongit (changes requested):

src/main.rs
```diff
@@ -83,7 +83,8 @@ fn main() -> ExitCode {
         // somewhere else to also create the directories as specified in a the
         // reloaded config.
         for dir in [
-            &*state.config.kmip_credentials_store_path,
+            &*state.config.keys_dir,
+            state.config.kmip_credentials_store_path.parent().unwrap(),
```
mozzieongit, line 87: The unwrap will panic if `kmip_credentials_store_path = "/filename.db"`. Unlikely to happen, though.

mozzieongit (approved):

src/main.rs
ximon18: Yeah, I figured I'd just leave that for now.

PR #114, opened 2025-10-02
Add more context on KMIP connection & state saving errors.

ximon18:
The original error that triggered this issue now includes context and that context is now reported back to the CLI user, e.g.:

```
[ximon@secondary4nl ~]$ ./cascade hsm add --insecure hsmrelay 127.0.0.1 --port 5696 --username OpenDNSSEC --password 1234 
[2025-10-02T19:12:12.925Z] ERROR cascade: Error: Add KMIP server command failed: unable to open KMIP credentials file /home/ximon/cascade_dir/kmip/credentials.db in create-read-write mode: No such file or directory (os error 2)
```

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -515,7 +515,7 @@ impl std::fmt::Display for HsmServerAddError {
                 err,
             } => write!(
                 f,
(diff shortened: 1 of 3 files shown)


mozzieongit (approved):
Just one missing single quote

PR #118, opened 2025-10-03
Fix zone remove after restart

mozzieongit:
Zone's were never added to the `policy.zones` map on start-up.

### DIFF
--- a/src/zone/state/mod.rs
+++ b/src/zone/state/mod.rs
@@ -94,6 +94,14 @@ impl Spec {
                 state.min_expiration = min_expiration;
                 state.next_min_expiration = next_min_expiration;
                 state.history = history;
+
+                // This should always be some at this stage...
+                if let Some(ref policy) = state.policy {
+                    let p = policies
+                        .get_mut(&*policy.name)
+                        .expect("zone policy references should not be kept around");
+                    p.zones.insert(zone.name.clone());
+                }
             }
         }
     }


mozzieongit:
> Should they be added on startup? 

In the current architecture they need to be added on startup when reading in state. (reason follows)

> Or is this a link to be established during runtime when a change is made? 

It is, when adding a zone during runtime. However, this PR is about continuing from existing state and therefore from startup.

> Or is a link even needed, a zone presumably knows which policy it is using and the reverse mapping can be looked up, except maybe for very large numbers of zones but then we're into the realm of caching.

Preferably, it would be enough for the zone to know which policy applies to it. Unfortunately, the amount of state duplication in the current approach means that zones also are added to the policy instances, as seen in

[`cascade::policy::Policy.zones`](https://github.com/NLnetLabs/cascade/blob/main/src/policy/mod.rs#L36)
```
pub struct Policy {
// ...

    /// The zones using this policy.
    pub zones: foldhash::HashSet<Name<Bytes>>,
}
```
and [`cascade::center::add_zone`](https://github.com/NLnetLabs/cascade/blob/main/src/center.rs#L96)
```
        zone_state.policy = Some(policy.latest.clone());
        policy.zones.insert(name.clone());
```
and deleting a zone requires [deleting it from the Policy.zones HashSet](https://github.com/NLnetLabs/cascade/blob/efe045d10a5c167c01b861d086cdb151223542b0/src/center.rs#L160)
```
assert!(policy.zones.remove(&name), "zone policies are consistent");
```

ximon18 (approved):
I'm not sure if this is really the right fix or not but you've looked at it more closely and can defend your chosen solution, plus I just hit this problem in local testing and this branch allowed me to remove the zone whereas without this PR I could not remove the zone.

PR #125, opened 2025-10-04
Change default ports to 4539-4543

mozzieongit:
Unassigned port range: https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=4539-4544


ximon18:
@mozzieongit: What makes you prefer those numbers over the current numbers?

mozzieongit:
Initial reason to select different numbers was that we had the TODO left in the config files.
Another reason is that the 8000-range is pretty crowded with services and "temporary" ports.

Then, I just looked at lower thousands, so just replacing the 8xxx with 1xxx (there was something with dynamic DNS, so I decided against), and then 2xxx.

ximon18:
Alternate random idea: 25107x, for the alpha launch date ;-)

mozzieongit:
I can't tell if you're joking or not

PR #127, opened 2025-10-05
Handle signing failures better.

ximon18:
Internally the signer tracks queued, in progress and finished signing operations, but (a) on failure to sign would leave the queue entry at "requested" or "in-progress" rather than setting it to done (failed), as it only had support for done (succeeded).

This PR adds done (failed), makes sure the queue entry gets set to done (failed) if signing fails, and reports the state better via the `cascade zone status` command.

This PR also:
- Replaces two calls to `unreachable!()` with error handling as it turns out they are not unreachable...
- Logs what the signer is doing to serial numbers and which serial policy it is using.
- Logs the content of the signing queue. (this should be exposed via the CLI but this is better than nothing).


ximon18:
The code in this PR and https://github.com/NLnetLabs/cascade/pull/129 have evolved a bit further as part of integration into the `alpha` branch (https://github.com/NLnetLabs/cascade/pull/136). It's quite tricky to back port the changes to the original PRs. The extra changes mainly concern introducing a SignerError return error type, otherwise the essence of this PR is contained in the `alpha` branch. So I'll keep this PR here temporarily for discussion but the code we should use is in the `alpha` branch.

ximon18:
Obsoleted by https://github.com/NLnetLabs/cascade/pull/136.

PR #129, opened 2025-10-05
Don't sign the same zone concurrently.

ximon18:
Also:
- Don't sign a halted zone.
- Block if the queue is full, rather than blindly overwriting the head of the queue.


mozzieongit (commented):
I just had a quick look and will continue in the alpha release pr later

ximon18:
Obsoleted by https://github.com/NLnetLabs/cascade/pull/136.

PR #153, opened 2025-10-07
Resume the pipeline when a new zone is loaded

bal-e:
If a zone's pipeline is soft-halted, and a new version of the zone is loaded, the pipeline doesn't resume.  Part of the reason for this was to stop the zone signer from signing an unsigned zone that hasn't been approved.  A secondary issue is that a 'SignZone' command (e.g. issued autonomously by the key manager), while the pipeline is halted, causes the zone signer to hard-halt.  This accidentally turns soft-halts into hard-halts.

Now, a new zonetree for ready-to-sign zones has been added.  Unsigned zones are added to this tree when they are approved.  This separates the pipeline mode from the unsigned zone approval state; the zone signer can be run at any time, and it will only touch approved zones.  The zone signer has also been adjusted to leave the pipeline mode alone if it is already halted.

### DIFF
--- a/src/units/zone_signer.rs
+++ b/src/units/zone_signer.rs
@@ -1997,12 +1997,10 @@ enum SignerError {
 
 impl SignerError {
     fn is_benign(&self) -> bool {
-        match self {
-            SignerError::CannotSignUnapprovedZone | SignerError::CannotResignNonPublishedZone => {
-                true
-            }
-            _ => false,
-        }
+        matches!(
+            self,
+            SignerError::CannotSignUnapprovedZone | SignerError::CannotResignNonPublishedZone
+        )
     }
 }
 


src/targets/central_command.rs
```diff
@@ -159,10 +159,17 @@ impl CentralCommand {
...
 
                 if let Some(zone) = get_zone(&self.center, &zone_name) {
-                    if let Ok(zone_state) = zone.state.lock() {
-                        if matches!(zone_state.pipeline_mode, PipelineMode::HardHalt(_)) {
-                            warn!("[CC]: NOT instructing review server to publish the unsigned zone as the pipeline for the zone is hard halted");
-                            return;
+                    if let Ok(mut zone_state) = zone.state.lock() {
+                        match zone_state.pipeline_mode.clone() {
+                            PipelineMode::Running => {}
+                            PipelineMode::SoftHalt(message) => {
```
ximon18, line 165: Is there a reason you chose to include the soft-halt message in your log message, but didn't also add it to the hard-halt log message?

src/units/zone_server.rs
```diff
@@ -754,13 +755,34 @@ impl ZoneReviewApi {
...
-                    "unsigned",
-                    Update::UnsignedZoneApprovedEvent {
-                        zone_name: zone_name.clone(),
-                        zone_serial,
-                    },
-                ),
+                Source::UnsignedZones => {
+                    let unsigned_zones = self.center.unsigned_zones.load();
+                    let Some(zone) = unsigned_zones.get_zone(&zone_name, Class::IN) else {
+                        debug!("Couldn't find the approved zone");
```
ximon18, line 761:
```suggestion
                        debug!("Couldn't find an approved zone called '{zone_name}.");
```

src/units/zone_server.rs
```diff
@@ -754,13 +755,34 @@ impl ZoneReviewApi {
...
+                    };
+
+                    let signable_zones = self.center.signable_zones.load();
+
+                    // Create a deep copy of the set of unsigned zones. We will add
+                    // the new zone to that copied set and then replace the original
+                    // set with the new set.
+                    info!("[{unit_name}]: Adding '{zone_name}' to the set of signable zones.");
+                    let mut new_signable_zones = Arc::unwrap_or_clone(signable_zones.clone());
+                    let _ = new_signable_zones.remove_zone(&zone_name, Class::IN);
```
ximon18, line 772: What is this remove doing?

src/units/zone_signer.rs
```diff
@@ -469,7 +469,9 @@ impl ZoneSigner {
 
             // Do NOT sign a zone that is halted.
             if zone_state.pipeline_mode != PipelineMode::Running {
-                return Err(SignerError::PipelineIsHalted);
+                // TODO: This accidentally sets an existing soft-halt to a hard-halt.
+                // return Err(SignerError::PipelineIsHalted);
+                return Ok(());
```
ximon18, line 474: If you return Ok here the zone will end up in zone status output as having finished signing rather than being aborted. The alternative is to add an error variant like `CannotResignNonPublishedZone` and handle it specially in `handle_command()`.

ximon18 (commented):
Nice work @bal-e!

I'll build and test it, if that looks good I'll approve the PR.

ximon18:
Ahha, maybe it only adds the zone to the new signable zones collection if it is approved, but I have no approval required at present.

ximon18:
@Arya: I pushed a change that also promotes the zone if review is not required. It now works for me.

mozzieongit:
> I also have some issues when doing (with local zone file):
> 
>     * add policy with rejecting review script (and review required)
> 
>     * add zone
> 
>     * change policy to use different script (that still rejects)
> 
>     * run `policy reload`
> 
>     * run `zone status ...` (now shows the new review script, and the existing error messsage (so is out of sync, but that is known, I think)
> 
>     * run `zone reload ...`
> 
>     * change script to accept zone
> 
>     * run `zone reload ...`
> 
>     * `zone status` says loaded new version bla 1s ago, but is still waiting for approval and error that the review rejected the zone.
> 
> 
> (I haven't looked further through the code yet, but want to get this out already)

I might be an idiot ... `command not found` ...

mozzieongit (commented):
After @ximon18 's [comment](https://github.com/NLnetLabs/cascade/pull/153/files#r2410211755) about the `Ok(())`, this would be approved

PR #214, opened 2025-10-14
Serve fake RRSIG RRSET with corrected TTL.

ximon18:
To work around the zone tree storing RRSIGs as an RRSET with a single TTL which is incorrect for RRSIGs.

Note: Switches from `ApexZone` to `LightWeightZone` for signed zones so that the zone walk behaviour (used when serving the zone via XFR) can be overridden, which means we lose the more proper DNS response generation, but gain reduced memory consumption.

This PR is an alternative to https://github.com/NLnetLabs/domain/pull/577 to address the same problem without having to hack on `domain`.

### DIFF
--- a/src/units/zone_signer.rs
+++ b/src/units/zone_signer.rs
@@ -30,7 +30,7 @@ use domain::rdata::{Dnskey, Nsec3param, Rrsig, Soa, ZoneRecordData};
 use domain::zonefile::inplace::{Entry, Zonefile};
 use domain::zonetree::types::{StoredRecordData, ZoneUpdate};
 use domain::zonetree::update::ZoneUpdater;
-use domain::zonetree::{StoredName, StoredRecord, Zone, ZoneBuilder};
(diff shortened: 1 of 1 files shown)


ximon18:
> I don't know if there is any negative effect of switching to the lightweight zone tree, but the change looks good to me.

There is a negative effect, but for Cascade today it is not a regression compared to what we aim to provide, namely that AXFR queries and simple queries like SOA still work as intended, but queries that maybe should result in a nuanced DNS response e.g. should it be NODATA or NXDOMAIN etc may not be answered per the RFCs, as the `LightWeightZone` has really simplistic logic for responding to queries compared to `domain::zonetree::Zone`.

Also, `domain::zonetree::Zone` has built-in support for accumulating a diff for serving via IXFR, but we're not using that yet nor is it clear if we will use it do IXFR or will solve that a different way.

The other effect is reduced memory usage.

I'll update this PR with examples of the differing query responses and impact on memory usage.

PR #217, opened 2025-10-15
Note changes introduced in 0.1.0-alpha2.

ximon18:
WIP

### DIFF
--- a/doc/manual/source/man/cascaded.rst
+++ b/doc/manual/source/man/cascaded.rst
@@ -44,8 +44,8 @@ Options
           file:<PATH>, syslog].
 
 .. versionchanged:: 0.1.0-alpha2
-   Added types `stdout` and `stderr`. Type `file` with values `/dev/stdout`
-   and `/dev/stderr` can still be used but may not work properly in some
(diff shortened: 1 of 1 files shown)


AlexanderBand:
I never tried using these directives in a man page, so I'm not sure what `rst2man` will do with these. You may need to add `.. only:: html` like we do with the badges on the index page.

doc/manual/source/man/cascaded.rst
```diff
@@ -43,6 +43,11 @@ Options
           Where logs should be written to [possible values: stdout, stderr,
           file:<PATH>, syslog].
 
+.. versionchanged:: 0.1.0-alpha2
+   Added types `stdout` and `stderr`. Type `file` with values `/dev/stdout`
+   and `/dev/stderr` can still be used but may not work properly in some
+   cases, e.g. when running under systemd.
```
mozzieongit:
```suggestion
.. versionchanged:: 0.1.0-alpha2
   Added types ``stdout`` and ``stderr``. Type ``file`` with values ``/dev/stdout``
   and ``/dev/stderr`` can still be used but may not work properly in some
   cases, e.g. when running under systemd.
```

PR #218, opened 2025-10-15
Simplify review script example and mention need for policy reload

mozzieongit:


doc/manual/source/review-hooks.rst
```diff
@@ -80,4 +74,8 @@ Next, we update the zone's policy to use the review script for both stages:
     cmd-hook = "/usr/local/bin/cascade-review.sh"
 
 
+Finally, we need to reload the policy with ``cascade policy reload`` (and in
```
ximon18: @mozzieongit: This seems like an unrelated change?

PR #230, opened 2025-10-17
Note steps required to upgrade an alpha version of Cascade.

ximon18:


maertsen:
Will the listed commands remove key material too? In that case, it would be useful to explicitly mention that.

ximon18:
> > Will the listed commands remove key material too? In that case, it would be useful to explicitly mention that.
> 
> It removes everything that is in files. It won't touch anything created in HSMs.

I have updated the text to mention the impact on signing keys.

ximon18:
> > > Will the listed commands remove key material too? In that case, it would be useful to explicitly mention that.
> > 
> > 
> > It removes everything that is in files. It won't touch anything created in HSMs.
> 
> I have updated the text to mention the impact on signing keys.

We _could_ indicate how to delete *only* the affected files as key file format is unlikely to change, but that's harder to describe and a bit brittle as if we affect more files in a later release or add, remove or rename files the instructions will need updating to match.

doc/manual/source/building.rst
```diff
@@ -112,6 +112,33 @@ Then update the ``config.toml`` to use the appropriate paths.
...
+   In its current alpha version form Cascade will likely report errors if a
+   newer version is started using existing state and policy files created by
+   an older version.
+
+   Before upgrading delete all state and policy files. To enable you to keep
+   using existing signing keys after upgrade, the instructions below backup
+   your on-disk key files. Keys stored in HSMs are not affected by these
+   instructions. After upgrade you will need to use the ``--import-xxx``
+   options to the ``cascade zone add`` command to use them with any zones that
+   you (re)create.  will delete signing keys stored on disk**. If HSM signing
```
tertsdiepraam: This is grammatically wrong, but I don't know what exactly you want to write here.

doc/manual/source/building.rst
```diff
@@ -112,6 +112,33 @@ Then update the ``config.toml`` to use the appropriate paths.
...
+
+   *(if you modified any of the filesystem locations specified in your Cascade
+   config file, use the updated paths instead of the default paths shown in
+   these instructions)*
+ 
+   .. code-block:: bash
+ 
+      sudo pkill cascaded
+      sudo mkdir /tmp/keys
+      sudp cp /var/lib/cascade/keys/{*.key,*.private} /tmp/keys/
```
tertsdiepraam:
```suggestion
      sudo cp /var/lib/cascade/keys/{*.key,*.private} /tmp/keys/
```
Same for the other copies of this text

doc/manual/source/building.rst
ximon18: I thought I already corrected that, but after some discussion I'm going to revert the instructions about preserving keys anyway.

Issue #243, opened 2025-10-18
cascade status uses colors in spite of !isatty() and TERM=dumb


jpmens:
There are situations where colours are impossible (for me) to see:

![image](https://github.com/user-attachments/assets/cfed657d-46a0-45ed-891b-d14d787ed33c)

Neither a useless use of cat `| cat` nor setting `TERM=dumb` changes anything.

```

Signing queue:
  Key: ESC[0;33mIn Progress ()ESC[0m, ESC[38;5;248mPending ()ESC[0m, ESC[0;32mFinished   (✔)ESC[0m
   #:   When                      Zone             Action
   0: ESC[0;32m✔ESC[0m 2025-10-18T17:27:38+00:00 example.net      Finished
```

jpmens:
The colour difficulty applies in many places, understandably.

![image](https://github.com/user-attachments/assets/a7ceef40-b390-4ce8-b1db-55329ac80ef9)

I feel colours ought to either be customizable or at least should be able to be disabled.

PR #254, opened 2025-10-20
Allow disabling colors

tertsdiepraam:
Closes https://github.com/NLnetLabs/cascade/issues/243.

This might be a completely overengineered solution, but here's my thinking:
- We want to allow separate settings for stdout and stderr, so we can't just set a global "use color" variable.
- We want to ensure that we catch all cases where we print color and don't accidentally print color when we're not supposed to. That rules out [anstream](https://docs.rs/anstream), which wraps the `println!` macro to optionally strip all color. We can't ensure that we only use that macro instead of `std::println`.

So I made a new trait `Print` which is `Display` with a parameter for whether color should be printed. All `Display` types implement `Print` and a type `Styled` implements _only_ `Print` (and not `Display`). This `Styled` type can only be printed if it knows whether to print colors or not, which is information that we pass from our own `println!` and `eprintln!` macros.

The downside is that the wrapped `println!`/`eprintln!` macros are not quite as powerful as `std`'s macros. They don't support `println!("{x}", x = y)` and all the styled arguments cannot appear inside the string, so `println!("{x}")` doesn't work if `x` is a `Styled`, only `println!("{}", x)`.

To figure out whether the terminal supports colors we use [supports_color](https://docs.rs/supports_color), which gets a lot of things right out of the box.

image

If this is too overengineered we could just replace all our uses of `println!` with the anstream equivalent instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -274,6 +274,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_with",
+ "supports-color",
  "syslog",
  "tempfile",
  "tokio",
@@ -1131,6 +1132,12 @@ dependencies = [
  "memchr",
  "serde",
 ]
+
+[[package]]
+name = "is_ci"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7655c9839580ee829dfacba1d1278c2b7883e50a277ff7541299489d6bdfdc45"
 
 [[package]]
 name = "is_terminal_polyfill"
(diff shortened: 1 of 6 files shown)


bal-e:
Right, that's a cool way of solving thus.  While the API ends up being alright, I do think this solution is a bit clunky.  I know that it's possible to forget to use `anstream::println!`, but we just need to put `use anstream::{println, eprintln};` at the top of every CLI module.  Maybe we'll miss some cases sometimes, but IMO this would be a simpler and more maintainable solution.

Issue #249, opened 2025-10-19
It would be nice to have mnemonics for durations, like in OpenDNSSEC


bortzmeyer:
```
% grep zsk.val /etc/cascade/policies/default.toml
zsk.validity = "1W" 

% cascade policy reload                          
[2025-10-19T13:40:39.788Z] ERROR cascade: Error: /etc/cascade/policies/default.toml: "1W" is not 'forever' or an integer
```

This is not very user-friendly.

jpmens:
Agree. It would seem it's planned:

```console
$ cascade template policy | grep '=.* TODO'
ksk.validity = "31536000" # TODO: "1Y"
zsk.validity = "2592000" # TODO: "1M"
ttl = 3600 # TODO: "1h"
dnskey.signature-inception-offset = 86400 # TODO: "1d"
dnskey.signature-lifetime = 1209600 # TODO: "2w"
dnskey.signature-remain-time = 604800 # TODO: "1w"
signature-inception-offset = 86400 # TODO: "1d"
signature-lifetime = 1209600 # TODO: "2w"
signature-remain-time = 604800 # TODO: "1w"
```

mozzieongit:
Indeed it is planned.

To have your input. What time would you expect 1M and 1Y to be (there were a few discussions internally before...)?
1M = 30 days? 30.436875 days? 28 days? 31 days?
1Y = 365 days? 365.2425 days? 364 days (aka 52 weeks)? 366 days?

Or any one of the above as long as it is documented (it would be documented either way)?

bortzmeyer:
Ticket still open but:

```
ksk.validity = "60d"
zsk.validity = "15d"
csk.validity = "120d"
```

works fine on "0.1.0-beta5".

I regret the choice to use a different syntax from OpenDNSSEC. Expressing months and years is not convenient.

PR #262, opened 2025-10-22
Improve timespan parsing

tertsdiepraam:
Improve the timespan parsing in two ways:
- Allow bare integers as seconds
- Allow all the parsing goodness of `jiff`: https://docs.rs/jiff/latest/jiff/fmt/friendly/index.html. Except, `jiff` doesn't allow us to do anything larger than weeks, because that would be ambiguous. That means that `1Y` is not supported but the less ambiguous `365d` is fine.

I couldn't figure out a nicer implementation that handwriting the visitors for this, but that's fine, because it's a fairly complex bit of parsing.

Here are some examples:
```
10           # 10 seconds
"10"         # 10 seconds
"10s"        # 10 seconds
"10m"        # 10 minutes
"10h"        # 10 hours
"10d"        # 10 days
"10w"        # 10 weeks
"2h 3m 4s"   # 2 hours, 3 minutes and 4 seconds
"P35DT2H30M" # 35 day, 2 hours and 30 minutes
```

This (partially) resolves the issue in https://github.com/NLnetLabs/cascade/issues/249, although we intentionally don't accept units larger than weeks.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -431,6 +431,7 @@ pub struct ReviewPolicyInfo {
 #[derive(Deserialize, Serialize, Debug, Clone)]
 pub struct SignerPolicyInfo {
     pub serial_policy: SignerSerialPolicyInfo,
+    // TODO: These fields should have a type that explains that they represent durations.
     pub sig_inception_offset: u32,
     pub sig_validity_offset: u32,
     pub denial: SignerDenialPolicyInfo,
--- a/src/common/datetime.rs
+++ b/src/common/datetime.rs
@@ -7,7 +7,7 @@ use serde::{
     Deserialize, Deserializer, Serialize,
 };
 
-/// A wrapper around [`Duration`] with fancier (de)serialization
+/// A wrapper around [`u32`] representing a [`Duration`]
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub struct TimeSpan(u32);
 


mozzieongit:
Don't forget to update the policy.template.toml file and document the ability to use the strings

doc/manual/source/man/cascaded-policy.toml.rst
```diff
@@ -180,7 +180,9 @@ The ``[key-manager]`` section.
    security of the zone.
 
    Independent validity times are set for KSKs, ZSKs, and CSKs.  An integer
-   value will be interpreted as seconds; ``forever`` means keys never expire.
+   value will be interpreted as seconds, 'forever' means keys never expire, and
+   a time string such as "365d" will be interpreted as 365 days. Supported
+   suffixes include "s", "m", "h", "d" and "w".
```
mozzieongit:
```suggestion
   value will be interpreted as seconds, ``forever`` means keys never expire, and
   a time string such as ``"365d"`` will be interpreted as 365 days. Supported
   suffixes include ``s``, ``m``, ``h``, ``d`` and ``w``.
```

doc/manual/source/man/cascaded-policy.toml.rst
```diff
@@ -417,7 +424,8 @@ zone) are signed by the key manager, rather than the zone signer; see the
    signatures will be regenerated; it must be less than the ``signature-lifetime``
    setting.
 
-   An integer value is interpreted as seconds.
+   An integer value is interpreted as seconds. A string is interpreted as time
+   string with a number followed by a unit (i.e. "s", "m", "h", "d", or "w").
```
mozzieongit:
```suggestion
   An integer value is interpreted as seconds. A string is interpreted as a time
   string and is a number followed by a unit (i.e. ``s``, ``m``, ``h``, ``d``, or ``w``).
```

mozzieongit (commented):
(Still need to look at `datetime.rs` and the `policy/v1/mod.rs` file, but submitting this already)

tertsdiepraam:
Something else to note: the time parsing is not fully documented at the moment just the simple versions. Ultimately we should organize this in a way that we don't need all that repetition.

src/common/datetime.rs
```diff
@@ -0,0 +1,166 @@
+use std::{fmt, str::FromStr, time::Duration};
+
+use domain::base::Ttl;
+use jiff::{Span, SpanRelativeTo};
+use serde::{
+    de::{self, Visitor},
+    Deserialize, Deserializer, Serialize,
+};
+
+/// A wrapper around [`Duration`] with fancier (de)serialization
```
mozzieongit: Is this meant to say `Duration` or `u32`, because the struct below wraps `u32`

mozzieongit (approved):
One note, but not blocking

src/api.rs
```diff
@@ -431,8 +431,8 @@ pub struct ReviewPolicyInfo {
 #[derive(Deserialize, Serialize, Debug, Clone)]
 pub struct SignerPolicyInfo {
     pub serial_policy: SignerSerialPolicyInfo,
-    pub sig_inception_offset: Duration,
-    pub sig_validity_offset: Duration,
+    pub sig_inception_offset: u32,
+    pub sig_validity_offset: u32,
```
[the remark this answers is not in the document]
bal-e: I think they should all have a type that indicates they are temporal values.

PR #280, opened 2025-10-28
Use soft halt instead of hard halt for zone parsing errors

### LINKED ISSUE #269: Not all zone issues should hard halt the zone (e.g. unknown RRtype hard halts zone)
When getting a zone that fails parsing the zone is hard halted and cannot be resumed after the broken record has been fixed.

tertsdiepraam:
Closes #269.

This includes 2 main changes:
1. Don't terminate the zone loader (crashing cascade) on a zone reload error.
2. Only soft halt the zone after an error so it's easy to recover with a reload.

### DIFF
--- a/src/units/zone_loader.rs
+++ b/src/units/zone_loader.rs
@@ -310,7 +310,7 @@ impl ZoneLoader {
                         &zone_updated_tx,
                         receipt_info,
                     )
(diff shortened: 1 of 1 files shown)


bal-e (approved):
Sweet!

Issue #278, opened 2025-10-28
Review script inherits cascade's stdout/stderr instead of the output getting collected and printed to the configured log-target


mozzieongit:
> Worse: with alpha-3, I get the output of the review script in the daemon *standard output*, not in the log. It means that, unless something (systemd?) captures this output, it may be unnoticed. Tested with and without ` --daemonize` )
> 
> This is all I get in the log:
> 
> ```
> Oct 27 14:30:36 js user.debug js cascaded[17969]: [RS] Received command: SeekApprovalForUnsignedZone { zone_name: Name(example.org.), zone_serial: Serial(2025102705) }
> Oct 27 14:30:36 js user.info js cascaded[17969]: [RS]: Seeking approval for unsigned zone 'example.org' at serial 2025102705.
> Oct 27 14:30:36 js user.info js cascaded[17969]: [RS]: Executed hook '/etc/cascade/reviews/review.sh unsigned' for unsigned zone 'example.org' at serial 2025102705
> Oct 27 14:30:36 js daemon.notice cascade: Validating example.org of serial 2025102705 from 127.0.0.1:4541
> Oct 27 14:30:36 js user.info js cascaded[17969]: preprocess; request_ip=127.0.0.1
> Oct 27 14:30:36 js user.info js cascaded[17969]: AXFR for example.org from 127.0.0.1:42971
> Oct 27 14:30:36 js user.debug js cascaded[17969]: [RS]: Hook '/etc/cascade/reviews/review.sh unsigned' exited with status exit status: 1
> Oct 27 14:30:36 js user.debug js cascaded[17969]: [CC]: Event received: ReviewZone { name: Name(example.org.), stage: Unsigned, serial: Serial(2025102705), decision: Reject }
> ```
> 
> In the output, there was:
> 
> ```
> /tmp/.cascade_zone.rEe2jKjlJD:15: there should be at least two NS records per name
> /tmp/.cascade_zone.1mZ3qab6qJ:14: there should be at least two NS records per name
> ```
> 
> 
>  

 _Originally posted by @bortzmeyer in [#215](https://github.com/NLnetLabs/cascade/issues/215#issuecomment-3451597163)_

PR #281, opened 2025-10-28
Log review script stdout and stderr to the log target

tertsdiepraam:
Closes https://github.com/NLnetLabs/cascade/issues/278


bal-e (approved):
Looks good!  I think `.next_line()` is allocating a new `String` every time you call it, but it's not a real issue.  LGTM.

PR #287, opened 2025-10-29
Use tracing subscriber to log the log messages

tertsdiepraam:
🌈 Pretty colors 🌈 

A next step would be to ~completely remove the `log` crate and~ start using `tracing` spans to give the logging more info, so we can start doing more filtering (e.g. filter on the zone).~

I apologize for the branch name.

### DIFF
--- a/src/log.rs
+++ b/src/log.rs
@@ -6,7 +6,7 @@ use std::os::unix::net::UnixDatagram;
 use std::path::Path;
 
 use tracing::field::{self, Field};
-use tracing::{error, warn, Level, Subscriber};
+use tracing::{error, Level, Subscriber};
 use tracing_subscriber::filter::LevelFilter;
 use tracing_subscriber::fmt::Layer as FmtLayer;
 use tracing_subscriber::layer::SubscriberExt;


tertsdiepraam:
I'd like to not wait to long with this as I don't want to keep syncing it.

src/log.rs
```diff
@@ -25,178 +32,90 @@ impl Logger {
...
-}
-
-impl log::Log for Logger {
-    fn enabled(&self, metadata: &log::Metadata) -> bool {
-        if let Ok(inner) = self.inner.read() {
-            if let Some(inner) = &*inner {
-                return inner.enabled(metadata);
-            }
+    pub fn launch(config: &LoggingConfig) -> Result<&'static Logger, String> {
+        let mut filter = EnvFilter::from_env("CASCADE_LOG");
```
mozzieongit: Is the use of `CASCADE_LOG` already documented somewhere?

mozzieongit (commented):
~I wonder why we even keep the `log::` and now the `tracing::` around instead of just importing the macro itself~

ximon18:
So do I understand correctly that "pretty colours" are only when logging to stdxxx? I don't see the colour codes in the output e.g. when logging to a file and also need to disable daemonizing to see the whole log output on stdxxx.

tertsdiepraam:
> So do I understand correctly that "pretty colours" are only when logging to stdxxx? I don't see the colour codes in the output e.g. when logging to a file and also need to disable daemonizing to see the whole log output on stdxxx.

Yes, this was on purpose. Happy to change it. Even for files you can force it with `FORCE_COLOR=1`. Maybe syslog should include colors? Unsure how to determine that.

src/log.rs
```diff
@@ -25,178 +32,85 @@ impl Logger {
...
-                return inner.flush();
-            }
-        }
-
-        let mut logger = &self.fallback;
-        let _ = logger.flush();
+    pub fn apply(&self, config: &LoggingConfig) -> Result<(), String> {
+        self.filter
+            .reload(make_env_filter(config)?)
+            .map_err(|_| "could not reload filter".into())
```
bal-e: Can you elaborate on this error message?

src/log.rs
```diff
@@ -25,178 +32,85 @@ impl Logger {
...
-struct Inner {
-    /// The primary logger.
-    primary: PrimaryLogger,
-
-    /// A log level filter.
-    level: log::LevelFilter,
-
-    /// A list of log targets to trace.
-    trace_targets: foldhash::HashSet<Box<str>>,
+fn make_env_filter(config: &LoggingConfig) -> Result<EnvFilter, String> {
```
bal-e: Could you document this function?

mozzieongit:
> Yes, this was on purpose. Happy to change it. Even for files you can force it with `FORCE_COLOR=1`. Maybe syslog should include colors? Unsure how to determine that.

Files should not contain ansi color codes. If a user wants them, they should use FORCE_COLOR=1
Syslog should not include ansi color codes, the log level determines the color in output like journalctl for example.

src/log.rs
```diff
@@ -25,178 +32,85 @@ impl Logger {
...
+    // Create an envfilter with the empty string, which won't read any env vars
+    // and only print ERROR by default. Which we then immediately override with
+    // another LevelFilter.
+    let mut filter = EnvFilter::new("");
+    filter = filter.add_directive(LevelFilter::from(*config.level.value()).into());
+    for target in config.trace_targets.value().iter() {
+        filter = filter.add_directive(
+            target
+                .parse()
+                .map_err(|_| format!("invalid trace target: \'{}\'", target))?,
```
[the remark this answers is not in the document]
tertsdiepraam: We can `warn!` about them instead?

Issue #303, opened 2025-11-03
Rework inter-component communication


bal-e:
At the moment, the components within Cascade (zone loader, zone signer, and zone server) communicate with each other in very ad-hoc ways.  This makes it difficult to enumerate all the interactions between components and makes it harder to persist all of Cascade's state.  The components need to be restructured so they can all communicate through the central command in a consistent way, and store their state in a single place.

- [x] #329

PR #329, opened 2025-11-05
Introduce `Manager` and begin simplifying unit startup

bal-e:
This PR begins work on #303, simplifying how units are started up and making it possible to interact with them through direct methods.  It begins cutting away at the `update_tx` and `app_cmd_tx` channels and moves some state into `ZoneState`.  `CentralCommand` moves into `manager.rs`, becoming `Manager`, with direct access to all the units.  The goal is to move to inter-component communication via direct synchronous methods, eliminating failure cases (`.send().unwrap()`) and passing around richer information.  This helps us isolate inter-component communication into the central command, so we have a clearer understanding of Cascade's state machine.

### DIFF
--- a/src/units/zone_server.rs
+++ b/src/units/zone_server.rs
@@ -4,7 +4,6 @@ use std::net::IpAddr;
 use std::pin::Pin;
 use std::process::Stdio;
 use std::sync::Arc;
-use std::time::Instant;
 
 use arc_swap::ArcSwap;
 use bytes::Bytes;
@@ -31,7 +30,7 @@ use domain::tsig::{Algorithm, Key};
 use domain::zonetree::types::EmptyZoneDiff;
 use domain::zonetree::Answer;
 use domain::zonetree::{StoredName, ZoneTree};
-use tokio::sync::{mpsc, RwLock};
+use tokio::sync::mpsc;
 use tracing::{debug, error, info, warn};
 
 use crate::api::{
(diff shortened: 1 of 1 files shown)


src/units/zone_server.rs
```diff
@@ -630,14 +568,76 @@ impl ZoneServer {
...
+        };
+
+        // Look up the version of the zone being reviewed.
+        match self.source {
+            Source::Unsigned => {
+                {
+                    let mut zone_state = zone.state.lock().unwrap();
+                    let Some(version) = zone_state.unsigned.get_mut(&zone_serial) else {
+                        debug!("[{unit_name}] Got a review for {zone_name}/{zone_serial}, but it was not pending review");
+                        let _ = tx.send(Err(ZoneReviewError::NotUnderReview));
```
tertsdiepraam: This error implies that you also check the review state? The previous version seemed to do that too.

src/manager.rs
```diff
@@ -1,226 +1,630 @@
...
+                "Passing back zone review",
+                match stage {
+                    api::ZoneReviewStage::Unsigned => "RS",
+                    api::ZoneReviewStage::Signed => "RS2",
+                },
+                ApplicationCommand::ReviewZone {
+                    name,
+                    serial,
+                    decision,
+                    tx: tokio::sync::oneshot::channel().0,
```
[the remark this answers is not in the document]
bal-e: Yeah, we'll slowly get replace the channels entirely, and these one-shot channels will go with them.

src/units/zone_server.rs
bal-e: OH!  Good catch!

(and yeah, sorry about that)

PR #355, opened 2025-11-18
Propagate policy changes to keyset.

Philip-NLnetLabs:


bal-e (commented):
The code looks great!  Just double-checking: were you able to test that this propagates policy changes properly?  E.g. to update key expiry times?

Issue #317, opened 2025-11-03
Set names for Tokio's worker threads


bal-e:
Cascade uses Tokio internally, and it spawns threads.  These threads currently use the default name `tokio-worker` -- we probably want to change it to something Cascade-specific.

PR #356, opened 2025-11-20
set names for cascade's tokio threads

tertsdiepraam:
I would've loved to set names for the _tasks_ but that is still unstable in tokio. Instead, we just change the names of the treads. This makes it less important but maybe still a nice improvement.

Closes https://github.com/NLnetLabs/cascade/issues/317


ximon18 (approved):

tertsdiepraam:
image

Seems to work based on a quick look at htop

ximon18 (approved):

PR #365, opened 2025-11-22
first attempt at describing how to wire up a Nitrokey NetHSM

jpmens:
I hope I haven't made too many rST blunders ... ;)

### DIFF
--- a/doc/manual/source/nethsm.rst
+++ b/doc/manual/source/nethsm.rst
@@ -1,0 +1,466 @@
+Integrating with a Nitrokey NetHSM
+==================================
+
+.. Note:: The instructions on this page are for a Debian 12 host and assume
+   that Cascade has already been installed using our DEB package, and that
+   Docker is installed and can run the NetHSM container.
+
+.. Note:: The instructions on this page assume you will be using the NetHSM
+   exclusively for the task at hand and for testing. Using the provided
+   test image is in no way suggested for production. Please study the
+   product documentation to find out how to do so.
+
+.. Note:: Most instructions on this page assume you will be working as a
+   normal, i.e. non-``root``, user.
+
+.. epigraph::
+
+   NetHSM is an open hardware security module created and distributed by
+   Nitrokey. It's a secure store for cryptographic keys powered by
+   open source, which enables people to verify exactly how it works.
+   The HSM is accessible through a REST interface, and a PKCS #11 shared object
+   interfaces that with Cascade's :program:`kmip2pkcs11`. The software is also
+   provided as an OCI image (used here for demonstration purposes) which can be
+   used with Docker or Podman for experimentation without having to purchase
+   the device proper.
+
+   -- https://www.nitrokey.com/products/nethsm
+
+
+Launch the NetHSM container
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We launch the container to be removed on exit which means all settings and keys
+will be wiped when the container is stopped.
+
+.. code-block:: bash
+
+   $ docker run --rm -ti -p 127.0.0.1:8443:8443 docker.io/nitrokey/nethsm:testing
+
+Install the Prerequisites
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: bash
+
+   $ sudo apt install -y opensc opensc-pkcs11 pipx
+   $ pipx install pynitrokey    # for the nitropy utility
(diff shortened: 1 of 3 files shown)


ximon18:
Thank you very much @jpmens!

The lack of a readable label is I think a limitation of the NitroKey NetHSM. For example something similar is described in their documentation [here](https://docs.nitrokey.com/nethsm/pkcs11-setup#):

```
# Set this option to true to enable the compatibility option for the C_SetAttributeValue() function.
# This allows the applications using the Java Sun PKCS11 module (like EJBCA) to generate keys.
# When using this, the names given to the keys will be ignored and the keys will have random names.
# Under the hood it will store in memory the name given to the key when calling C_SetAttributeValue(). When a certificate is uploaded it will check if the name was previously passed to C_SetAttributeValue() and translate it to the real name on the NetHSM.
enable_set_attribute_value: false
```

Did you try looking at the output of `cascade zone status --detailed` to see if that helps identify which keys belong to which zone?

jpmens:
@ximon18 thanks for the input regarding `--detailed`, which does indeed produce the label; I've added that to the text.

Changing the value of `enable_set_attribute_value` in the nethsm configuration and restarting kmip2pkcs11 doesn't have an effect here, an option which is [being removed in the 2.0.0 driver version](https://github.com/Nitrokey/nethsm-pkcs11/blob/f67e9a3e2fbf11e89b4dfdc839982fca4f267093/CHANGELOG.md?plain=1#L14) ;-) I cannot judge [what, if anything](https://github.com/Nitrokey/nethsm-pkcs11/commit/eed2b173ec697bf7ed296c9e1d9bb776ec92cf03) this used to do.

doc/manual/source/nethsm.rst
```diff
@@ -0,0 +1,464 @@
...
+
+Launch the NetHSM container
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We launch the container to be removed on exit which means all settings and keys
+will be wiped when the container is stopped.
+
+.. code-block:: bash
+
+   $ docker run --rm -ti -p 8443:8443 docker.io/nitrokey/nethsm:testing
```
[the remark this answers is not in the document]
jpmens: done

doc/manual/source/nethsm.rst
```diff
@@ -0,0 +1,464 @@
...
+   Command line tool to interact with Nitrokey devices 0.11.2
+   Unlock passphrase: nlnetlabs001
+   Repeat for confirmation:
+   Admin passphrase: lecascadeur
+   Repeat for confirmation:
+   Warning: The unlock passphrase cannot be reset without knowing the current value...
+   NetHSM 127.0.0.1:8443 provisioned
+
+
+Configure NetHSM's TLS certificate
```
and0x000: imo this entire section conflicts with the use of `danger_insecure_cert: true` PKCS#11 driver and the assumption that the HSM container is run locally. personally I'd ditch this and stick to `--no-verify-tls` with the nitropy tool. Maybe with a short reference link to the Nitrokey documentation for how this is done right in a prod setup.

jpmens: I hadn't yet got around to actually testing `danger_insecure_cert: false`, but you are right in that.

I documented the effort because I couldn't actually find documentation for it and under the _assumption_ that people wouldn't typically be using the OCI image as an HSM ...

jpmens: `danger_insecure_cert: false` works as advertised.

and0x000: to be honest, I've also started documenting the nitrokey integration after my findings in https://github.com/NLnetLabs/kmip2pkcs11/issues/14 but It didn't meet my requirements yet for having this somewhere public, the commands haven't been tested front to back to be all copy/pastable etc ... However, as you are pretty much writing the same thing, feel free to compare and rip anything from my mumblings https://gist.github.com/and0x000/a50bdb2f883247bf00426c1c1140c091

It's a pretty shameless rewrite of the thales documentation anyway 🙈

AlexanderBand:
> I hope I haven't made too many rST blunders ... ;)

I read through the render of the page [here](https://cascade-signer--365.org.readthedocs.build/en/365/nethsm.html) and did not spot any mistakes in the ReStructuredText syntax. Nicely done! 

The only thing missing is a reference to this page in `index.rst`so it shows up in the Table of Contents.

PR #382, opened 2025-11-28
Split `KeyRollCommand` between CLI and API

bal-e:
This is a subtle but valuable change: the CLI and API will evolve differently over time and sharing definitions across them will cause problems in the long run.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -671,7 +671,7 @@ pub mod keyset {
         Algorithm,
     }
 
-    #[derive(Deserialize, Serialize, Clone, Debug, clap::Subcommand)]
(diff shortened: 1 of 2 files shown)


mozzieongit (commented):
Sure. My only concern is the annoyance of having multiple definitions "everywhere".
While I can see how that would become an issue when e.g. the CLI only supports some KeyRollCommand's while the API (e.g. via Web) allows more, I don't think that would happen anytime soon in this case.

I don't know if this particular example of code sharing is problematic.

Or is this in the context of (potentially/actually) splitting cascade up into different crates?

But in the end, I don't object strongly enough to block this.

PR #390, opened 2025-11-28
Split the CLI and API into separate crates

bal-e:
Summary of the changes:

- The daemon remains in place but is renamed to `cascaded`.
- The CLI has been moved into `crates/cli`, and assigned the crate name `cascade`.
- The API has been moved into `crates/api`, and assigned the crate name `cascade-api`.
- Links between the CLI and the daemon have been removed.
- The Cargo manifest is converted into a workspace.
- Common dependencies are moved to `[workspace.dependencies]` and documented.
- The ReadTheDocs builder is updated to look at `workspace.package.version`.

This change is justified for a few reasons:

- This enforces the natural logical separation between Cascade's
  high-level components.  Until now, it was easy to accidentally
  introduce dependencies between the CLI and the daemon.

- The source code for the CLI was previously split between different
  locations.  With this change, it is now unified.

- We can now categorize our dependencies by high-level component, e.g.
  `reqwest` is only used by the CLI.  This helps us maintain our
  dependency graph.

- We expect additional crates to be added in the future (e.g. for zone
  storage).  Performing the split now will ease that effort.

- Splitting into crates introduces new points of parallelism and so
  should benefit compilation time.


Cargo.toml
```diff
@@ -1,45 +1,103 @@
...
+# 'domain' is a general-purpose library for DNS.
+#
+# It is maintained by us (NLnet Labs).
+[workspace.dependencies.domain]
+git = "https://github.com/NLnetLabs/domain"
+branch = "patches-for-nameshed-prototype"
+
+# 'jiff' is a general-purpose time library.
+#
+# TODO: Maintenance?
```
tertsdiepraam: Burntsushi so all good :)

crates/cli/Cargo.toml
```diff
@@ -0,0 +1,80 @@
...
+# --- Dependencies -------------------------------------------------------------
+
+[dependencies]
+
+# The CLI uses 'anstream' to conditionally disable ANSI colorization in its
+# output (e.g. in response to 'TERM=dumb').
+#
+# TODO: Maintenance?
+[dependencies.anstream]
+version = "0.6.21"
```
tertsdiepraam: We could unify everything into the workspace Cargo.toml?

tertsdiepraam (approved):
Not the biggest fan of how you've written the deps out, but it's fine and I understand why. The rest is fantastic.

src/lib.rs
```diff
@@ -19,21 +19,3 @@ pub mod zonemaintenance;
...
-}
-
-#[macro_export]
-macro_rules! eprintln {
-    ($($t:tt)*) => {{
-        #[allow(clippy::disallowed_macros)]
-        let x = anstream::eprintln!($($t)*);
-        x
-    }};
-}
```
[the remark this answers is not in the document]
tertsdiepraam: Well the daemon definitely shouldn't println should it? In daemon mode that would just be lost I think?

PR #392, opened 2025-11-28
Bump to MSRV 1.85 and overhaul CI

bal-e:
We're using MSRV 1.85 in 'domain' already, and it unlocks edition 2024 support.  I noticed the CI was hard-coding the MSRV, so I copied our existing CI workflow from 'domain' which is much more thorough.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ panic = "abort"
 #
 # TODO: Maintenance?
 [workspace.dependencies.camino]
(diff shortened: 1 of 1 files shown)


bal-e:
Important note: I just noticed that the existing CI runs tests for many unnecessary combinations, e.g. `--no-default-features` and default features and `--all-features`.  The CI workflow I copied from `domain` should be more efficient.

PR #401, opened 2025-12-03
Improvements to the integration tests framework

mozzieongit:
- Introduce a custom Docker image based on the slimmed down `act` image (and an accompanying wrapper script) to reduce image size, build time, and test run time.
  - The act-wrapper script first generates an updates container image with a freshly built cascade (making use of caching; if no source file changed, a cached container is re-used for the next test run making test iteration more efficient)
- Add installed binaries (`cascade{,d}`/`dnst`) to PATH for easier use
- Add a script to get default paths for cascade configurations
- Add a script to add scaffolding for new tests
- Reduce systemd-tests.yml workflow file verbosity
- Reduce verbosity for successful tests
  - Now only failed tests produce the whole pallet of text (used config file, logs, resolv.conf content, running processes)
- Document the default paths for stuff (Cascade, config, state, ...)
- Document the default paths for stuff (Cascade, config, state, ...)
- Add a build profile selection action
- Remove need/option/docs for standalone artifact server

Consider:
- [ ] Replace bind9 with NSD serving `.test` and provide a script to update the DS RR for `example.test`? (I had used bind9 for it's DNS UPDATE capability)
- [x] Check that error output from composite actions is visible when running the main workflow
- ~Provide more errors in the test env management script when components fail (so lots of `|| { echo "bla"; exit 11; }`; to make spotting failing components easier)~ (obsolete with logging every command run using `set -x`)

### DIFF
--- a/act-wrapper
+++ b/act-wrapper
@@ -42,11 +42,11 @@ if [[ "$1" != "+nobuild" ]]; then
 	cargo build
 	cargo build --release
 
-	docker buildx build . -f integration-tests/cascade-test-image/Dockerfile -t nlnetlabs/cascade-runner --build-arg MSRV=$MSRV "${args[@]}"
+	docker buildx build . -f integration-tests/cascade-test-image/Dockerfile -t nlnetlabs/cascade-tests-runner --build-arg MSRV=$MSRV "${args[@]}"
 	echo "Built Cascade integration test runner based on local repository."
 fi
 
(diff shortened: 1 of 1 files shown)


bal-e:
> It will work even less under GitHub Actions. The container use different users. If that is something we think we should have, I'd make a separate PR. But I would be fine with local only running of integration tests.

I see. My only concern is that it's easy to forget to run the integration tests; but regardless, your work here is making them easier to use, and that's great. We can continue examining this issue in the future.

PR #402, opened 2025-12-08
New zone loader v2

tertsdiepraam:
Supersedes https://github.com/NLnetLabs/cascade/pull/15.

I took that PR and started reviving it. That is, bringing it up to date with main and making it compile.

Some tasks remain:

- [x] Zone removal
- [x] ~Sending notifies~
- [x] Reporting statistics
- [x] Tests! Well, there's 1 additional test in this PR.

### DIFF
--- a/src/center.rs
+++ b/src/center.rs
@@ -9,6 +9,7 @@ use std::{
 
 use arc_swap::ArcSwap;
 use bytes::Bytes;
+use domain::base::iana::Class;
 use domain::rdata::dnssec::Timestamp;
 use domain::zonetree::StoredName;
 use domain::{base::Name, zonetree::ZoneTree};
@@ -194,6 +195,29 @@ pub fn remove_zone(center: &Arc<Center>, name: Name<Bytes>) -> Result<(), ZoneRe
 
     let mut state = center.state.lock().unwrap();
     let zone = state.zones.take(&name).ok_or(ZoneRemoveError::NotFound)?;
+
+    // Remove the zone from all the places it might be stored.
+    // The zone might not have made it to these places, but that's not an issue
+    // so we just ignore any errors.
+
+    center.unsigned_zones.rcu(|z| {
+        let mut z = Arc::unwrap_or_clone(z.clone());
+        let _ = z.remove_zone(&name, Class::IN);
+        z
+    });
+
+    center.signed_zones.rcu(|z| {
+        let mut z = Arc::unwrap_or_clone(z.clone());
+        let _ = z.remove_zone(&name, Class::IN);
+        z
+    });
+
+    center.published_zones.rcu(|z| {
+        let mut z = Arc::unwrap_or_clone(z.clone());
+        let _ = z.remove_zone(&name, Class::IN);
+        z
+    });
+
     let mut zone_state = zone.0.state.lock().unwrap();
 
     // Update the policy's referenced zones.


src/loader/mod.rs
```diff
@@ -0,0 +1,656 @@
...
+        let initial_zones = {
+            let state = this.center.state.lock().unwrap();
+            state.zones.clone()
+        };
+
+        for zone in initial_zones {
+            let zone_state = zone.0.state.lock().unwrap();
+            let source = zone_state.source.clone();
+            drop(zone_state);
+            this.set_source(&zone.0, source);
```
bal-e: `set_source()` also locks the zone state -- maybe pass the mutex guard on via a function parameter?

src/loader/mod.rs
```diff
@@ -0,0 +1,656 @@
...
+                    // a specific case for that.
+                    Change::ZoneSourceChanged(name, zone_load_source) => {
+                        let zone = {
+                            let center = self.center.state.lock().unwrap();
+                            let zone = center.zones.get(&name).unwrap();
+                            zone.0.clone()
+                        };
+                        self.set_source(&zone, zone_load_source);
+                        Ok(())
+                    }
```
bal-e: Optional: would it be possible to figure out where this event gets spawned from and make it call `set_source` there?

src/zone/loader.rs
```diff
@@ -0,0 +1,616 @@
...
+    Zone, ZoneState,
+};
+
+//----------- LoaderState ------------------------------------------------------
+
+/// State for loading new versions of a zone.
+#[derive(Debug, Default)]
+pub struct LoaderState {
+    /// The source of the zone.
+    pub source: Source,
```
bal-e: Hang on... `ZoneState` already has a `source` field.  Can you delete that field, and redirect all its uses to this?  That should simplify stuff.

src/zone/loader.rs
```diff
@@ -0,0 +1,616 @@
...
+    }
+}
+
+//----------- Source -----------------------------------------------------------
+
+/// The source of a zone.
+//
+// TODO: Support multiple sources for a zone?
+#[derive(Clone, Debug, Default)]
+#[allow(clippy::large_enum_variant)]
```
bal-e: Is this because of `tsig::Key`?  I thought it was wrapped in an `Arc` internally.  Do we need to do that?

src/zonemaintenance/types.rs
bal-e: Would it be possible to remove this module entirely?  The serde functions can be moved to `crate::util`.

src/zone/loader.rs
```diff
@@ -0,0 +1,616 @@
...
+                debug!("Loaded serial {serial:?} for {:?}", zone.name);
+
+                // Update the old-base contents.
+                let latest = state.contents.as_ref().unwrap().latest.clone();
+
+                let zone_copy = zone.clone();
+                let loader_copy = loader.clone();
+
+                // Go update the zone in another thread
+                tokio::spawn(async move {
```
bal-e: Should we omit or `.await` the `tokio::spawn()`?  It could lead to a weird situation where two refreshes have finished, and both are updating the zonetree simultaneously.

src/units/http_server.rs
```diff
@@ -537,30 +478,35 @@ impl HttpServer {
...
-                    started_at,
-                    finished_at: Some(finished_at),
-                    byte_count: zone_loaded_bytes,
-                    record_count: zone_loaded_record_count,
-                })
-            }
-            (finished_at, None, Some(started_at)) => Some(ZoneLoaderReport {
+        let metrics = {
+            let s = state.center.state.lock().unwrap();
+            let zone = s.zones.get(&name);
```
bal-e: Perhaps use `center::get_zone()` or drop the global state lock manually?

src/loader/mod.rs
bal-e: This should disappear once the entity spawning `ZoneSourceChanged` uses `LoaderState::set_source()`.

src/loader/zonefile.rs
```diff
@@ -0,0 +1,197 @@
...
+        match record {
+            Parsed::Soa(soa_record) => {
+                if soa.is_some() {
+                    return Err(Error::MultipleSoaRecords);
+                }
+                soa = Some(soa_record);
+            }
+            Parsed::Record(regular_record) => all.push(regular_record),
+        }
+    }
```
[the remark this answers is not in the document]
tertsdiepraam: We discussed that wasn't correct right?

src/units/http_server.rs
```diff
@@ -1343,3 +1287,25 @@ impl HttpServer {
         Json(Err(()))
     }
 }
+
+#[allow(clippy::borrowed_box)]
+pub async fn read_soa(
+    read: &Box<dyn ReadableZone>,
+    qname: Name<Bytes>,
+) -> Result<Option<(Soa<Name<Bytes>>, Ttl)>, OutOfZone> {
```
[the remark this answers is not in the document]
tertsdiepraam: I left it because it was in the current zone loader, but I agree

src/loader/mod.rs
tertsdiepraam: Instructions unclear :) It should but do we we want that right now?

.github/workflows/system-tests.yml
```diff
@@ -187,6 +185,81 @@ jobs:
...
+          until target/debug/cascade zone status example.test | grep -q "Published zone available"; do
+            if (($(date +%s) > (start + timeout))); then
+              echo "timeout: zone status did not report published zone available" >&2
+              exit 1
+            fi
+            sleep 1
+          done
+      - name: Query zone
+        run: |
+          dig @127.0.0.1 -p 4543 example.test AXFR
```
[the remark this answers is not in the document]
tertsdiepraam: Oh it didn't, I forgot to copy that part. It does now.

src/loader/mod.rs
```diff
@@ -0,0 +1,664 @@
...
+    fn enqueue_refresh(self: &Arc<Self>, zone: &Arc<Zone>) {
+        let mut state = zone.state.lock().unwrap();
+        LoaderState::enqueue_refresh(&mut state, zone, true, self);
+    }
+
+    fn set_source(self: &Arc<Self>, zone: &Arc<Zone>, source: Source) {
+        let mut state = zone.state.lock().unwrap();
+        state.loader.source = source;
+        LoaderState::enqueue_refresh(&mut state, zone, true, self);
+    }
```
bal-e: Are you sure the `reload` parameter to `enqueue_refresh()` should be `true` here?  That would prohibit the use of IXFRs.

tertsdiepraam: Well it's all a mess now. Because refreshes aren't supported by zonefiles. My idea was that we'd want to force a fully new version whenever we change the source or start up to be sure of a correct version.

tertsdiepraam: I could bring back refreshes for zonefiles (implemented as reloads) or something. The required flow is honestly unclear to me at this point.

src/loader/zonefile.rs
```diff
@@ -0,0 +1,200 @@
...
+        inplace::Entry::Record(record) => record,
+        inplace::Entry::Include { .. } => return Err(Error::UnsupportedInclude),
+    };
+
+    let record_name = record.owner();
+
+    record.compose(buf).unwrap();
+    let record = Record::<_, BoxedRecordData>::parse_bytes(buf)
+        .expect("'Record' serializes records correctly")
+        .transform(|name: RevNameBuf| name.unsized_copy_into(), |data| data);
```
bal-e: I don't remember whether the original code had this, but could you clear `buf` again after this?  It's nice to clear at the start and end of usage, to avoid accidental spillage.

PR #408, opened 2025-12-12
Add a test that verifies that DNSSEC signed zones are accepted and re-signed by Cascade.

### LINKED ISSUE #407: NSEC3PARAM RR is not stripped when signing.
If an input zone file contains an apex NSEC3PARAM RR but is signed with an NSEC policy (NOT NSEC3) then the resulting signed zone still has the NSEC3PARAM RR in it which casues `dnssec-verify` to complain about missing NSEC3 RRs.

ximon18:
Includes `dnssec-verify` output that shows that the zone is DNSSEC valid.
```
| Loading zone 'example.test' from file 'example-test.axfr.log'
|
| Verifying the zone using the following algorithms:
| - ECDSAP256SHA256
| Zone fully signed:
| Algorithm: ECDSAP256SHA256: KSKs: 1 active, 0 stand-by, 0 revoked
|                             ZSKs: 1 active, 0 stand-by, 0 revoked
```

The test could check more things, e.g. that the original DNSSEC records are indeed gone and replaced by new ones.

### DIFF
--- a/integration-tests/strip-dnssec-rrs/example.test
+++ b/integration-tests/strip-dnssec-rrs/example.test
@@ -1,0 +1,15 @@
+; This zone contains DNSSEC records that the signer should strip out.
+example.test.           3600    IN      SOA     ns1.example.test. hostmaster.example.test. 2025121201 1800 900 604800 86400
+example.test.           3600    IN      NSEC3PARAM 1 0 0 -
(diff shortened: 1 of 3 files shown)


withjannisNLnetLabs (approved):
LGTM!

PR #417, opened 2026-01-07
Overhaul how zone data is stored

bal-e:
Zone data is a complicated topic, and Cascade currently doesn't handle it well. Not only does it store many copies of each zone unnecessarily, but we have noticed several race conditions around it. This PR introduces a `cascade-zonedata` crate with a fresh approach to memory-efficient, thread-safe zone data storage.

Concurrent access to the zone data is important; e.g. Cascade should always serve the latest published version of the zone. At the same time, new instances of the zone may be created, as new versions are loaded and signatures are re-generated. New instances shouldn't require a whole copy of the zone data; they should be stored efficiently as diffs.

`cascade-zonedata` provides an API that will support this nicely. The overall flow for using it is quite complex, and I'm still looking for good ways to enforce correctness via Rust's type system; but I am confident that we can implement highly optimized data structures for zone storage within this API.


crates/zonedata/src/viewer.rs
```diff
@@ -0,0 +1,197 @@
...
+/// [`ZoneViewer`] provides access to (the signed and unsigned components of)
+/// the _authoritative_ or _upcoming_ instance of a zone.
+pub struct ZoneViewer {
+    /// The underlying data.
+    ///
+    /// [`ZoneViewer`] has a read lock over the signed and unsigned components
+    /// of the authoritative or upcoming instance (depending on `self.curr`).
+    data: Arc<AuthData>,
+
+    /// Whether the authoritative instance is being read.
```
tertsdiepraam: This comment is a bit confusing. The field is also always true according to RA, but that's fine.

crates/zonedata/src/reader.rs
```diff
@@ -0,0 +1,49 @@
...
+    }
+}
+
+//----------- SignedZoneReader -------------------------------------------------
+
+/// A reader for a signed instance of a zone.
+pub struct SignedZoneReader<'d> {
+    /// The underlying data.
+    pub(crate) data: &'d InstanceHalf,
+}
```
tertsdiepraam: Why is this two different types? I fear that will inhibit some useful abstractions over the readers. Is the reason that you want to change the structure in a way that requires them to be different types? That might be good to note.

crates/zonedata/src/writer.rs
```diff
@@ -0,0 +1,689 @@
...
+
+    /// The records to add.
+    added_records: Vec<(RegularRecord, usize)>,
+
+    /// The patchset counter.
+    ///
+    /// Multiple patches (i.e. sets of added/removed records) can be processed;
+    /// this is incremented between patches, so it is unique, and later patches
+    /// are preferred over earlier ones.
+    patchset: usize,
```
tertsdiepraam: What's the use case for this?

tertsdiepraam (changes requested):
This is tough. I don't like this very much but then I don't know if there's any implementation of a system like this that I would like, so take this with a grain of salt.

What I don't like is all the ways that using this API uncarefully can panic, especially around the custom read-write lock mechanisms. That's worrying, but it's better than deadlocking or UB. And it will probably get better with more careful management of the state to prevent problems with trying to acquire multiple locks. It just currently feels like half an abstraction if that makes sense, because it feels fairly low level. In the spirit of moving things forward though, we should probably merge this fairly soon and then just move towards implementing the rest.

The cleanups that happen as a result of this in the main cascade crate are very good by the way. It does show that this API has a lot of potential and is really good in practice.

Apart from that, I do have some questions about it and some things that should be clarified in comments, but nothing major.

crates/zonedata/src/writer.rs
tertsdiepraam: Mention in a comment?

src/zone/storage.rs
```diff
@@ -0,0 +1,419 @@
...
+
+        // Commit the update with the SOA record.
+        let soa: cascade_zonedata::OldRecord = reader.soa().clone().into();
+        Self::force_future(updater.apply(ZoneUpdate::Finished(soa))).unwrap();
+
+        zone
+    }
+
+    /// Force a [`Future`] to evaluate synchronously.
+    fn force_future<F: IntoFuture>(future: F) -> F::Output {
```
tertsdiepraam: Is this here to work around some stuff being async? So it's temporary?

PR #433, opened 2026-01-20
Suggest `ods2cascade` for OpenDNSSEC users.

ximon18:


bal-e (approved):

ximon18:
Actually I think this should wait until open `ods2cascade` PRs have been merged and it has been packaged and released, as this PR points to installation instructions which don't work until a package is published.

PR #442, opened 2026-02-02
Picky correctness improvement to inline TOML comment in policy template.

ximon18:
`send-notify-to` requires quoted `IP:[port]` values but the TOML doc doesn't mention the quotes, and unlike the server address binding examples in the main config file, here there is no default value making the syntax self-documenting. It also says a collection while strictly speaking it should be a set, and to be even more pedantic I noted that it should be comma-separated.


bal-e (commented):
I agree that the comment should be more descriptive, but I'm worried about the phrase "comma-separated". For anybody who is familiar with TOML, "comma-separated" can also cause confusion. Perhaps we should just provides some concrete examples (while also adding quotations around "IP:[port]")?

ximon18:
> I agree that the comment should be more descriptive, but I'm worried about the phrase "comma-separated". For anybody who is familiar with TOML, "comma-separated" can also cause confusion. Perhaps we should just provides some concrete examples (while also adding quotations around "IP:[port]")?

I made some changes and synced with `main`, please re-review.

bal-e (approved):

PR #445, opened 2026-02-04
App cmd refactor

tertsdiepraam:
Removes `ApplicationCommand` entirely in favor of direct method calls on the units. All the units now live under manager and have a similar structure to the `ZoneLoader` in that they have a `new` and `run` method instead of `init`. The `new` method initializes the unit and the `init` starts whatever background task the unit needs returning an `AbortOnDrop` handle for it.

As a result, I was able to remove all but 3 channels being created by returning values from the methods on the units directly. There are now no `oneshot` channels left only some `mpsc` channels that served other purposes.

To facilitate this change, I changed some `RwLock`s and `Mutex`es in the signer to be the versions from `std` that don't need await, but also can't be hold onto past `await` points. We could go even further with that in the future by also removing the `Semaphore`s.

This PR should not alter any functionality.


src/manager.rs
```diff
@@ -196,49 +140,49 @@ impl Manager {
                 }
 
                 // Inform all units about the change.
-                for name in ["ZL", "RS", "KM", "ZS", "RS2", "PS"] {
-                    self.on_app_cmd(name, ApplicationCommand::Changed(change.clone()));
-                }
-                return;
+                let center = &self.center;
```
bal-e: Can we move this line above the match statement, so it doesn't have to be duplicated in every arm?

PR #456, opened 2026-02-13
Remove `Update` and `Change` enums

tertsdiepraam:
Removes the broadcasting `Update` mechanism, which wasn't really used for broadcasting anyways. The most difficult changes are around the config changes and reloading the policies.

### DIFF
--- a/src/zone/storage.rs
+++ b/src/zone/storage.rs
@@ -14,8 +14,7 @@ use domain::zonetree;
 use tracing::{info, trace, trace_span, warn};
 
 use crate::{
-    center::{Center, get_zone},
-    manager::record_zone_event,
+    center::Center,
     util::AbortOnDrop,
     zone::{HistoricalEvent, PipelineMode, Zone, ZoneHandle, ZoneState},
 };
@@ -105,6 +104,15 @@ impl StorageZoneHandle<'_> {
 
                 let (s, ureviewer) = s.finish(built);
                 *machine = ZoneDataStorage::ReviewLoadedPending(s);
+
+                // TODO: Use the instance ID here, which will not require
+                // examining the zone contents.
+                let serial = ureviewer.read_loaded().unwrap().soa().rdata.serial;
+                self.state.record_event(
+                    HistoricalEvent::NewVersionReceived,
+                    Some(domain::base::Serial(serial.into())),
+                );
+
                 self.start_loaded_review(ureviewer);
             }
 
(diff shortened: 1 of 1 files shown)


src/policy/mod.rs
```diff
@@ -38,34 +37,76 @@ pub struct Policy {
...
+        }
+    }
+
+    // Update the set of policies.
+    *policies = new_policies;
 
+    Ok(updated)
+}
+
+pub fn load_all(
```
bal-e: It would be nice to have a doc comment here! Also mentioning that `policies` is read in for logging purposes (the actual contents of existing policies are not examined).

src/state/v1.rs
```diff
@@ -91,10 +88,7 @@ impl Spec {
                 }
             })
             .collect();
-        for zone in state.zones.drain() {
-            info!("Removing zone '{}'", zone.0.name);
-            (on_change)(Change::ZoneRemoved(zone.0.name.clone()));
-        }
+
```
bal-e: Shouldn't we still log these zone removals? Or are they being logged elsewhere now (which I missed)?

src/state/v1.rs
```diff
@@ -137,20 +131,12 @@ impl PolicySpec {
...
-            let old = core::mem::replace(&mut policy.latest, new.clone());
-            (on_change)(Change::PolicyChanged(old.clone(), new.clone()));
-            for zone in &policy.zones {
-                (on_change)(Change::ZonePolicyChanged {
-                    name: zone.clone(),
-                    old: Some(old.clone()),
-                    new: new.clone(),
-                });
-            }
+            let _ = core::mem::replace(&mut policy.latest, new.clone());
```
bal-e: This can just be `policy.latest = Arc::new(latest)`.

bal-e (commented):
This is really cool! I'm really excited to see us moving to an all-sync-function-call paradigm. Overall, I like the way you've addressed the policy reloading issue. Aside from my few comments, we should be able to merge this PR soon.

src/state/v1.rs
```diff
@@ -39,7 +39,7 @@ pub struct Spec {
 
 impl Spec {
     /// Parse from this specification.
-    pub fn parse_into(self, state: &mut State, mut on_change: impl FnMut(Change)) {
+    pub fn parse_into(self, state: &mut State) {
```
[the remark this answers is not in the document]
tertsdiepraam: Feel free to, but I think it's out of scope for this PR.

src/zone/mod.rs
```diff
@@ -473,74 +467,6 @@ pub fn save_state_now(center: &Center, zone: &Zone) {
     }
 }
 
-/// Change the policy used by a zone.
-pub fn change_policy(
-    center: &Arc<Center>,
-    name: Name<Bytes>,
-    policy: Box<str>,
-) -> Result<(), ChangePolicyError> {
```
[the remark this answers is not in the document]
bal-e: I expect we'll need it later ... maybe you can just comment it out? Looking up the code when we re-introduce the function would be quite frustrating otherwise.

PR #485, opened 2026-02-27
Integrate unsigned zone approval with the new zone storage

bal-e:
This allows the `SignedZoneBuilder` to be obtained at the right time, and is the first step for integrating the zone signer with the new zone storage.

### DIFF
--- a/src/zone/storage.rs
+++ b/src/zone/storage.rs
@@ -7,8 +7,8 @@
 use std::{fmt, sync::Arc};
 
 use cascade_zonedata::{
-    LoadedZoneBuilder, LoadedZoneBuilt, LoadedZoneReader, LoadedZoneReviewer, SignedZoneReviewer,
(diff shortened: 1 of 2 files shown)


src/zone/storage.rs
```diff
@@ -188,11 +188,28 @@ impl StorageZoneHandle<'_> {
...
                 }
                 PipelineMode::HardHalt(_) => {
+                    // TODO: Is this the right behavior?
                     warn!("Not reviewing newly-loaded instance because pipeline is hard-halted");
                     false
                 }
             };
 
+            // TODO: Pass on the reviewer to the zone server.
+            let old_ureviewer = std::mem::replace(&mut state.storage.loaded_reviewer, ureviewer);
```
tertsdiepraam: `ureviewer`? Is that `unsigned_reviewer`? Maybe clarify that a bit.

Issue #493, opened 2026-03-02
IXFR-in fails with error "a patchset did not change the SOA record"


ximon18:
_**Note:** I believe this issue does NOT affect the alpha-5 release as it is caused by code changes since alpha-5 in the Git `main` branch, specifically the new zone loader._

---

Given the following IXFR served by NSD:

```
example.test.		5	IN	SOA	ns1.example.test. mail.example.test. 2 60 60 3600 5 <- Start of IXFR for serial 2
example.test.		5	IN	SOA	ns1.example.test. mail.example.test. 1 60 60 3600 5 <- Start of deletions from serial 1
www         A   169.254.1.1
example.test.		5	IN	SOA	ns1.example.test. mail.example.test. 2 60 60 3600 5 <- Start of addiitions in serial 2
www         A   192.168.0.1
example.test.		5	IN	SOA	ns1.example.test. mail.example.test. 2 60 60 3600 5 <- End of IXFR for serial 2
```

Cascade incorrectly rejects the IXFR with:

```
| 2026-03-02T14:35:07.112263Z  INFO refresh{zone=example.test source=Server { addr: 127.0.0.1:1055, tsig_key: None }}: cascaded::loader: Refreshing Name(example.test.)
| 2026-03-02T14:35:07.112284Z DEBUG refresh{zone=example.test source=Server { addr: 127.0.0.1:1055, tsig_key: None }}:refresh{zone=example.test addr=127.0.0.1:1055}: cascaded::loader::server: Refreshing Name(example.test.) from server 127.0.0.1:1055
| 2026-03-02T14:35:07.112298Z TRACE refresh{zone=example.test source=Server { addr: 127.0.0.1:1055, tsig_key: None }}:refresh{zone=example.test addr=127.0.0.1:1055}: cascaded::loader::server: Attempting an IXFR against 127.0.0.1:1055 for Name(example.test.)
| 2026-03-02T14:35:07.112335Z DEBUG refresh{zone=example.test source=Server { addr: 127.0.0.1:1055, tsig_key: None }}:refresh{zone=example.test addr=127.0.0.1:1055}:ixfr{zone=example.test addr=127.0.0.1:1055}: cascaded::loader::server: Attempting an IXFR against 127.0.0.1:1055 for Name(example.test.)
[...]

PR #497, opened 2026-03-03
Prevent deadlock due to lack of `approve_loaded()`

bal-e:
If review is not enabled, `approve_loaded()` was not called, causing a deadlock in Cascade. The `approve_loaded()` call has been moved so that it occurs whether review is enabled or not.

Thanks for catching and debugging this @ximon18!

### DIFF
--- a/src/zone/storage.rs
+++ b/src/zone/storage.rs
@@ -15,7 +15,7 @@ use tracing::{info, trace, trace_span, warn};
 
 use crate::{
     center::Center,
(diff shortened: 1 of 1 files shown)


ximon18 (approved):
`./act-wrapper --job ixfr-in` works for me again with this PR.

It still fails due to https://github.com/NLnetLabs/cascade/issues/493 but no longer for other reasons.

PR #499, opened 2026-03-03
Prepare new control flow around zone signing

bal-e:
This PR introduces `src/signer`, a module into which the zone signer module should eventually be refactored. It introduces the `signer::zone` model, which works very similarly to `loader::zone`; signing and re-signing operations can be enqueued and started when possible.

Signing operations need to be enqueued because the backing zone storage might be busy. I have built the enqueuing logic to account for limiting simultaneous signing operations, but that code was hard to extract from `zone_signer.rs`. I'll get to it later.

At the moment, none of the new functions are used anywhere. They need to wait until the actual signing code is adapted to use the `SignedZoneBuilder`, otherwise they risk deadlocking the zone storage state machine. The PR performing the actual integration will deal with that.

### DIFF
--- a/src/zone/storage.rs
+++ b/src/zone/storage.rs
@@ -75,7 +75,7 @@ impl StorageZoneHandle<'_> {
     ///
     /// - [`Self::finish_load()`] when loading succeeds.
     ///
-    /// - [`Self::give_up_load()`] when loading fails.
+    /// - [`Self::abandon_load()`] when loading fails.
     ///
     /// If the zone data storage is busy, [`None`] is returned; the loader
     /// should enqueue the load operation and wait for a passive notification.
@@ -320,7 +320,7 @@ impl StorageZoneHandle<'_> {
     ///
     /// - [`Self::finish_sign()`] when signing succeeds.
     ///
-    /// - [`Self::give_up_sign()`] when signing fails.
+    /// - [`Self::abandon_sign()`] when signing fails.
     ///
     /// If the zone data storage is busy, [`None`] is returned; the
     /// signer should enqueue the re-sign operation and wait for a passive


src/signer/zone.rs
```diff
@@ -0,0 +1,258 @@
...
+            };
+
+            let handle = tokio::task::spawn(super::sign(
+                self.center.clone(),
+                self.zone.clone(),
+                builder,
+                trigger,
+            ));
+
+            self.state.signer.ongoing = Some(handle.into());
```
ximon18: Same comment as above.

src/zone/storage.rs
```diff
@@ -311,6 +312,239 @@ impl StorageZoneHandle<'_> {
...
+        let machine = &mut self.state.storage.machine;
+        match machine.take() {
+            ZoneDataStorage::Signing(s) => {
+                trace!(
+                    zone = %self.zone.name,
+                    "Giving up on the ongoing (re-)sign"
+                );
+
+                let (s, loaded_reviewer) = s.give_up(builder);
+                // TODO: Communicate the new reviewer handle to the zone server.
```
ximon18: What is the "new reviewer handle" (is that `self.state.storage.loaded_reviewer`?) and why and how would one communicate it to the zone server?

src/zone/storage.rs
```diff
@@ -311,6 +312,239 @@ impl StorageZoneHandle<'_> {
...
+            // Read the signed instance.
+            let signed_reader = signed_reviewer
+                .read_signed()
+                .unwrap_or_else(|| unreachable!("The signer never returns an empty instance"));
+            let serial = signed_reader.soa().rdata.serial;
+
+            // Build a `zonetree` for the new instance.
+            let zonetree = Self::build_signed_zonetree(&zone, &loaded_reader, &signed_reader);
+
+            // Insert the new `zonetree`.
```
ximon18: This seems to do a replace, not an insert.

src/zone/storage.rs
```diff
@@ -311,6 +312,239 @@ impl StorageZoneHandle<'_> {
...
+            );
+
+            state = zone.state.lock().unwrap();
+
+            state.storage.background_tasks.finish()
+        });
+    }
+
+    /// Build a `zonetree` for an signed instance of a zone.
+    fn build_signed_zonetree(
```
ximon18: This returns a one, not a zone tree.

src/zone/storage.rs
```diff
@@ -311,6 +312,239 @@ impl StorageZoneHandle<'_> {
...
+
+        // Add every record in turn.
+        for record in signed_reader.records() {
+            let record: cascade_zonedata::OldParsedRecord = record.clone().into();
+            force_future(updater.apply(ZoneUpdate::AddRecord(record))).unwrap();
+        }
+
+        // Add every loaded record in turn (excluding SOA).
+        //
+        // TODO: Which other records to exclude? DNSKEY, RRSIGs?
```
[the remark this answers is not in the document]
bal-e: 1. `.records()` does not include the SOA record, that is only provided by `.soa()`. So we never add the SOA record of the loaded instance of the zone.

2. The loaded instance of the zone may contain RRSIGs and DNSKEY records, which the signed instance is supposed to replace (e.g. in pass-through mode). At the moment, this code will add those records from the loaded instance of the zone instead of hiding them.

src/signer/zone.rs
```diff
@@ -0,0 +1,258 @@
...
+        // task.
+
+        let handle = tokio::task::spawn(super::sign(
+            self.center.clone(),
+            self.zone.clone(),
+            builder,
+            SigningTrigger::ZoneChangesApproved,
+        ));
+
+        self.state.signer.ongoing = Some(handle.into());
```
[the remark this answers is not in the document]
bal-e: It isn't possible because the spawned operation will lock the mutex before stopping; it cannot lock the mutex until the caller (that is setting `.ongoing`) releases its lock. I will document that.

src/signer/zone.rs
```diff
@@ -0,0 +1,258 @@
...
+    /// has been enqueued, it will be initiated, and `true` will be returned.
+    ///
+    /// This method cannot initiate enqueued signing operations; when a signing
+    /// operation is enqueued, it prevents the data storage from idling.
+    pub fn start_pending(&mut self) -> bool {
+        // An enqueued or ongoing signing operation holds a 'SignedZoneBuilder',
+        // which prevents the zone data storage from being idle. This method is
+        // only called if the zone data storage is idle.
+        assert!(self.state.signer.enqueued_sign.is_none());
+        assert!(
```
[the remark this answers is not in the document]
bal-e: I think you've misunderstood the assertion. This line asserts that 1) there is no pending re-sign; **or** 2) that the re-sign does not have a `builder` field. It will panic if there **was** a pending-resign with a non-empty `builder` field ... because of the comment on top.

src/zone/storage.rs
```diff
@@ -311,6 +312,239 @@ impl StorageZoneHandle<'_> {
...
+            ),
+        }
+    }
+
+    /// Give up on the ongoing signing operation.
+    ///
+    /// Intermediate artifacts in the signed instance, and the upcoming loaded
+    /// instance (if any), will be cleaned up automatically, in the background.
+    /// Once the zone storage is idle, a notification will be sent.
+    pub fn give_up_sign(&mut self, builder: SignedZoneBuilder) {
```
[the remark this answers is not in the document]
bal-e: Yes, I like that suggestion! I'll rename `give_up` to `abandon` across this module.

crates/cli/src/commands/zone.rs
```diff
@@ -388,29 +388,49 @@ impl Zone {
...
                                         match trigger {
-                                            SigningTrigger::ExternallyModifiedKeySetState =>
-                                                "externally modified keyset state",
-                                            SigningTrigger::SignatureExpiration =>
-                                                "pending signature expiration",
-                                            SigningTrigger::ZoneChangesApproved =>
-                                                "unsigned zone review approved",
-                                            SigningTrigger::KeySetModifiedAfterCron =>
-                                                "keyset cron modified keyset state",
+                                            SigningTrigger::Load => "loading a new instance",
```
ximon18: Nit: This block and the block below look identical. Factor them out? I'm fine if you leave it as this is a knock-on effect of changes for prior review feedback and perfect is the enemy of getting this merged ;-)

PR #500, opened 2026-03-04
Integrate the zone signer with the new zone storage

bal-e:
Building on top of #499, this PR makes the core `sign_zone()` function use the new zone storage. It adjusts the rest of Cascade to call into the new signer flow functions (from #499) since they are now properly implemented.

For reviewers: please check the signing flow, e.g. that the signer correctly launches signed review when it finishes.

**NOTE**: While developing this PR, I noticed a subtle logic error in our parallelized zone signing code: it arbitrarily splits the unsigned records into segments to pass to `sign_sorted_zone_records()`, but that function assumes it has received the full zone in order to locate zone cuts. I have retained this bug in the new code; we need to copy over `domain`'s signing code and explicitly parallelize it. This effort would be greatly simplified with better zone data structures, so perhaps that should be addressed first.

### DIFF
--- a/src/units/zone_signer.rs
+++ b/src/units/zone_signer.rs
@@ -81,7 +81,6 @@ pub struct ZoneSigner {
     // TODO: Discuss whether this semaphore is necessary.
     max_concurrent_operations: usize,
     concurrent_operation_permits: Arc<Semaphore>,
-    max_concurrent_rrsig_generation_tasks: usize,
     signer_status: ZoneSignerStatus,
     kmip_servers: Arc<Mutex<HashMap<String, SyncConnPool>>>,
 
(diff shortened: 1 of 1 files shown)


src/units/zone_signer.rs
ximon18: While this was in the original code, so not part of this PR, why call `new_from_owned`, can't this just use `unsigned_records.len()`?

src/signer/mod.rs
```diff
@@ -54,38 +53,59 @@ pub mod zone;
...
+        state: &mut state,
+        center: &center,
+    };
+    handle.state.signer.ongoing.finish();
+
+    match result {
+        Ok(()) => {
+            let built = builder.finish().unwrap_or_else(|_| unreachable!());
+            handle.storage().finish_sign(built);
+            status.status.finish(true);
```
ximon18: Possible improvement here (can be later): currently one could mistakenly set, or forget to set, `status.current_action = "Aborted"`, I wonder if that could be incorporated into the line above somehow that calls `status.status.finish()`?

PR #505, opened 2026-03-09
Add a system test to check that receiving multipart AXFR works as expected.

ximon18:
This test is currently passing.

The following error was seen sometimes when the PR was first created but has not been seen recently:

```
| 2026-03-09T14:36:39.039859Z  INFO refresh{zone=example.test source=Server { addr: 127.0.0.1:1055, tsig_key: None }}: cascaded::loader: Refreshing Name(example.test.)
| 2026-03-09T14:36:39.039887Z DEBUG refresh{zone=example.test source=Server { addr: 127.0.0.1:1055, tsig_key: None }}: cascaded::loader::server: Refreshing Name(example.test.) from server 127.0.0.1:1055
| 2026-03-09T14:36:39.039912Z DEBUG refresh{zone=example.test source=Server { addr: 127.0.0.1:1055, tsig_key: None }}: cascaded::loader::server: Attempting an IXFR against 127.0.0.1:1055 for Name(example.test.)
| 2026-03-09T14:36:39.040245Z ERROR refresh{zone=example.test source=Server { addr: 127.0.0.1:1055, tsig_key: None }}: cascaded::daemon: thread 'cascade-worker' (ProcessId(219), ThreadId(61)) panicked at src/loader/server.rs:231:14: internal error: entered unreachable code
```

The test could be improved, to actually detect multipart AXFR was handled (I verified manually by inserting some extra logging temporarily) and that the entirety of the large zone is received and published (currently it just does a line count check).


integration-tests/tests/multipart-axfr/action.yml
```diff
@@ -0,0 +1,82 @@
...
+            cascade zone status example.test
+            echo "::error:: timeout: zone status did not report published zone available"
+            exit 1
+          fi
+          sleep 1
+        done
+
+    # TODO: Verify that Cascade actually handled more than one AXFR part.
+    # Currently there is no NSD or Cascade metric that can be used to check
+    # this.
```
Philip-NLnetLabs, line 69: You should be able to use ZONEMD to let ldns-verify-zone check in the unsign review hook that the zone was received correctly.

ximon18: Done. This still won't tell us if the zone was received in multiple parts or not however.

integration-tests/multipart-axfr/example.test.zone
```diff
@@ -0,0 +1,519 @@
...
+@           NS  example.test.
+@           NS  ns1.example.test.
+@           A   127.0.0.1
+ns1         A   127.0.0.1
+
+; Define an in-memory zone to be served by the server. The zone is large
+; enough that it cannot be fit into a single TCP AXFR response and so must
+; be split by the server across multiple responses.
+$ORIGIN example.test.
+txt0000       3600  IN  CNAME  some.very.long.domain.name.that.takes.up.a.lot.of.bytes.so.as.to.fill.the.dns.response.with.as.few.rrs.as.possible.
```
bal-e: Name compression might reduce the size of these RRs. Maybe you could vary the last label in each owner name? That gives the zone a much better chance of requiring multi-part AXFR.

integration-tests/tests/multipart-axfr/action.yml
```diff
@@ -0,0 +1,108 @@
...
+
+    - name: Query the zone at the unsigned preview server
```
bal-e: Typo in "preview"? Technically it _is_ a preview, but "review" would be more consistent here IMO.

bal-e (commented):
Maybe the test could be named `multipart-axfr` to clarify exactly what it is trying to test? It currently says `-large` but the zone isn't _that_ large, compared to the TLD zones we expect Cascade to be used with.

integration-tests/tests/multipart-axfr/action.yml
ximon18: This test was added long ago back when I think we actually called it the preview server. I'll update it. Good catch.

integration-tests/multipart-axfr/example.test.zone
ximon18: The zone was taken from an [automated test in domain](https://github.com/NLnetLabs/domain/blob/main/test-data/server/axfr_large_out.rpl) of its split AXFR handling and was shown there to cause splitting to occur so AFAIK it does result in a multi-part response.

ximon18:
> Maybe the test could be named `multipart-axfr` to clarify exactly what it is trying to test? It currently says `-large` but the zone isn't _that_ large, compared to the TLD zones we expect Cascade to be used with.

I think you make a valid point. FYI in this case the name was just copied from the test that I copied [from domain](https://github.com/NLnetLabs/domain/blob/main/test-data/server/axfr_large_out.rpl).

integration-tests/multipart-axfr/example.test.zone
bal-e: That's really confusing...

ximon18: > Name compression might reduce the size of these RRs. Maybe you could vary the last label in each owner name? That gives the zone a much better chance of requiring multi-part AXFR.

Is it better now?

bal-e (approved):
Cool!!

PR #512, opened 2026-03-10
Add a system test that checks that a signed zone is re-used on re-start.

ximon18:
I.e. that we persist signed zones to disk and reload and reuse the persisted zone instead of re-signing the zone on startup.

Currently fails as after restart Cascade no longer lists the zone in the output of `cascade zone list`.


ximon18:
Obsoleted by PR #550.

PR #517, opened 2026-03-10
Refactor many get_zone instances away

### LINKED ISSUE #511: Crash if NOTIFY received for non-existing zone.
```
| 2026-03-09T14:22:20.120637Z  INFO domain::net::server::middleware::notify: NOTIFY received from 127.0.0.1:49540 for zone 'example.test' with serial Some(Serial(1))
| 2026-03-09T14:22:20.120676Z  INFO cascaded::units::zone_server: [CC]: Instructing zone loader to refresh the zone
| 2026-03-09T14:22:20.120701Z ERROR cascaded::daemon: thread 'cascade-worker' (ProcessId(184), ThreadId(64)) panicked at src/loader/mod.rs:89:64: zone exists
| note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

tertsdiepraam:
This should reduce the possibility for deadlocks a bit and reduces the contention on the `center.state` mutex.

Some calls are just move from the callee to the caller, because then they happen more at the "edge". For example, the http server now does more of these calls because it needs to translate the names from the API to actual zones and it makes sense to do that there.

Builds on https://github.com/NLnetLabs/cascade/pull/504 but doesn't merge into it because I think that shouldn't be merged directly.

### DIFF
--- a/integration-tests/system-tests.yml
+++ b/integration-tests/system-tests.yml
@@ -54,7 +54,6 @@ defaults:
   run:
     # see: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defaultsrunshell
     shell: bash --noprofile --norc -eo pipefail -x {0}
-  
 
 jobs:
   test-release-version:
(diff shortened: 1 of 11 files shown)


bal-e (approved):
Looks great!

bal-e (approved):

PR #520, opened 2026-03-12
Make `State` load only from scratch and not into existing state

tertsdiepraam:
This allows us to get rid of some tracking of whether policies and zones are added, retained, removed or updated.


bal-e (approved):
Awesome! Simplifying this makes it easier for us to control the zone initialization code (which would be useful for resolving some workarounds in #552, although we'll get to them later).

PR #528, opened 2026-03-17
Zone instances and IDs

bal-e:
This PR introduces `zone/instance.rs` for tracking zone instances. It introduces `{Loaded,Signed}InstanceID` as a way to uniquely identify loaded and signed instances of zones, respectively; unlike SOA serial numbers, instance IDs are guaranteed to be unique (even if two instances have the same SOA serial, or if an instance is rejected). The rest of the codebase has been integrated with this system, so IDs are correctly available at every step (except review to some extent; I will work on it next).

The new state for instances will eventually replace the `.{unsigned,signed}` fields of `ZoneState`. They will accumulate more instance-specific data, e.g. loader/signer statistics, diffs (for persistence, rollback, and IXFR out), maybe even logs/history.

---

### DIFF
--- a/src/loader/mod.rs
+++ b/src/loader/mod.rs
@@ -25,7 +25,7 @@ use crate::{
     common::scheduler::Scheduler,
     loader::zone::EnqueuedRefresh,
     util::AbortOnDrop,
-    zone::{Zone, ZoneByPtr, ZoneHandle},
+    zone::{Zone, ZoneByPtr, ZoneHandle, instance::LoadedInstanceID},
 };
 
 mod server;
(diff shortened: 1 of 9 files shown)


bal-e:
Converting to draft while I consider adding some more state.

ximon18 (commented):
I'm having trouble reviewing this as I'm wondering why this introduces another state machine, and how that relates to the storage and zone state machines. The actual code is not that complex, but I think it would be good to sit with @bal-e and review it together to better understand the thinking behind this PR.

bal-e:
Closing in favor of #617.

PR #538, opened 2026-03-20
Add prometheus metrics for zone transfers and loaded zone bytes and records

mozzieongit:
Add metrics for zone transfers, labelled by transfer type (AXFR vs IXFR),
transport (UDP vs TCP), and zone name.

Metric names are not set in stone and maybe need some work. Maybe now, maybe
later when we decide how to represent metrics for different states of a zone
(e.g. the number of records in a zone differs between unsigned input and signed
output). Suggestions should consider that we probably want to track outgoing
transfers to downstream nameservers too.

This PR also adds the build's commit id to the cascade_build_info metric.

---

Querying the metrics in the failing `ixfr-in` test:
```
| # HELP cascade_build Cascade build information.
| # TYPE cascade_build info
| cascade_build_info{version="0.1.0-alpha5",build="8c2ebb243c-dirty"} 1
...
| # HELP cascade_xfr_requests_to_upstream_attempted Number of zone transfers attempted by Cascade towards the upstream primary.
| # TYPE cascade_xfr_requests_to_upstream_attempted counter
| cascade_xfr_requests_to_upstream_attempted_total{zone="example.test",xfrtype="AXFR",transport="TCP"} 1
| cascade_xfr_requests_to_upstream_attempted_total{zone="example.test",xfrtype="IXFR",transport="UDP"} 1
| # HELP cascade_xfr_requests_to_upstream_succeeded Number of zone transfers succeeded by Cascade towards the upstream primary.
| # TYPE cascade_xfr_requests_to_upstream_succeeded counter
| cascade_xfr_requests_to_upstream_succeeded_total{zone="example.test",xfrtype="AXFR",transport="TCP"} 1
| # HELP cascade_zone_loaded_last_successful_records Number of records loaded in last successful zone transfer or zonefile load.
| # TYPE cascade_zone_loaded_last_successful_records gauge
| cascade_zone_loaded_last_successful_records{zone="example.test"} 8
| # HELP cascade_zone_loaded_last_successful_bytes Number of bytes loaded in last successful zone transfer or zonefile load.
| # TYPE cascade_zone_loaded_last_successful_bytes gauge
| # UNIT cascade_zone_loaded_last_successful_bytes bytes
| cascade_zone_loaded_last_successful_bytes{zone="example.test"} 248
| # HELP cascade_zone_loaded_last_records Number of records loaded in last successful zone transfer or zonefile load.
| # TYPE cascade_zone_loaded_last_records gauge
| cascade_zone_loaded_last_records{zone="example.test"} 4
| # HELP cascade_zone_loaded_last_bytes Number of bytes loaded in last successful zone transfer or zonefile load.
| # TYPE cascade_zone_loaded_last_bytes gauge
| # UNIT cascade_zone_loaded_last_bytes bytes
| cascade_zone_loaded_last_bytes{zone="example.test"} 230
```

`cascade_zone_loaded_last` shows the last load even if it failed:

```
cascade_zone_loaded_last_successful_records{zone="example.com"} 3
cascade_zone_loaded_last_successful_bytes{zone="example.com"} 434
cascade_zone_loaded_last_records{zone="example.com"} 5
cascade_zone_loaded_last_bytes{zone="example.com"} 635
```

on a successful load both show the same
```
cascade_zone_loaded_last_successful_records{zone="example.com"} 5
cascade_zone_loaded_last_successful_bytes{zone="example.com"} 602
cascade_zone_loaded_last_records{zone="example.com"} 5
cascade_zone_loaded_last_bytes{zone="example.com"} 602
```

---

    - (Only the two known-to-fail tests ixfr-in and incremental-signing are failing)


tertsdiepraam (commented):
I think this is pretty good, but I find it hard to judge. I'd like it if the `prometheus_metrics` didn't have to be passed to all the functions, but that's just what needs to happen than that's what we should do.

tertsdiepraam (approved):

tertsdiepraam:
I'm reviving this PR. I fixed the conflicts and will now start looking at the code more in depth to see whether it's still doing the correct thing.

(Requested review from myself to remove the approval, since I'll now be author instead of reviewer)

src/loader/mod.rs
```diff
@@ -39,13 +44,55 @@ mod zonefile;
...
+        metrics.register_with_unit(
+            "zone_loaded_last_successful",
+            "Number of bytes loaded in last successful zone transfer or zonefile load",
+            Unit::Bytes,
+            loader_metrics.zone_loaded_last_successful_bytes.clone(),
+        );
+
+        metrics.register(
+            "zone_loaded_last_records",
+            "Number of records loaded in last successful zone transfer or zonefile load",
```
ximon18: I think this would be more consistent with other metric descriptions:

```suggestion
            "Number of records loaded in last attempted zone transfer or zonefile load",
```

src/loader/mod.rs
```diff
@@ -39,13 +44,55 @@ mod zonefile;
...
+            loader_metrics.zone_loaded_last_records.clone(),
+        );
+
+        metrics.register_with_unit(
+            "zone_loaded_last",
+            "Number of bytes loaded in last successful zone transfer or zonefile load",
```
ximon18: I think this would be more consistent with other metric descriptions:

```suggestion
            "Number of bytes loaded in last attempted zone transfer or zonefile load",
```

ximon18: I think this would be more consistent with other metric descriptions:

```suggestion
            "Number of bytes loaded in last attempted zone transfer or zonefile load",
```

src/loader/mod.rs
```diff
@@ -39,13 +44,55 @@ mod zonefile;
...
+
+        metrics.register(
+            "xfr_requests_to_upstream_attempted",
+            "Number of zone transfers attempted by Cascade towards the upstream primary",
+            loader_metrics.xfr_requests_to_upstream_attempted.clone(),
+        );
+
+        metrics.register(
+            "xfr_requests_to_upstream_succeeded",
+            "Number of zone transfers succeeded by Cascade towards the upstream primary",
```
ximon18: I think this would be more consistent with the descriptions of the other metrics:

```suggestion
            "Number of successful zone transfers by Cascade towards the upstream primary",
```

src/loader/mod.rs
```diff
@@ -428,6 +513,27 @@ impl ActiveLoadMetrics {
     }
 }
 
+//----------- LoaderPrometheusMetrics ----------------------------------------
+
+/// Prometheus metrics for the zone loader.
+#[derive(Debug, Default)]
+struct LoaderPrometheusMetrics {
```
ximon18, line 520: Maybe remove all references to `prometheus` in this type name and variable/field names that refer to it? Yes it is currently only used for Prometheus metric reporting but maybe that won't be the case in the future, and the use case it is put to is for the caller / consumer to decide, not the type itself. I find the distinction somewhat artificial because while `LoadMetrics` is not currently published toward Prometheus clients, a field like `LoadMetrics::duration` definitely could be interesting as a Prometheus metric. Removing '`prometheus' would also make the type name and references to the type shorter making code less verbose.

src/metrics.rs
```diff
@@ -275,21 +276,59 @@ impl EncodeLabelValue for StoredName {
...
+pub enum HaltMode {
     HardHalt,
 }
 
+//------------ XfrLabels -----------------------------------------------------
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq, EncodeLabelSet)]
+pub struct XfrLabels {
+    pub zone: StoredName,
+    pub xfrtype: XfrType,
```
ximon18: Perhaps rename `xfrtype`, especially the name used for this field when it is published as a Prometheus metric, to something that doesn't include "xfr" as context makes it clear  that it relates to XFR.

src/metrics.rs
```diff
@@ -73,15 +73,16 @@ impl MetricsCollection {
         // software build information via labels and will always be 1. It
         // cannot be stored inside of `MetricsCollection` as it does not
         // implement Clone.
-        let _cascade_version = Info::new(vec![("version", clap::crate_version!())]);
+        let _cascade_version = Info::new(vec![
+            ("version", clap::crate_version!()),
+            ("build", env!("CASCADE_BUILD_COMMIT")),
```
ximon18: Maybe rename "build" to "commit" as I find a field called "build" inside a metric named "build" redundant and not adding any value?

src/metrics.rs
```diff
@@ -275,21 +276,59 @@ impl EncodeLabelValue for StoredName {
...
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq, EncodeLabelValue)]
+pub enum XfrType {
+    AXFR,
+    IXFR,
+}
+
+//------------ XfrTransport --------------------------------------------------
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq, EncodeLabelValue)]
```
ximon18: Nit. This could derive `Copy` too.
```suggestion
#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq, EncodeLabelValue)]
```

src/metrics.rs
```diff
@@ -275,21 +276,59 @@ impl EncodeLabelValue for StoredName {
...
+#[derive(Debug, Clone, Hash, PartialEq, Eq, EncodeLabelSet)]
+pub struct XfrLabels {
+    pub zone: StoredName,
+    pub xfrtype: XfrType,
+    pub transport: XfrTransport,
+}
+
+//------------ XfrType -------------------------------------------------------
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq, EncodeLabelValue)]
```
ximon18: Nit. This could derive `Copy` too.
```suggestion
#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq, EncodeLabelValue)]
```

src/metrics.rs
```diff
@@ -275,21 +276,59 @@ impl EncodeLabelValue for StoredName {
...
-    zone: StoredName,
-    mode: HaltMode,
+pub struct ZoneHaltMode {
+    pub zone: StoredName,
+    pub mode: HaltMode,
 }
 
 //------------ HaltMode ------------------------------------------------------
 
 #[derive(Debug, Clone, Hash, PartialEq, Eq, EncodeLabelValue)]
```
ximon18: Nit. This could derive `Copy` too.
```suggestion
#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq, EncodeLabelValue)]
```

Changelog.md
```diff
@@ -22,6 +22,9 @@ Released yyyy-mm-dd.
 
 ### New
 
+- Add prometheus metrics for zone transfers and loaded zone bytes and records.
+  ([#538] by @mozzieongit)
```
ximon18:
```suggestion
  ([#538] by @mozzieongit and @tertsdiepraam)
```

src/metrics.rs
```diff
@@ -73,15 +73,16 @@ impl MetricsCollection {
         // software build information via labels and will always be 1. It
         // cannot be stored inside of `MetricsCollection` as it does not
         // implement Clone.
-        let _cascade_version = Info::new(vec![("version", clap::crate_version!())]);
+        let _cascade_version = Info::new(vec![
+            ("version", clap::crate_version!()),
+            ("build", env!("CASCADE_BUILD_COMMIT")),
+        ]);
 
         // The prometheus docs linked to
```
ximon18: Should this say?
```suggestion
        // See the prometheus docs at
```

ximon18 (changes requested):
LGTM, some comments, mainly nits, only one actual issue ("successful" included in a metric description where I think it should not be).

src/loader/mod.rs
tertsdiepraam: I agree in principle, but I think removing `Prometheus` right now would create more confusion because `LoadMetrics` already exists and that would make the names too similar.

The distinction right now is that `LoadMetrics` applies to a single load while the `LoaderMetrics` applies to the entire loader (so multiple zones and multiple loads).

This is unfortunate and confusing, but better fixed in another PR I think. Also note that the fields in this type have types from the `prometheus_client` crate so it is kind of specifically for Prometheus.

ximon18: Yes I realized that the types used are from the Prometheus crate but chose to consider than an implementation detail. If you have doubts then leave it, I find it a bit clunky but I guess we have more to improve in how we update, gather and publish metrics anyway.

src/loader/mod.rs
```diff
@@ -39,13 +44,55 @@ mod zonefile;
...
+
+        metrics.register(
+            "xfr_requests_to_upstream_attempted",
+            "Number of zone transfers attempted by Cascade towards the upstream primary",
+            loader_metrics.xfr_requests_to_upstream_attempted.clone(),
+        );
+
+        metrics.register(
+            "xfr_requests_to_upstream_succeeded",
+            "Number of succeeded zone transfers by Cascade towards the upstream primary",
```
ximon18: "Number of succeeded" sounds really odd to me and is inconsistent with "Number of zone transfer attempted".

ximon18 (approved):
Just one metric description remaining that I find odd, but I don't want to block this work because of that.

src/loader/mod.rs
tertsdiepraam: Is "successful" better?

src/loader/mod.rs
tertsdiepraam: I think you'll like my PR that's gonna build on this one!

PR #550, opened 2026-03-26
Persist zones

### LINKED ISSUE #295: Zone persistence
Cascade internally has three different versions of a zone: an unsigned version, a signed version, and a published version.  At the moment, none of these are persisted; upon a restart, Cascade has to fetch an unsigned zone and sign it immediately.  This can cause a number of issues; zone loading failure could block Cascade and zones might be resigned unnecessarily.

We need to decide when and how zones get persisted and where necessary implement saving and restoring.

ximon18:
# Status

Zone edits (e.g. via file edit and `zone reload` or due to changes received via XFR) are persisted both for loaded and signed zones. On application restart persisted zones are restored and `cascade zone status` and `dig AXFR` appear to work as they did prior to application shutdown.

The code has been initially reviewed and led to some wanted changes:
- [x] @bal-e will extend the `zonedata` crate interface to enable moving some of the changes to Cascade itself.
- [x] Invoke `save_now()` instead of `mark_dirty()`. This is because persisting zone data to disk not enough, as if there was a power outage between calling `mark_dirty()` and publication of the zone the paths to the persisted files (that are recorded in state) would not have been written to disk, thus `save_now()` should be invoked instead. We may actually want to invoke `save_now()` in the publication server itself to ensure all published zone related state is persisted, but that is out of scope for this PR.

# tl;dr

- Extend `src/persistence/persist.rs` to write snapshots and deltas to disk (currently in DNS AXFR/IXFR wire format).
- Extend `src/persistence/restore.rs` to restore zones by loading the zone snapshot into a zone replacer and passing the zone diffs one at a time to a zone patcher.
- Extend `src/persistence/` to store generated/restored diffs for use by IXFR out.
- Extend `ZoneState` with two vecs of paths to persisted snapshot and delta files.
  - _Note: https://github.com/NLnetLabs/cascade/pull/657 moves these paths into dedicated persistence related state which I didn't notice existed when I created this PR._
- Extend the CLI to show the zone is restoring in `cascade zone status` output.

## Known issues:
- No support (yet) for condensing/compacting/purging old persisted deltas.
  - _See https://github.com/NLnetLabs/cascade/pull/657_
- Binary snapshot/delta format is not inspectable by operators. We could add a `cascade debug <xxx> subcommand to render a snapshot/delta in XFR presentation format but that is not included in this PR.
- There's no "scheduling" or spreading out of zone restoration activity, like refreshing it happens as soon as possible.
- There are no metrics for the restoration process.
- Snapshots and deltas are loaded entirely into memory, while they could be processed in a streaming manner instead.
- The persistence location is hard-coded to store files in the same directory as `zone-state`.

# Alternative snapshot/delta storage formats

A choice would need to be driven by requirements:
- Do we need to support IXFR out at the loaded review nameserver? If so, is just one delta enough? Actually do we even need to keep a loaded diff, a review hook would only need it for the next received zone update, not after restart.
- What do we tune for? Memory usage? Persistence speed? Restoration speed?

Some ideas:

- Presentation XFR format: was not done in this PR due to the pain of having to convert new base to old base to render in presentation format, and parsing DNS records in presentation format is only currently properly supported by domain/cascade for entire zones, though one could use the "treat a single RR line as a zonefile" hack that the Stelline code uses.
- (De)serialize cascade `DiffData` to JSON/CBOR/whatever. Would tie the format to the internal format, changing the internal format would require writing additional code to handle storage format migration or support parsing old as well as new format. A PRO of this approach is being able to storage additional meta data with the snapshot/delta, if needed.

---


crates/zonedata/src/storage/transitions.rs
```diff
@@ -286,6 +295,79 @@ impl ReviewingLoadedStorage {
...
+                true,
+                Some(loaded_built.diff),
+                Some(signed_built.diff),
+            )
+        };
+
+        let viewer = unsafe { ZoneViewer::new(self.data.clone(), true, true) };
+
+        let signed = unsafe { &*self.data.signed[0].get() };
+        let serial = signed.soa.as_ref().unwrap().rdata.serial;
```
bal-e: I don't think the serial number should be extracted here. Perhaps we need methods on `LoadedZoneBuilt` and `SignedZoneBuilt` that give you readers so you can observe the data, or just getters for the serial?

src/loader/mod.rs
```diff
@@ -264,6 +266,85 @@ async fn refresh(
...
+async fn restore(
+    zone: Arc<Zone>,
+    loaded_source: PathBuf,
+    loaded_count: usize,
+    signed_source: PathBuf,
+    signed_count: usize,
+    restorer: PersistedZoneRestorer,
+    center: Arc<Center>,
+    // metrics: Arc<ActiveLoadMetrics>,
+) {
```
bal-e: I don't think this function belongs in the zone loader. Perhaps a dedicated `persistence` module or the like would be a better fit.

src/loader/zone.rs
```diff
@@ -111,6 +113,60 @@ impl LoaderZoneHandle<'_> {
...
+    pub fn restore(&mut self) -> &mut Self {
+        if !self.state.persisted_loaded_diffs.is_empty()
+            && self.state.loader.ongoing_restore.is_none()
+        {
+            // Determine the paths to read from. Each zone is persisted as an
+            // AXFR plus zero or more IXFRs. The restorer takes a base path
+            // ending in an unsigned integer number and loads that file plus
+            // N more, where the final number in the path is replaced by the
+            // previous number plus one each time.
+            let unsigned_source = format!("/tmp/{}.loaded.0", self.zone.name);
```
bal-e: Yep, we definitely want to change this. I guess we want a field in `Config` to specify where zone data should be persisted?

src/zone/storage.rs
```diff
@@ -86,7 +105,16 @@ impl StorageZoneHandle<'_> {
...
-        let (transition, state) = transition(&mut self.state.storage.machine);
+        let (t, state) = transition(&mut self.state.storage.machine);
+
+        let (transition, state) = if let ZoneDataStorage::Uninitialized(s) = state {
+            let s = s.initialize();
+            t.move_to(ZoneDataStorage::Passive(s));
+            transition(&mut self.state.storage.machine)
+        } else {
+            (t, state)
+        };
```
bal-e: I'd prefer the above commented-out `initialize()` function over this silent automatic conversion. Especially since it should only occur at startup and we _can_ handle it outside this function.

bal-e: With the current setup, we might actually want to do this check in other places, but it is hard to tell.

bal-e (commented):
@ximon18 and I had a discussion while I was writing this review, and decided that I will add commits changing the PR to implement a new approach for the zone data storage state machine. I've left these comments here for us to come back once I'm done; I also didn't review everything, so there will be more to come.

src/persistence/restore.rs
```diff
@@ -1,14 +1,34 @@
...
+    utils::dst::UnsizedCopy,
+};
+use tracing::{info, trace};
 
 use crate::{center::Center, zone::Zone};
 
 /// Restore the loaded instance data of a zone.
 #[tracing::instrument(
-    level = "trace",
+    level = "info",
```
tertsdiepraam: Was this for debugging? Should we put it back on trace or keep it like this?

src/persistence/zone.rs
```diff
@@ -195,6 +197,35 @@ impl ZonePersistenceHandle<'_> {
     }
 }
 
+fn clear_persisted_zone_data(center: &Center, state: &mut ZoneState) {
```
tertsdiepraam, line 252: I could see this going slightly wrong if somebody wants to debug something with the persistence first. But I guess we're just kind of assuming that the persistence will be ok.

tertsdiepraam (commented):
I'll approve after trying it out, but the code looks pretty good.

tertsdiepraam:
image

I seem to have broken something. Presumably this was because I had some very old zone state file that interfered, but that still indicates that somewhere in the error handling something is going wrong.

ximon18:
> image
> 
> 
> 
> I seem to have broken something. Presumably this was because I had some very old zone state file that interfered, but that still indicates that somewhere in the error handling something is going wrong.

That looks like an issue on the state machine rather than in the persist-restore changes, is that right?

tertsdiepraam:
I'm not sure where the actual cause it, but it definitely is triggered by this PR because this PR starts returning `Ok` from `restore_loaded`.

Maybe you have to call `Self::clear` if there are no files to read instead of doing nothing? Or move the `if !self.persisted_loaded_diff_paths.is_empty()` check out of `restore_loaded` so that you won't try a restore if there are no diffs.

You can reproduce it by clearing out the `persisted_loaded_diff_paths` in the zone state. Not a huge issue of course. Seems to work fine otherwise.

src/zone/mod.rs
```diff
@@ -258,6 +259,16 @@ pub struct ZoneState {
     /// History of interesting events that occurred for this zone.
     pub history: Vec<HistoryItem>,
 
+    /// Locations of persisted unsigned zone diffs to enable IXFR from
+    /// the upstream to resume on restart, and to enable a complete latest
+    /// unsigned version of the zone to be reconstituted.
+    pub persisted_loaded_diff_paths: Vec<PathBuf>,
```
ximon18, line 266: Note to self: maybe these should be moved into `PersistenceState` (which I only just noticed that @bal-e added).

src/persistence/restore.rs
ximon18: I don't recall, I'll put it back to trace.

ximon18:
> Maybe you have to call `Self::clear` if there are no files to read instead of doing nothing? Or move the `if !self.persisted_loaded_diff_paths.is_empty()` check out of `restore_loaded` so that you won't try a restore if there are no diffs.

Or just pass back `Ok(())` and leave the `restorer` as-is? When `restore_loaded()` is invoked the `restorer` is in the state created by `LoadedZoneRestorer::new()` which is not quite the same as the state if `LoadedZoneRestorer::clear()` is invoked, the difference being a `None` diff in the former case and a `Some(empty diff)` in the latter case?

src/zone/mod.rs
ximon18: I left a TODO comment about this for now in the code.

src/persistence/zone.rs
```diff
@@ -195,6 +247,35 @@ impl ZonePersistenceHandle<'_> {
...
+    // the corresponding files on disk.
+    let zone_state_dir = center.config.zone_state_dir.canonicalize().unwrap();
+    for p in state
+        .persisted_loaded_diff_paths
+        .iter()
+        .chain(state.persisted_signed_diff_paths.iter())
+    {
+        if p.exists()
+            && let Ok(ref p) = p.canonicalize()
+            && p.starts_with(&zone_state_dir)
```
bal-e: What is the purpose of this canonicalization? Why would a path be outside the zone state dir?

src/units/http_server.rs
```diff
@@ -474,7 +474,13 @@ impl HttpServer {
...
 
             progress = match zone_state.machine {
-                ZoneStateMachine::Waiting(..) => Progress::Waiting,
+                ZoneStateMachine::Waiting(..) => {
+                    if zone_state.storage.is_restoring() {
+                        Progress::Restoring
+                    } else {
+                        Progress::Waiting
+                    }
+                }
```
bal-e, line 483: Not for now, but maybe we should make `Restoring` an explicit part of the zone state machine.

src/zone/machine.rs
```diff
@@ -265,27 +265,27 @@ impl<'a> ZoneHandle<'a> {
...
+    //     let (transition, state) = self.state.machine.transition();
+    //     let ZoneStateMachine::Waiting(waiting) = state else {
+    //         panic!("'start_sign_after_restore()' called when the zone was not passive");
+    //     };
+    //     transition.move_to(ZoneStateMachine::Signing(
+    //         waiting.start_sign_after_restore(),
+    //     ));
+
+    //     self.signer().enqueue_new_sign(builder);
+    // }
```
bal-e: I think it's fine to delete this function completely.

bal-e (approved):

src/persistence/zone.rs
ximon18: Removed after discussing.

src/zone/machine.rs
ximon18: Removed.

PR #552, opened 2026-03-26
Integrate the zone server(s) with the new zone storage

bal-e:
This PR integrates the last of Cascade's major components, the zone server, with the new zone storage. A new `ZoneService` type is added, which implements `domain::net::server::service::Service`; it stores a `LoadedZoneReviewer`/`SignedZoneReviewer`/`ZoneViewer` and returns data from that. This type is used in place of the previous `zone_server_service()` function, which read from zonetrees.

`src/zone/storage.rs` has been modified to exchange viewers with these zone server units when a zone undergoes a change. This process is `async`, which complicates the control flow a bit; I hope to simplify them soon.

This change also interacts with how zones are initialized; at initialization, the zone (re)viewer types are held in `ZoneState` and are extracted by the zone server units. This is a hack to work around design issues in zone initialization, and they should be addressed in the future. This also interacts with zone persistence.

This PR removes the last major use of zonetrees within Cascade. However, it does not remove the zonetrees yet; they will be removed in a future PR (once some minor uses are also resolved).

---

Note that the `ixfr-in` test is currently failing, but for apparently minor reasons (the new zone server only supports AXFR and SOA queries, but the test attempts A and MX queries). The test will be changed in a separate PR, _before_ this PR is merged.

### DIFF
--- a/src/server/service.rs
+++ b/src/server/service.rs
@@ -1,0 +1,319 @@
+//! Servicing DNS requests.
+
+use std::sync::{Arc, RwLock};
+
+use cascade_zonedata::{LoadedZoneReviewer, SignedZoneReviewer, SoaRecord, ZoneViewer};
+use domain::{
+    new::base::{
+        name::{RevName, RevNameBuf},
+        wire::ParseBytes,
+    },
+    utils::dst::UnsizedCopy,
+};
+
+use crate::zone::Zone;
+
+//----------- ZoneService ------------------------------------------------------
+
+/// Cascade's DNS service.
+///
+/// This type is responsible for answering DNS queries received by Cascade,
+/// using zone viewer objects. When a query is received, the corresponding zone
+/// is looked up, the appropriate information is retrieved from its viewer, and
+/// the appropriate DNS answer is synthesized and returned.
+///
+/// The viewer type has to implement [`Viewer`]. It should only be
+/// [`LoadedZoneReviewer`], [`SignedZoneReviewer`], or [`ZoneViewer`].
+///
+/// The service can be interacted with through a [`ZoneServiceHandle`], which
+/// is also returned by [`ZoneService::new()`].
+pub struct ZoneService<V> {
+    /// The underlying state.
+    //
+    // TODO: The state is currently wrapped in an 'RwLock'. This is necessary
+    // as the state might have to change, e.g. in response to changes to zones.
+    // It would be preferable to hold the state directly, and use channels to
+    // communicate the need for changes. This is currently impossible because
+    // of the limitations of 'domain::net::server'; that architecture should be
+    // gradually replaced locally for better flexibility and efficiency.
+    state: Arc<std::sync::RwLock<ZoneServiceState<V>>>,
+}
+
+impl<V> ZoneService<V> {
+    /// Construct a new [`ZoneService`].
+    ///
+    /// In addition to the service, a [`ZoneServiceHandle`] is returned through
+    /// which the service can be interacted with.
+    pub fn new() -> (ZoneService<V>, ZoneServiceHandle<V>) {
+        let state = Arc::new(std::sync::RwLock::default());
+        let service = ZoneService {
+            state: state.clone(),
+        };
+        let handle = ZoneServiceHandle { state };
+        (service, handle)
+    }
+}
+
+impl<V> Clone for ZoneService<V> {
+    fn clone(&self) -> Self {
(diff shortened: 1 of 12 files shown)


src/server/service.rs
```diff
@@ -0,0 +1,303 @@
...
+
+    fn soa(&self) -> &SoaRecord {
+        self.read().unwrap().soa()
+    }
+}
+
+//----------- ZoneServiceHandle ------------------------------------------------
+
+/// A handle for controlling a [`ZoneService`].
+pub struct ZoneServiceHandle<V> {
```
tertsdiepraam: At this point in reviewing I'm wondering why the handle is needed if it contains the same field as `ZoneService`. Maybe the next commits will clear that up.

src/zone/storage.rs
```diff
@@ -772,19 +880,25 @@ pub struct StorageState {
     machine: ZoneDataStorage,
 
     /// The current loaded zone reviewer.
+    ///
+    /// This is only used during initialization.
     //
-    // TODO: Move into the zone server unit.
-    loaded_reviewer: LoadedZoneReviewer,
+    // TODO: Output it directly somehow?
```
tertsdiepraam: Output what directly?

tertsdiepraam (approved):
I think this looks good. There's not much that I'm really surprised by. I think you wanted a second opinion on it and that's probably a good idea, but the commits tell a good story and if it passes the integration tests then this should be good.

src/units/zone_server.rs
```diff
@@ -202,8 +205,7 @@ impl ZoneServer {
             center: center.clone(),
         };
 
-        // let svc = ZoneServerService::new(zones.clone());
-        let svc = service_fn(zone_server_service, zones.clone());
+        let svc = service;
         let svc = XfrMiddlewareSvc::new(svc, zones.clone(), max_concurrency);
```
ximon18: Why are you still using `XfrMiddlewareSvc` if you have implemented XFR support in your new ZoneService service impl?

src/server/service.rs
```diff
@@ -0,0 +1,416 @@
...
+                    match zone_request.kind {
+                        ZoneRequestKind::Soa => Box::pin({
+                            let viewer = zone.viewer.clone();
+                            async move {
+                                let viewer = viewer.read_owned().await;
+                                soa(old_request.message(), &*viewer)
+                            }
+                        }) as Response,
+
+                        ZoneRequestKind::Axfr => {
```
ximon18: This new code doesn't check the transport context in order to reject XFR over UDP like the original XfrMiddlewareSvc code does per https://datatracker.ietf.org/doc/html/rfc5936#section-4.2.

ximon18 (commented):
Some initial feedback, unfortunately I don't have time to continue further at this point.

src/server/service.rs
```diff
@@ -0,0 +1,416 @@
...
+    async fn axfr<V: Viewer + Send + Sync + 'static>(
+        request: Request<Vec<u8>, Option<Arc<tsig::Key>>>,
+        zone: ServedZone<V>,
+    ) -> ResponseStream {
+        // Obtain a read lock to read the zone for an extended duration.
+        let viewer = zone.viewer.read_owned().await;
+
+        if viewer.is_empty() {
+            // The zone is empty, so we don't have anything to serve.
+            return error(request.message(), Rcode::NXDOMAIN);
```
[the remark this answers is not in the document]
bal-e: This is not unexpected; it should also be NOAUTH as per your comment.

src/server/service.rs
```diff
@@ -0,0 +1,416 @@
...
+            state: self.state.clone(),
+        }
+    }
+}
+
+/// A compatibility layer to [`domain::net::server`].
+///
+/// In the future, the network server stack should be gradually inlined here,
+/// so it can use [`domain::new`] and support more functionality (e.g. handling
+/// XFRs by spawning OS threads).
```
ximon18: In what way does `domain::net::server` prevent using a thread here if wanted? I'm not convinced that "should" is the right term here, I don't believe we as a team discussed the changes you have in mind for `net::server` yet, I'd prefix "the network server stack should be gradually inlined here" with "I think that".

src/zone/storage.rs
bal-e: This code will be addressed by the changes I will make to #550, so I'm going to leave it alone here.

PR #566, opened 2026-04-02
Also include CDS and CDNSKEY RRsets.

Philip-NLnetLabs:
Include CDS and CDNSKEY RRsets in the apex.
---

### DIFF
--- a/src/units/zone_signer.rs
+++ b/src/units/zone_signer.rs
@@ -476,6 +476,18 @@ impl ZoneSigner {
             }
         }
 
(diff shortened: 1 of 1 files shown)


ximon18 (approved):
LGTM. I'm not sure if anything is missing or if it fits in with the surrounding code but as a diff I don't see an issue with it. I haven't tested it, do you have a test somewhere that relates to this?

Philip-NLnetLabs:
It tested it manually. At some point we should have key manager tests in Cascade, but that depends on better fake time.

PR #568, opened 2026-04-07
Change how global options are referred to in man pages.

ximon18:
As it is confusing to see [OPTIONS] .. [OPTIONS] defined where [OPTIONS] in one position refers to a subset of the [OPTIONS] in the other position.

Also includes a couple of unrelated minor fixes.

Provide your description here

---


bal-e (approved):
Looks fine overall. Just a few small notes.

PR #572, opened 2026-04-09
[zone/machine] Allow abandoning a signing operation

bal-e:
This is needed for incremental signing; the incremental signer might discover that a zone does not require any changes.

### DIFF
--- a/src/zone/machine.rs
+++ b/src/zone/machine.rs
@@ -33,6 +33,9 @@ use crate::{
 /// the pipeline anyway. `SigningFailure` cannot be overridden but only `reset`.
 ///
 /// Here is the diagram for it:
+//
+// TODO: There is an additional transition from 'Signing' to 'Waiting', in case
+// signing is abandoned (e.g. incremental signing turns out to be a no-op).
 ///
 /// ```text
 /// ┌──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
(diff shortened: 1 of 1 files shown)


src/zone/machine.rs
```diff
@@ -33,6 +33,9 @@ use crate::{
 /// the pipeline anyway. `SigningFailure` cannot be overridden but only `reset`.
 ///
 /// Here is the diagram for it:
+//
+// TODO: There is an additional transition from 'Signing' to 'Waiting', in case
+// signing is abandoned (e.g. incremental signing turns out to be a no-op).
```
tertsdiepraam, line 38: I'm not doing it. That's up to you to fix ;p

src/zone/machine.rs
```diff
@@ -252,6 +255,22 @@ impl<'a> ZoneHandle<'a> {
         self.storage().finish_sign(built);
     }
 
+    // Abandon the ongoing signing operation (but not due to failure).
+    #[expect(dead_code)]
+    pub(crate) fn abandon_signing(&mut self, builder: SignedZoneBuilder) {
```
tertsdiepraam, line 260: I wonder if we can come up with a better name? "signing_not_needed"? "skip_signing"?

tertsdiepraam (approved):
I suppose this comes from a suggestion by @Philip-NLnetLabs? Looks good to me although the name could be better I think.

src/zone/machine.rs
bal-e: Well, we need to decide whether we want names in passive or imperative form. I am not against a change, as long as it's consistent.

PR #574, opened 2026-04-10
Reduce git conflicts and verbose test definitions.

ximon18:
Combine the almost identical system test definitions into a single job that invokes them as a matrix requiring only that new test names be added to the 'tests' matrix key rather than duplicating the entire job block every time, and each PR that adds a system test then conflicting with the others due to each modifying the same part of system-tests.yml.

Notes added to `act-wrapper` help output:
- To limit which tests are run by the new tests job one must now supply --matrix tests:<TEST NAME TO RUN> instead of --job <JOB NAME TO RUN>..
- To limit the number of concurrent 'tests' sub-jobs one must now supply --input concurrent-tests:1

The tests fail when run, not sure if that also happens on `main` at present.

---

### DIFF
--- a/act-wrapper
+++ b/act-wrapper
@@ -34,6 +34,7 @@ function usage {
 	echo "Inputs that can be provided via --input:"
 	echo "	--input build-profile=debug|release"
 	echo "	--input log-level=error|warn|info|debug|trace"
+	echo "  --input concurrent-tests=1"
 	echo ""
 	echo "Use --matrix test:<TEST NAME TO RUN> to limit which jobs the 'tests' job runs."
 }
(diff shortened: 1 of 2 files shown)


mozzieongit:
> and each PR that adds a system test then conflicting with the others due to each modifying the same part of system-tests.yml.

That will still be the case. Now, conflicting in modifying the same matrix definition. There is no way to avoid that conflict other than not all adding to the bottom, but some adding in between other tests (which requires them to be their own blocks instead of a matrix value).

ximon18:
True but it's much less irritating to resolve.

PR #577, opened 2026-04-13
[metrics] Stop using zonetrees

bal-e:
This PR removes the last usage of the zonetrees. They can be removed afterwards.

---

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -21,6 +21,8 @@ use prometheus_client::metrics::info::Info;
 use prometheus_client::registry::{Metric, Registry, Unit};
 
 use crate::center::Center;
+use crate::zone::ZoneByName;
+use crate::zone::machine::ZoneStateMachine;
 
 // Further metrics to track?:
 // - last time batching operation for zone signing succeeded (push to central metrics collection)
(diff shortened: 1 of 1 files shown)


src/metrics.rs
```diff
@@ -105,31 +107,52 @@ impl MetricsCollection {
...
+                    | ZoneStateMachine::LoadedReview(_)
+                    | ZoneStateMachine::HaltLoaded(_) => {
+                        zones_unsigned += 1;
+                    }
+
+                    ZoneStateMachine::Signing(_)
+                    | ZoneStateMachine::SigningFailed(_)
+                    | ZoneStateMachine::SignedReview(_)
+                    | ZoneStateMachine::HaltSigned(_) => {
+                        zones_signed += 1;
```
ximon18: If this is `Signing` should `zones_signed += 1` be done?

ximon18 (approved):
LGTM. I was able to remove all references to zone trees in my local copy of the branch and still compile and run cascade.

src/metrics.rs
bal-e: I guess not. Will fix.

PR #580, opened 2026-04-13
Introduce public-nameservers and send it to keyset. Handle TSIG related

Philip-NLnetLabs:
Introduce public-nameservers and send it to keyset. Handle TSIG related aspects as well.

---


src/policy/file/v1.rs
```diff
@@ -144,6 +144,13 @@ pub struct KeyManagerSpec {
 
     /// How keys are generated.
     pub generation: KeyManagerGenerationSpec,
+
+    /// The upstream nameservers to use when checking for RRSIG propagation
+    /// during a key roll. The value is a list of strings. Current each string
+    /// has the following syntax: <IP-address>:<port>[^<tsig-key-name>].
+    /// The port is mandatory. The TSIG key name is option and the name
+    /// of the key is preceeded by a caret character (^).
```
ximon18:
```suggestion
    /// The upstream nameservers to use when checking for RRSIG propagation
    /// during a key roll. The value is a list of strings. Each string has the following
    /// syntax: `<IP-address>:<port>[^<tsig-key-name>].`
    /// The port is mandatory. The TSIG key name is optional and the name
    /// of the key is preceded by a caret character (`^`).
```

src/policy/mod.rs
```diff
@@ -191,6 +197,36 @@ pub fn load_all(
...
+        // Get TSIG key name.
+        let Some(tsig_name) = iter.next() else {
+            continue;
+        };
+        let tsig_name = Name::from_str(tsig_name).map_err(|e| {
+            PolicyReloadError::Check(format!(
+                "unable to convert TSIG key name {tsig_name} to DNS name: {e}"
+            ))
+        })?;
+        tsig_store
```
ximon18: One issue with this is that this checks the store in memory while `dnst keyset` will check the store on disk, and a key present in memory may not yet have been written to disk, so should this PR also update `tsig::import_key()` to do a save now instead of invoking `mark_dirty()` (as the latter will only ensure the state is updated after a 5 second delay) ?

src/policy/mod.rs
```diff
@@ -191,6 +197,36 @@ pub fn load_all(
...
+            .get(&tsig_name)
+            .ok_or(PolicyReloadError::Check(format!(
+                "TSIG key {tsig_name} not found in TSIG store"
+            )))?;
```
ximon18: Using `split_once()` might be simpler:

```suggestion
        if let Some((_, tsig_name)) = n.split_once('^') {                                                                                                                                                                                                                                                              
            let tsig_name = Name::from_str(tsig_name).map_err(|e| {                                                                                                                                                                                                                                                    
                PolicyReloadError::Check(format!(                                                                                                                                                                                                                                                                      
                     "unable to convert TSIG key name {tsig_name} to DNS name: {e}"                                                                                                                                                                                                                                     
                ))                                                                                                                                                                                                                                                                                                     
            })?;                                                                                                                                                                                                                                                                                                       
            tsig_store                                                                                                                                                                                                                                                                                                 
                .get(&tsig_name)                                                                                                                                                                                                                                                                                       
                .ok_or(PolicyReloadError::Check(format!(                                                                                                                                                                                                                                                               
                    "TSIG key {tsig_name} not found in TSIG store"                                                                                                                                                                                                                                                     
                )))?;                                                                                                                                                                                                                                                                                                  
            }                                
```

ximon18 (approved):

src/policy/mod.rs
Philip-NLnetLabs: I'll locate that spot and try to force a flush.

Philip-NLnetLabs: Done

PR #583, opened 2026-04-13
Remove zonetrees!

bal-e:
It's finally here! After a months-long effort to introduce a new zone data storage, the old zonetrees are unused and can be removed.

Note: I tried to remove the `unstable-zonetree` feature from the `domain` dependency, but it turns out that we're still using bits of it (specifically `ZoneUpdate`) for XFR. I would like to trim the feature out entirely, but this will have to wait until the XFR interpreter is ported to `new-base`.

---


tertsdiepraam (approved):
And arc-swap gone too! I love it!

PR #586, opened 2026-04-15
Speed up system tests by reducing time spent during 'docker cp'. 

ximon18:
Relies on a symoblic link in `integration-tests/` to itself as lots of existing paths break without that and a trivial attempt to fix the paths didn't solve all the problems, while a symbolic link makes it 'just work' in my limited testing.

On `main` on my machine this reduces a 36 second full system test run (assuming no compilation required) to 18 seconds.

On my `tsig-downstream-support` branch the change is even more beneficial, reducing from 1 minute 58 seconds to 55 seconds.

The command I used to run the tests and time them was:

```
clear ; date > /tmp/start; time ./act-wrapper --rm --input log-level=info ; cat /tmp/start ; date
```

_(then manual timestamp subtraction ;-))_

I tried finding other solutions to this problem but `docker cp` ignores `.dockerignore` and `.gitignore`, the `act` `--use-gitignore` argument had no effect.

---


bal-e (approved):
Awesome, you know I love optimizations :D

Do we still need the file moves from `.github` to `integration-tests/.github`? Can we just have a `integration-tests/.github` -> `.github` symlink?

ximon18:
> Do we still need the file moves from `.github` to `integration-tests/.github`? Can we just have a `integration-tests/.github` -> `.github` symlink?

That might work but as these "GitHub" system-tests don't work on GitHub but only under nektos/act I thought it might actually make sense to move them under the system-tests.

bal-e:
> That might work but as these "GitHub" system-tests don't work on GitHub but only under nektos/act I thought it might actually make sense to move them under the system-tests.

That would be even better!

Issue #301, opened 2025-11-03
TSIG support


bal-e:
While Cascade was initially intended to support TSIG from the get-go, this feature got lost during the alpha development process.  A lot of the infrastructure necessary for it is already in the codebase, it just has to be tied back together.  A major concern is going to be the zone loader -- with the current architecture, it's hard to going to be difficult to integrate the new TSIG key store with it.  We may have to wait until a new zone loader (#15) is available.  TSIG also concerns outgoing zone transfers, for which some form of access control policy needs to be designed.

ximon18:
Currently existing TSIG support in `main`:
- TSIG key specification as JSON support exists in `main` (see https://github.com/NLnetLabs/cascade/blob/main/src/tsig/mod.rs).
- `main.rs` loads the store from the configured `tsig-store-path` resulting in an in-memory mapping of TSIG key names to key algorithm and key data bytes.
- `src/loader/server.rs` configures the transport to use a given TSIG key when doing AXFR/IXFR, and the key name to use is taken from zone state.
- `src/center` contains an `old_tsig_key_store` member field which uses a TSIG key store based on `domain` types, and this key store is made available in `src/units/zone_server.rs` to `XfrMiddlewareSvc` and `TsigMiddlewareSvc`.

What is missing regarding TSIG support in `main`:
- There is no way to specify which TSIG key should be used for a given upstream zone source. The key name to use **is** part of the zone source BUT `crates/api/src/lib.rs` defines `impl From<&str> for ZoneSource` which sets `tsig_key: None` and the CLI help string for `zone add` documents help "--source <SOURCE>" as "The zone source can be an IP address (with or without port, defaults to port 53) or a file path", i.e. no mention of a TSIG key.
- There is no way to add a key to the key store, e.g. no way to specify it via policy or config nor a CLI command, nor is a sample empty file with commented out sample values for example written to `tsig-store-path` so the user can't manually construct it, and `old_tsig_key_store` is never populated, `src/main.rs` sets it to `Default::default()` and it is not interacted with anywhere after this point of initial creation.
- There is no way to specify which key from `old_tsig_key_store`, were it to be populated, should be used with downstream secondary XFR clients. The closest we have is `server.outbound.send-notify-to` in `etc/policy.template.toml`.
- Any desired support for / integration with `dnst keyset` to tell it which TSIG key name relates to a given zone.
- Any desired support for / integration with `dnst keyset` to allow it to access the TSIG key algorithm and key data bytes from Cascade.

PR #587, opened 2026-04-15
TSIG downstream support

ximon18:
# Summary

This PR implements the last piece of work needed for #301: downstream TSIG support.

Downstream TSIG support consists of accepting TSIG signed XFR related requests sent by downstream nameservers to Cascadea nd replying with an appropriately signed response.

# Known issues

- TSIG state isn't properly restored on startup so that the key used by a zone will after restart become orphaned, the zone source is not properly reconstituted and so will lack the TSIG key. This issue wasn't caused by this PR but was found while testing it. It is fixed in separate PR https://github.com/NLnetLabs/cascade/pull/590.

# Changes made

CLI changes:

- Update `tsig list` to also show policies that use TSIG key(s) with downstreams, in addition to zones whose source uses a TSIG key with an upstream.

Policy changes:

- New `server.outbound.accept-xfr-from` policy setting. Actually it was already supported (under a slightly different name, `accept-xfr-requests-from`, but was not used anywhere).

Daemon changes:

- Extend `SimpleNameserverCommsSpec` from string parsing to handle a `[^<TSIG_KEY_NAME>]` suffix.
- Remove the unused `InboundPolicy` type, this is currently handled by the `--source` argument to the `zone add` CLI subcommand, not via policy.
- Enforce the `accept-xfr-from` ACL on incoming XFR requests handled by the zone server, rejecting requests that have an incorrect source IP address or mismatched TSIG key.
- Use the TSIG key defined in `send-notify-to` to sign NOTIFY messages sent by Cascade.
- Note: no changes were required to check XFR request TSIG keys or to TSIG sign responses as this was already handled by the existing `TsigMiddlewareSvc`.

Documentation changes:

- Extend the `Nameserver Integrations/NSD` page showing how to use NSD as a downstream, optionally with TSIG.

System test changes:

- Extend the secondary NSD configuration to include a TSIG `tsig-key` definition and new versions of the existing secondary zone which use TSIG for XFR and for NOTIFY+XFR.
- Adds a `downstream-tsig` system test which tests XFR and NOTIFY against NSD with and without TSIG keys for NOTIFY and XFR communication.

---


src/units/http_server.rs
```diff
@@ -1126,41 +1128,87 @@ impl HttpServer {
...
+                                zone_names: HashSet::new(),
+                                policy_names: [policy_name.to_string()].into(),
+                            });
+                        }
                     }
-                })
-                .collect::<Vec<ZoneName>>();
-            tsig_keys.insert(tsig_key_name.clone(), TsigListResultItem { zones });
+                }
+            }
```
bal-e, line 1252: I think this code intended to use `if tsig_key_names.peek().is_some()`, but it uses `.next()` accidentally and thus misses the first key name. I think the `if` can be removed and then the iterator does not need `.peekable()`.

src/units/zone_server.rs
```diff
@@ -769,16 +756,31 @@ pub fn send_notify_to_addrs(
...
+                nameserver
+                    .tsig_key_name
+                    .as_ref()
+                    .and_then(|tsig_key_name| state.tsig_store.get(tsig_key_name))
+                    .map(|key| key.inner.clone())
+            };
+
+            if let Some(key) = &tsig_key {
+                debug!(
+                    "Found TSIG key '{}' (algorith {}) for NOTIFY to {nameserver}",
```
bal-e: "algorith" => "algorithm"?

bal-e (commented):
LGTM aside from one minor bug.

PR #588, opened 2026-04-15
Instructions for building dnst

jpmens:
If I can believe the command output, there's a `dnst` missing in the description of building `dnst`:

```
$ cargo install --locked --bin dnst --git https://github.com/nlnetlabs/dnst
    Updating git repository `https://github.com/nlnetlabs/dnst`
error: multiple packages with binaries found: dnst, dnst-ldnsutils. When installing a git repository, cargo will always search the entire repo for any Cargo.toml.
Please specify a package, e.g. `cargo install --git https://github.com/nlnetlabs/dnst dnst`.
```

Hence this tiny PR.

### DIFF
--- a/doc/manual/source/building.rst
+++ b/doc/manual/source/building.rst
@@ -100,7 +100,7 @@ as simple as running:
 .. code-block:: bash
 
   cargo install --locked --git https://github.com/nlnetlabs/cascade cascade cascaded
-  cargo install --locked --bin dnst --git https://github.com/nlnetlabs/dnst
+  cargo install --locked --bin dnst --git https://github.com/nlnetlabs/dnst dnst
 
 The command will build Cascade and install it in the same directory that
 Cargo itself lives in, likely ``$HOME/.cargo/bin``. Ensure this directory is


ximon18:
Note that in our internal testing we build `dnst` from sources using `cargo install` but [use a slightly different syntax](https://github.com/NLnetLabs/cascade/blob/8a9403816d5ec9285d65140a64cd4f3dee27c430/integration-tests/cascade-test-image/Dockerfile#L54):

```
cargo install ... --bin=dnst dnst
```

jpmens:
okay ... but I don't understand the difference (and to be honest I don't know I actually want to understand :-)
Point is the documentation is, as far as I can tell, incorrect.

Kindly close this and patch the README whichever way you think is best.

ximon18 (approved):
The proposed change worked for me when tested locally.

ximon18:
> Note that in our internal testing we build `dnst` from sources using `cargo install` but [use a slightly different syntax](https://github.com/NLnetLabs/cascade/blob/8a9403816d5ec9285d65140a64cd4f3dee27c430/integration-tests/cascade-test-image/Dockerfile#L54):
> 
> ```
> cargo install ... --bin=dnst dnst
> ```

@mozzieongit: As you wrote the line in the `Dockerfile` I referenced above, do you know if there is a reason to include `--bin=dnst` in the instructions?

ximon18:
> okay ... but I don't understand the difference (and to be honest I don't know I actually want to understand :-) Point is the documentation is, as far as I can tell, incorrect.
> 
> Kindly close this and patch the README whichever way you think is best.

HI @jpmens,

My comment was more for internal communication than meant as a reply to you, sorry about that. We will of course fix the issue you found and kindly reported, thank you!

jpmens:
FWIW, if `--bin=dnst` is the same as `--bin dnst`, which I assume, then the line already contains that:

```
cargo install --locked --bin dnst --git https://github.com/nlnetlabs/dnst dnst
```

ximon18:
> FWIW, if `--bin=dnst` is the same as `--bin dnst`, which I assume, then the line already contains that:
> 
> ```
> cargo install --locked --bin dnst --git https://github.com/nlnetlabs/dnst dnst
> ```

Thanks, I didn't notice that, I believe you are right.

mozzieongit:
> FWIW, if `--bin=dnst` is the same as `--bin dnst`, which I assume, then the line already contains that:

Indeed, they are identical.

> @mozzieongit: As you wrote the line in the `Dockerfile` I referenced above, do you know if there is a reason to include `--bin=dnst` in the instructions?

We need to include `--bin` to only install dnst and not also install the test binary `ldns` we have in the dnst repository.

PR #567, opened 2026-04-07
Status rework

tertsdiepraam:
Reworks the `zone status` output to be clearer and more intentional in what information it displays.

It also makes a distinction between the last published zone and the current operation, though some information on the former is still missing.

---

- If you are changing Rust code or integration tests (`Cargo.*`, `crates/`, `etc/`, `integration-tests/`, `src/`):
  - [x] Did you run the integration tests with `act` through the `act-wrapper` (as described in `TESTING.md`)?

- If you are adding/deleting man pages:
  - [ ] Did you update the `man_pages` config in `doc/manual/source/conf.py`?
  - [ ] Did you update the packaged man pages in the `Cargo.toml`?
  - [ ] Did you commit the freshly built man pages?

- If you are modifying man pages:
  - [ ] Did you commit the updated built man pages?

PR #589, opened 2026-04-15
Implement control flow for restoring zones

bal-e:
This PR implements the zone data state machine logic for #550, and re-organizes Cascade's support for persistence (which now includes restoration) into a new `persistence` module. It does not break Cascade; we can merge it and then make a PR that implements the actual persistence and restoration operations (based on #550).

---


bal-e:
The `ixfr-in` test is broken due to #567, but all other tests pass, and I believe the PR is ready to go.

src/manager.rs
```diff
@@ -88,22 +90,18 @@ impl Manager {
             // Unwrap, because there should always be a valid IPv4/IPv6
             // address. Otherwise this socket couldn't have been created.
             let addr = socket.local_addr().unwrap();
-            info!(
-                "Obtained TCP listener for HTTP server for remote-control and metrics on address {addr}"
-            );
+            info!("Obtained a TCP listener for the remote-control server on address {addr}");
```
ximon18: I believe " and metrics" was specifically added based on review feedback on an earlier PR. What is the rationale for removing it here?

src/center.rs
```diff
@@ -97,14 +105,23 @@ pub async fn add_zone(
...
+            //
+            // This will clear the data for the zone and register it against the
+            // zone servers.
+            ZoneHandle {
+                zone: &zone,
+                state: &mut zone_state,
+                center,
+            }
+            .storage()
+            .abandon_loaded_restoration(restorer);
```
ximon18, line 124: I find abandoning something here that isn't clear that it is in progress (unless you deep dive into the zone data storage state machine) confusing.

Would it not be simpler and clearer to let the zone go through restoration and not actually restore anything because there is nothing to restore?

Or since this is the only place that `Zone::new()` is called, make `Zone::new()` initialize the state machine into the passive state or have it do this transition past restoration?

src/persistence/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+                    state: &mut state,
+                    center: &center,
+                };
+
+                let restorer = handle
+                    .state
+                    .storage
+                    .restorer
+                    .take()
+                    .expect("A restored zone leaves its 'restorer' field alone");
```
ximon18: At this point the zone is not yet restored, so is this expect text correct?

crates/zonedata/src/restorer.rs
```diff
@@ -0,0 +1,532 @@
...
+        // - 'diff' is empty, as checked by 'built()' above.
+        Some(LoadedZoneReplacer::new(curr, next, &mut self.diff))
+    }
+
+    /// Patch the instance being built.
+    ///
+    /// A [`LoadedZonePatcher`] is returned, which can be used to restore the
+    /// instance from a series of diffs.
+    ///
+    /// If an initial has not yet been built, [`None`] is returned.
```
ximon18:
```suggestion
    /// If an instance has not yet been built, [`None`] is returned.
```
?

crates/zonedata/src/restorer.rs
```diff
@@ -0,0 +1,532 @@
...
+
+        // Set the diff.
+        self.diff = Some(Box::new(DiffData::new()));
+    }
+
+    /// The new (loaded) instance.
+    ///
+    /// If a new instance instance has been built (with [`Self::fill()`] or
+    /// [`Self::patch()`]), it can be accessed here. Note that empty instances
+    /// (as built by [`Self::clear()`]) cannot be accessed.
```
ximon18:
```suggestion
    /// If a new instance has been built (with [`Self::fill()`] or 
    /// it can be accessed here. Note that empty instances
    /// (as built by [`Self::clear()`]) cannot be accessed.
```

src/persistence/zone.rs
```diff
@@ -0,0 +1,205 @@
...
+        self.state
+            .persistence
+            .ongoing
+            .spawn_blocking(span, move || {
+                debug!("Attempting to restore persisted zone data");
+
+                // Try to restore the loaded instance.
+                let mut restorer = restorer;
+                let restored = match super::restore_loaded(&zone, &center, &mut restorer) {
+                    Ok(()) => restorer.finish().unwrap_or_else(|_| {
```
ximon18, line 60: If this is not supposed to fail, why return an error from `finish()`?

Or if it is supposed to fail and the caller should retry (as indicated by the comment on `finish()`) why doesn't it retry here but instead panic?

src/zone/storage.rs
```diff
@@ -532,6 +550,117 @@ impl StorageZoneHandle<'_> {
...
+                transition.move_to(ZoneDataStorage::Passive(new_s));
+            }
+
+            _ => unreachable!(
+                "A 'SignedZoneRestored' is only available in the 'RestoringSigned' state"
+            ),
+        }
+
+        // Register the zone against the zone servers.
+        LoadedReviewServer::add_zone(self.center, self.zone.clone(), loaded_reviewer);
```
ximon18, line 598: Shouldn't the loaded reviewer server be updated in `finish_loaded_restoration()` instead of here?

src/state/mod.rs
```diff
@@ -53,9 +58,26 @@ pub enum Spec {
...
+    /// file, that need to be parsed and inserted in the state.
+    pub fn parse(
+        self,
+        zones: &mut foldhash::HashSet<Name<Bytes>>,
+        policies: &mut foldhash::HashMap<Box<str>, PolicySpec>,
+    ) -> State {
         match self {
-            Self::V1(spec) => spec.parse(),
+            Self::V1(mut spec) => {
+                // Extract and write out 'zones' and 'policies'.
```
ximon18, line 72: I see that extraction of zones is done because the type being parsed is only a zone name and the resulting state needs to contain zones, which is what the calls to `Zone::restore()` in `main()` create using the in/out zones (set of zone names) parameter is used for. This would be clearer to me if the in/out parameter `zones` passed to `init_from_file()` was called `zone_names` instead.

For policies it's a bit less clear to me what is going on here.

It seems that main state and zone state both store full policy objects/descriptions. I wonder is it okay to have the policy objects for the same policy stored twice, once in main state and then zero or more times in zone state, where the various copies may (shouldn't?!?) differ?

Actually I see that elsewhere you indeed hit this problem as I see code that warns:

> "Detected an inconsistency between the details of policy '{}' last used with the zone and the policy details in the global state; switching to the global state's details",

And:

> "The policy '{}' used by the zone is not present in the global state; restoring it into the global state"

I suspect that you are not parsing the `PolicySpec` into `Policy` directly in the call to `init_from_state()` because you want to track which policies are used by which zones which you do in `zone::state::mod::Spec::parse()` using an in/out policies parameter, which is the reason you pass policies out via an in/out parameter to `init_from_state()`.

However, couldn't you just let the policies by restored by `state::v1::Spec::parse()` and then in `main()` after calling `Zone::restore()` for each zone then update the set of zones referenced by each policy which would be simpler to read and reason about IMO. And then you also wouldn't need the separate calls to parse the policy in the zone parsing, and you wouldn't need the separate call to `state.policies.extend()` in `main()`.

Or am I missing something?

src/state/v1.rs
```diff
@@ -40,31 +37,19 @@ pub struct Spec {
...
-            policies,
-            rt_config: cascade_cfg::RuntimeConfig::default(),
-            tsig_store: TsigStore::default(),
-            enqueued_save: None,
-        }
+        let Self {
+            // The caller will extract 'zones' and 'policies' beforehand.
+            zones: _,
+            policies: _,
+            // TODO: More fields.
```
ximon18, line 49: Which "more fields"?

src/state/v1.rs
```diff
@@ -40,31 +37,19 @@ pub struct Spec {
...
+        };
+
+        // TODO: Initialize fields from 'Spec'.
```
ximon18, line 52: Which "fields from 'Spec'"?

src/zone/state/mod.rs
```diff
@@ -93,14 +91,14 @@ impl Spec {
...
                 };
 
-                // This should always be some at this stage...
-                let policy = policy.map(|policy| sync_policy(policy.parse(), zone, policies));
-                if let Some(policy) = &policy {
-                    let p = policies
-                        .get_mut(&*policy.name)
-                        .expect("zone policy references should not be kept around");
-                    p.zones.insert(zone.name.clone());
+                // TODO: Won't this always be a `Some`?
```
ximon18, line 100: Since zones cannot be added without policies and policies in use by zones cannot be removed one would indeed assume so!

src/zone/machine.rs
```diff
@@ -341,6 +376,23 @@ impl<'a> ZoneHandle<'a> {
...
+    /// Switch to an approved and persisted instance of the zone.
+    pub(crate) fn switch(&mut self, persisted: cascade_zonedata::SignedZonePersisted) {
+        // Move to the 'Waiting' state.
+        let (transition, state) = self.state.machine.transition();
+        let ZoneStateMachine::SignedReview(signed) = state else {
+            unreachable!(
+                "A 'SignedZonePersisted' can only exist when the zone is in signer review"
+            );
+        };
+        transition.move_to(ZoneStateMachine::Waiting(signed.approve()));
```
ximon18, line 398: This changes the zone state machine to waiting before the storage state machine has updated the publication zone server viewer, is this okay?

src/zone/storage.rs
```diff
@@ -532,6 +550,117 @@ impl StorageZoneHandle<'_> {
     }
 }
 
+/// # Persistence Tasks
+impl StorageZoneHandle<'_> {
+    /// Successfully finish loaded-instance restoration.
+    ///
+    /// A [`SignedZoneRestorer`] is returned so the signed instance can be
+    /// restored too.
```
ximon18:
```suggestion
    /// restored as well.
```
I was confused by "too" as I read it as a typo of "to" which was trying to say "in to". Your text is correct, but it would be easier to understand at a glance for me by saying "as well" instead.

ximon18 (approved):
Took a while to understand but looks good to me. The only slight thought I have is that there are multiple things called persist and restore and only one of them does actual persist and restore, e.g. a Persister doesn't persist anything nor does a Restorer restore anything, instead they give the required read/write access (respectively) to the underlying zone data storage that needs to be written to/read from to do persistence/restoration. So maybe the names could be made a bit more distinct.

src/manager.rs
bal-e: I'll check the git-blame, but it seemed unnecessarily verbose to me. The configuration option for these addresses is just `remote-control.servers`, and I think operators will understand that metrics are also served here.

src/center.rs
bal-e: You're right that "abandon" here is hard to understand. I had initially attempted to add a constructor for `ZoneDataStorage` that directly provides the zone viewer objects; but that complicates the zone initialization procedure (either those viewer objects are _also_ stored in `StorageState`, or they are directly passed out of `Zone::new()`). The current approach seemed more compact, but I didn't realize how strange it looks.

> Would it not be simpler and clearer to let the zone go through restoration and not actually restore anything because there is nothing to restore?

Hmm... that's not a bad idea. I'm just worried that the resulting logs will confuse operators ("why is Cascade trying to restore a newly created zone from disk?").

OTOH, the end result of this "just try restoring" approach would be the same -- `abandon_loaded_restoration()` would be called.

> Or since this is the only place that `Zone::new()` is called, make `Zone::new()` initialize the state machine into the passive state or have it do this transition past restoration?

See above; I was concerned about the (cognitive and syntactic) overhead of juggling the three viewer objects through `Zone::new()` and its internals.

I agree that the current approach is not ideal, but the alternatives also have drawbacks. I'm open to changing this if we can ascertain that one option is definitely better.

src/zone/storage.rs
bal-e: That is an alternative option, yes. When I had set up the transitions in the data storage state machine, it felt like returning all the viewer objects once restoration is fully complete was the simplest option. I don't think there's too much of a difference, but I can change it if you'd prefer.

src/zone/state/mod.rs
bal-e: The policy field was made optional back when we looked into multi-step zone creation. IIRC we never discounted that possibility, and we've never made a move to undo the policy field being optional. So this comment really concerns the delayed zone creation case.

src/state/v1.rs
bal-e: Right, I should have elaborated here. I have organized this code in a future-proof way; should we ever store more data in the global state (e.g. information about TSIG keys that does not belong in the TSIG store file), and add it to `Spec`, this pattern-match will fail so we remember to update it. The "more fields" here refers to those future field additions. Without that future consideration, this function does not have a purpose and can be removed; but IMO it's better to keep it here.

bal-e: I could remove this specific comment and make a more general one at the top of the function body.

src/state/v1.rs
bal-e: The same "more fields"; see the prior discussion.

bal-e: Again, I could remove this comment in favor of one at the top of the function body.

src/state/mod.rs
bal-e: I definitely need to better document the design I was aiming for here. The goal is to distinguish `PolicySpec` from `Policy`; `Spec` is a (version-independent) representation of a policy loaded from state, so e.g. it should contain `String`s to represent TSIG keys by name. `Policy` (or more specifically `PolicyVersion`), on the other hand, is the type that is used throughout Cascade, so e.g. it should store `Arc<TsigKey>` to refer to TSIG keys. This separation gives us an appropriate point for resolving references between different bits of data. This function returns `PolicySpec` so that the caller (i.e. `main()`) can decide when to resolve those references.

> It seems that main state and zone state both store full policy objects/descriptions. I wonder is it okay to have the policy objects for the same policy stored twice, once in main state and then zero or more times in zone state, where the various copies may (shouldn't?!?) differ?

This was intentional. If the user changes the policy files while Cascade is shut down, it needs a way to notice those differences and update its internal state (e.g. update dnst-keyset, maybe change how it signs stuff, etc.). Again, something I should document. Each zone also retains a copy of its policy in case a user copies around zone state files and accidentally changes policy details in between. Arguably, the copies of the policies in the global state file can be elided, but that's something to think about and resolve later.

Does that help explain the design? I will add some documentation here, but let me know if you think other changes are needed.

src/persistence/zone.rs
bal-e: I chose to return a result instead of panicking because `.finish()` is defined inside `cascade_zonedata`, i.e. across a crate boundary, and this felt more appropriate. The same is true in a few other places, e.g. in the `ZoneBuilder` types. I'm happy to change them all into panics.

src/persistence/mod.rs
bal-e: The message here was ambiguous, sorry; by "restored" here I meant "restored from a zone state file", not "data restored". The new `Zone::restored` field uses the same meaning (and mentions it in its RustDoc). Do you have any suggestions for how to disambiguate the two cases (ideally without making the names more verbose)?

src/zone/machine.rs
bal-e: Well spotted! I agree that it would be better to move to Waiting once the publication server is ready. I'll update the code.

src/manager.rs
ximon18: Here is the relevant discussion: https://github.com/NLnetLabs/cascade/pull/489#discussion_r2872142536

src/center.rs
ximon18: My preference would of course be own suggestion:

> let the zone go through restoration and not actually restore anything because there is nothing to restore

You also don't need to log anything if there's nothing to restore so that would resolve your concern:

> Hmm... that's not a bad idea. I'm just worried that the resulting logs will confuse operators ("why is Cascade trying to restore a newly created zone from disk?").

src/persistence/zone.rs
ximon18: I think if the caller can ever legally get into the situation where the result would be an error then returning a result is fine. But then I still wonder why you advise retrying but don't retry?

If however it should never happen and the caller will just panic, then while I see your point about crate boundaries I think that applies to public library crates intended for consumption by others and not to internal crates which are really just compilation optimization / code co-location mechanism, then I think this might as well be an assert in the zonedata crate.

src/state/v1.rs
ximon18: I'm inclined to simplify the code by removing the function and, maybe, leave a comment at the top to say this could be extended in future (though I wonder if that really adds any value).

src/zone/state/mod.rs
ximon18: I think we should get rid of the `Option` nature of the policy field on zones. We can add it in again later if ever needed and avoid the confusion and complexity until then.

src/persistence/mod.rs
ximon18:
```suggestion
                    .expect("zone is neither newly added nor already restored");
```

Maybe it would also be good to have a comment somewhere that we expect only a single instance of Restorer to ever be created and for its `run()` function to only ever be invoked a single time?

src/zone/storage.rs
ximon18: If a zone has persisted loaded data but not persisted signed data, would it even call this `finish_signed_restoration()` function and thus is it possible the loaded review server should get updated but would not?

src/state/mod.rs
ximon18: > This was intentional. If the user changes the policy files while Cascade is shut down, it needs a way to notice those differences and update its internal state (e.g. update dnst-keyset, maybe change how it signs stuff, etc.)

Okay, but why does that require zone files to have policy embedded in them?

> Arguably, the copies of the policies in the global state file can be elided, but that's something to think about and resolve later.

I'm not sure what you mean by this.

> Does that help explain the design? I will add some documentation here, but let me know if you think other changes are needed.

Yes, though it doesn't make the code any less confusing nor do I have the idea that it addresses my suggestion to _"let the policies by restored by state::v1::Spec::parse() and then in main() after calling Zone::restore() for each zone then update the set of zones referenced by each policy"_.

src/zone/state/mod.rs
bal-e: Not a problem for this PR.

src/zone/storage.rs
bal-e: No; cascade-zonedata always returns all three viewer objects when restoration is completed (whether loaded+signed restoration fails, only loaded restoration succeeds, or loaded+signed restoration succeeds). rustc would have complained if we forgot to use one of those objects.

src/persistence/mod.rs
bal-e: > Maybe it would also be good to have a comment somewhere that we expect only a single instance of Restorer to ever be created and for its run() function to only ever be invoked a single time?

The same is true of all the other units (with the minor exception of the zone server). I don't think it's necessary to mention it here.

bal-e:
As per internal discussion, we are merging this and leaving leftover issues (which are all quite minor) to be resolved later.

PR #590, opened 2026-04-16
FIX: Restore zone TSIG key state on startup.

ximon18:
On startup ensure that the TSIG key field on the server source of a zone is set to `Some(key)` rather than `None`. This required changing the sequence in which state is loaded by `main()` because TSIG keys must be loaded _before_ zones are reconstituted from their state files so that the loaded key can be placed in the `Some(key)` field of the zone server source.

---


src/zone/state/v1.rs
```diff
@@ -497,31 +500,49 @@ pub enum ZoneLoadSourceSpec {
...
             Self::Zonefile { path } => Source::Zonefile { path },
-            // TODO: Look up the TSIG key in the key store.
-            Self::Server { addr, tsig_key: _ } => Source::Server {
+            Self::Server {
+                addr,
+                tsig_key: Some(tsig_key),
+            } => {
+                let key_name = KeyName::from_str(&tsig_key.to_string()).map_err(|_| {
+                    ZoneLoadSourceSpecParseError::InvalidTsigKeyName(tsig_key.clone())
+                })?;
```
bal-e, line 513: This error is not possible. `KeyName` is just `Name<Array<255>>`, i.e. an arbitrary domain name. I would suggest using `tsig_key.flatten_into().expect("'KeyName' can hold any domain name")` and removing the `InvalidTsigKeyName` error.

src/main.rs
bal-e, line 1: Note: I also ended up making similar changes to this file in #589, so we will have merge conflicts to deal with.

bal-e (commented):
We can merge if necessary, but I'd like to have my comments addressed first.

ximon18:
Replaced by #626.

PR #597, opened 2026-04-20
Switch to exit_signalled from daemonbase to also handle SIGTERM.

Philip-NLnetLabs:
Provide your description here

---


Cargo.toml
```diff
@@ -320,3 +320,6 @@ cascade-dnst = "*"
 
 [package.metadata.generate-rpm.recommends]
 cascade-hsm-bridge = "*"
+
+[patch.crates-io]
+daemonbase = { git = "https://github.com/NLnetLabs/daemonbase", branch = "exit-signalled" }
```
bal-e: I thought `[patch]` was only meant for local testing, but apparently this approach works. But why not change the `[dependencies]` section?

Philip-NLnetLabs: The idea is to get the corresponding PR merged in daemonbase. Then I'll change this PR to use the git URL for daemonbase.

bal-e: I see, cool.

bal-e (approved):
LGTM! But it looks like you'll have to update `Cargo.lock` for CI to pass.

PR #600, opened 2026-04-22
Add last error to zone status

tertsdiepraam:
Show an error that happened during the last operation in zone status. I've included the most important errors based on when `error!` was called, but we might have to add more errors to the zone history. We might also want to collect all of those into a single enum variant in the history, but we can always do that later.

```
zone:   se
policy: foo
source: /home/terts/code/cascade/se.zone.txt

review hooks
  loaded: none
  signed: none

last published
  <no versions published yet>

status: idle

An error occurred during the last operation:
  ERROR: the zonefile could not be loaded: the zonefile contains multiple SOA records
  Run `cascade zone history se` for more information.
```

---

### DIFF
--- a/src/units/http_server.rs
+++ b/src/units/http_server.rs
@@ -353,6 +353,7 @@ impl HttpServer {
         let signed_serial;
         let published_serial;
         let last_published;
+        let error;
         {
             let locked_state = state.center.state.lock().unwrap();
             let keys_dir = &state.center.config.keys_dir;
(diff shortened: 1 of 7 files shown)


src/units/http_server.rs
```diff
@@ -465,6 +466,47 @@ impl HttpServer {
...
+                        break;
+                    }
+                    HistoricalEvent::SigningFailed { trigger: _, reason } => {
+                        found_error = Some(format!("signing failed: {reason}"));
+                        break;
+                    }
+                    HistoricalEvent::UnsignedZoneReview {
+                        status: ZoneReviewStatus::Rejected,
+                    } => {
+                        found_error = Some("loaded zone was rejected".into());
```
ximon18, line 489: Is it clear what the cause of rejection was here? To me it could be due to a bad zone file based on this text. Would it be better to say "loaded zone failed review" or something?

ximon18 (approved):

src/units/http_server.rs
tertsdiepraam: I wanted to write "rejected by review hook" at first, but we don't currently expose how the rejection happened. I think "loaded zone failed review" reads too close to "loaded zone review failed" for me, implying that the script failed.

PR #602, opened 2026-04-22
Report progress of zonefile loading

tertsdiepraam:
Work in progress. Requires https://github.com/NLnetLabs/domain/pull/642 to be merged first.

---

### DIFF
--- a/src/loader/zonefile.rs
+++ b/src/loader/zonefile.rs
@@ -41,6 +41,10 @@ pub fn load(
     // Parse all the records, extracting the SOA. We always read the whole zone.
     while let Some(record) = parse_record(&mut buf, zone, &mut reader)? {
         metrics.num_loaded_records.fetch_add(1, Relaxed);
+        metrics
+            .num_loaded_bytes
+            .store(reader.current_offset(), Relaxed);
+
         match record {
             Parsed::Soa(soa) => {
                 writer.set_soa(soa)?;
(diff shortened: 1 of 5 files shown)


src/loader/zonefile.rs
```diff
@@ -68,9 +72,7 @@ fn make_reader(
 
     let file_len = file.metadata().map_err(Error::Open)?.len();
 
-    metrics
-        .num_loaded_bytes
-        .fetch_add(file_len as usize, Relaxed);
+    metrics.num_total_bytes.store(file_len as usize, Relaxed);
```
ximon18, line 75: Why rename the metric field from `loaded` to `total`?

ximon18 (approved):

src/loader/zonefile.rs
tertsdiepraam: Because loaded sounds like "loaded so far" to me, while "total" means how many bytes we will load or have loaded. I'm happy to change the names though.

ximon18: The problem I have with total is it says nothing about loading, it could be total anything.

tertsdiepraam: I kind of agree, but it is in a type called `ActiveLoadMetrics` so that does imply loading

bal-e: Yeah, I think the name is fine. The context (between the file path, function name, and name of the containing struct) is obvious enough.

tertsdiepraam:
Force pushed to remove a commit that updated Cargo.lock. Nothing else changed.

Issue #300, opened 2025-11-03
Support for outgoing incremental zone transfers


bal-e:
Cascade does not currently support IXFR queries in its DNS servers (it does support consuming IXFRs from upstream servers).  Outgoing IXFR support would reduce network bandwidth and improve efficiency -- it may even be useful for reviewing zones.  This could be implemented in two ways:

- Maintain the existing infrastructure within Cascade and internally diff consecutive published zones to generate IXFRs.  This would improve network bandwidth but slow down zone publishing.

- Wait for Cascade to implement incremental signing (#299), which would improve the internal infrastructure and make fully incremental operation (including efficiently computing IXFR output) possible.

This also relates to #295 -- zones with multiple versions could be persisted using IXFR outputs, which would save disk space and improve efficiency.

ximon18:
This functionality was included in the 0.1.0-beta1 release.

PR #605, opened 2026-04-28
Add IXFR out support.

ximon18:
Implements IXFR out based using internally created signed diffs which are moved to zone storage after they have been made available to zone persistence. Code was based on XFR code in `domain`.

System tests have been moved out on request to #650.

Lacks:
- [x] Use of TSIG in the system test.
- [ ] Purging of old / excess diffs.
- [ ] RFC 1995 diff condensation.
- [x] Automated testing of IXFR out based on changes caused by incremental signing. Note: seen working manually.
- [ ] Automated testing of IXFR out to a downstream BIND instance. Note: seen working manually.
- [x] Automated testing of IXFR out from the loaded and signed review nameserver ports of Cascade. Note: seen working manually,

Relates to #300.

---


integration-tests/tests/ixfr-out/action.yml
```diff
@@ -0,0 +1,183 @@
...
+        cascade template policy | grep -Ev '(send-notify-to|accept-xfr-from)' > "${POLICY_DIR}/custom.toml"
+
+        sed -i -e 's/serial-policy = "date-counter"/serial-policy = "keep"/' "${POLICY_DIR}/custom.toml"
+        echo 'send-notify-to = ["127.0.0.1:1054"]' >> "${POLICY_DIR}/custom.toml"
+
+        # TODO: Extend the test to use TSIG (driven by inputs ala the tsig-downstream test)
+        # once PRs #564 and #587 have been merged.
+        # echo 'send-notify-to = ["127.0.0.1:1054^tsig-key"]' >> "${POLICY_DIR}/custom.toml"
+        # cascade tsig add tsig-key hmac-sha256 "COzoVsYQmXeXiyq1Quhp0bbVnMyxjPxsaGSoIWR98i0="
+        cascade policy reload
```
bal-e: Now that the TSIG PRs are merged, is this unblocked?

src/server/service.rs
```diff
@@ -149,10 +153,9 @@ mod compat {
                         }
 
                         // TODO: Support IXFR.
-                        ZoneRequestKind::Ixfr { .. } => Box::pin(std::future::ready(error(
-                            old_request.message(),
-                            Rcode::NOTIMP,
-                        ))),
+                        ZoneRequestKind::Ixfr { known_soa } => {
+                            Box::pin(ixfr(old_request, known_soa.rdata, zone.clone())) as Response
+                        }
```
bal-e: The TODO comment can be removed now, I guess?

src/server/service.rs
```diff
@@ -294,6 +297,151 @@ mod compat {
...
     }
 
+    async fn ixfr<V: Viewer + Send + Sync + 'static>(
+        request: Request<Vec<u8>, Option<Arc<tsig::Key>>>,
+        client_soa: Soa<Box<Name>>,
+        zone: ServedZone<V>,
+    ) -> ResponseStream {
+        // Refuse IXFR requests over UDP.
+        if request.transport_ctx().is_udp() {
+            tracing::warn!("Reject IXFR over UDP");
```
bal-e: Maybe "Reject" => "Rejecting" to make the event a bit more clear?

src/server/service.rs
```diff
@@ -294,6 +297,151 @@ mod compat {
...
+            return error(request.message(), Rcode::NOTIMP);
```
bal-e: An alternative implementation could return a response message with the latest SOA record we have for the zone. From what I remember of the IXFR RFC, this would inform clients if they already have an up-to-date copy of the zone, and otherwise tell them to try again over TCP (purportedly because the server thinks the IXFR message is too big to fit over UDP). It's basically a nicer way of telling clients to check TCP, and reduces the chance of confusing them (by making them think we don't implement IXFR at all).

In any case, this is not a blocker for merging.

src/server/service.rs
```diff
@@ -294,6 +297,151 @@ mod compat {
...
+        //    of the server is received, it is replied to with a single SOA
+        //    record of the server's current version, just as in AXFR."
+        //                                            ^^^^^^^^^^^^^^^
+        // Errata https://www.rfc-editor.org/errata/eid3196 points out that
+        // this is NOT "just as in AXFR" as AXFR does not do that.
+        let our_soa_serial = { viewer.soa().rdata.serial };
+
+        if client_soa.serial >= our_soa_serial {
+            trace!("Responding to IXFR with single SOA because query serial >= zone serial");
+            return soa(request.message(), &*viewer);
```
bal-e, line 425: Could you document on `soa()` that `ixfr()` calls into it? Otherwise its implementation could change in the future to assume that it received a SOA query.

src/server/service.rs
```diff
@@ -294,6 +297,151 @@ mod compat {
...
+
+        // https://datatracker.ietf.org/doc/html/rfc1995#section-2
+        // 2. Brief Description of the Protocol
+        //   "If an IXFR query with the same or newer version number than that
+        //    of the server is received, it is replied to with a single SOA
+        //    record of the server's current version, just as in AXFR."
+        //                                            ^^^^^^^^^^^^^^^
+        // Errata https://www.rfc-editor.org/errata/eid3196 points out that
+        // this is NOT "just as in AXFR" as AXFR does not do that.
+        let our_soa_serial = { viewer.soa().rdata.serial };
```
bal-e: Are the `{ }` necessary?

src/server/service.rs
```diff
@@ -294,6 +297,151 @@ mod compat {
...
+        //     AXFR response except the query type is IXFR."
+
+        // https://datatracker.ietf.org/doc/html/rfc1995#section-2
+        // 2. Brief Description of the Protocol
+        //   "If an IXFR query with the same or newer version number than that
+        //    of the server is received, it is replied to with a single SOA
+        //    record of the server's current version, just as in AXFR."
+        //                                            ^^^^^^^^^^^^^^^
+        // Errata https://www.rfc-editor.org/errata/eid3196 points out that
+        // this is NOT "just as in AXFR" as AXFR does not do that.
```
bal-e: Is the errata relevant to this function? We could also omit the ", just as in AXFR" portion in the quoted text.

src/server/service.rs
```diff
@@ -294,6 +297,151 @@ mod compat {
...
+            trace!("Responding to IXFR with single SOA because query serial >= zone serial");
+            return soa(request.message(), &*viewer);
+        }
+
+        let diffs = zone.handle.state.lock().unwrap().storage.diffs.clone();
+
+        // TODO: Add something like the Bind `max-ixfr-ratio` option that
+        // "sets the size threshold (expressed as a percentage of the size of
+        // the full zone) beyond which named chooses to use an AXFR response
+        // rather than IXFR when answering zone transfer requests"?
```
bal-e, line 455: Nice! That's a good idea!

src/server/service.rs
bal-e: Also, interestingly, this errata does not appear on https://www.rfc-editor.org/rfc/inline-errata/rfc1995.html.

src/server/service.rs
```diff
@@ -294,6 +297,151 @@ mod compat {
...
+        // anything about whether an IXFR response can consist of more than
+        // one response message, but given the 2^16 byte maximum response
+        // size of a TCP DNS message and the 2^16 maximum number of ANSWER
+        // RRs allowed per DNS response, large zones may not fit in a single
+        // response message and will have to be split into multiple response
+        // messages.
+
+        // Currently we have only a single diff available.
+        for (i, d) in diffs.iter().enumerate() {
+            tracing::info!(
```
bal-e: I don't think INFO is the right log level for this; TRACE (or at most DEBUG) feels much more appropriate.

src/server/service.rs
```diff
@@ -294,6 +297,151 @@ mod compat {
...
+        });
+
+        let Some(start_idx) = start_idx else {
+            tracing::trace!(
+                "Falling back from IXFR to AXFR because no diff is available for zone '{}' from serial {}",
+                zone.handle.name,
+                client_soa.serial,
+            );
+            return axfr(request, zone_clone).await;
+        };
```
bal-e, line 525: Nice! Like with `soa()`, it might be good to add a comment on `axfr()` that it can be called from `ixfr()`.

src/server/service.rs
```diff
@@ -294,6 +297,151 @@ mod compat {
...
+
+        let (tx, mut rx) = tokio::sync::mpsc::channel(1024);
+
+        // Stream the records in the background.
```
bal-e, line 542: Could you duplicate and/or reference the long comment in the AXFR code regarding the task spawning?

src/server/service.rs
```diff
@@ -294,6 +297,151 @@ mod compat {
...
+        tokio::task::spawn(async move {
+            // Collect the sequence of IXFR output records..
```
bal-e: Typo: ".." => "."

src/server/service.rs
```diff
@@ -294,6 +297,151 @@ mod compat {
...
+            let mut rrs = vec![new_soa.clone().into()];
+            for diff in &diffs[start_idx..] {
+                rrs.push(diff.removed_soa.clone().unwrap().into());
+                rrs.extend(diff.removed_records.clone());
+                rrs.push(diff.added_soa.clone().unwrap().into());
+                rrs.extend(diff.added_records.clone());
+            }
+            rrs.push(new_soa.into());
```
bal-e, line 636: You can use `diffs[start_index..].iter().flat_map()` to avoid the intermediate `Vec` allocation.

Not a blocker for merging.

bal-e (approved):
This looks great! I'm glad it didn't end up being a big PR, it looks like most of the surrounding architecture and control flow suited this addition well. I only have some minor comments; none of them are a blocker for merging. A few them only ask to add some code comments; that's probably the most important stuff in here, because it's easy to forget once the PR is merged. But again, not a blocker.

src/server/service.rs
ximon18: Oops, that loop wasn't meant to stay. If it does, it could perhaps also be guarded by an if debug enabled check or such.

src/server/service.rs
ximon18: No. Must be a left over of something, no idea otherwise why they would be there.

src/server/service.rs
ximon18: I updated the code, please re-check.

integration-tests/tests/ixfr-out/action.yml
ximon18: I have updated the test to run both with and without TSIG. Manually causing it to fail and inspecting the NSD secondary logs reveals that the TSIG key is used, e.g.:

```
| [2026-05-07 22:37:56.849] nsd[129]: info: xfrd: zone notify-and-xfr-tsig.test committed "received update to serial 1 at 2026-05-07T22:37:56 from 127.0.0.1@4542 TSIG verified with key tsig-key"                         
```

Ideally the test would prove this but I don't want to grep NSD logs to do that.

ximon18: I updated the test to have `dig` attempt to XFR the zone without a TSIG key when it was configured to require one, which fails with RCODE REFUSED as expected, thus proving that a request that works must have been correctly TSIG signed.

src/server/service.rs
ximon18: The part of RFC 1995 that you are referring to is:

> If the UDP reply does not fit, the query is responded to with a single SOA record of the server's current version to inform the client that a TCP query should be initiated.

If we were to return a single SOA it would not technically be the same as refusing [UDP, it implies that we do support UDP but as you say the response would not fit, which could be confusing if one expects the diff to be very small, so I'm not sure this is the right way to refuse to support UDP at all for XFR.

However, the RFC also says:

> Transport of a query may be by either UDP or TCP.

So to be strictly compliant it seems we MUST support UDP, and the graceful way to get what we want is to do as you suggest, return a single SOA.

src/server/service.rs
ximon18: Done.

src/server/service.rs
ximon18: As I couldn't immediately see exactly how to do that, for now I just added a NOTE to maybe do it in future.

src/server/service.rs
ximon18: It doesn't appear there because it is not a verified errata but instead is held for update: https://www.rfc-editor.org/errata/eid3196. I have removed the comment as indeed it doesn't add much, it was just copy pasted from code in `domain`.

src/server/service.rs
```diff
@@ -98,7 +102,8 @@ mod compat {
 
         fn call(&self, old_request: Request<Vec<u8>, Option<Arc<tsig::Key>>>) -> Response {
             // Parse the request.
-            let message = old_request.message().as_slice();
+            let slice = old_request.message().as_slice();
+            let message = slice;
```
Philip-NLnetLabs: Maybe remove the slice

PR #606, opened 2026-04-29
Introduce better `ZoneState` locks

### LINKED ISSUE #599: Add a wrapper around `Mutex<ZoneState>`
Locking the `ZoneState` mutex can fail as `std::sync::Mutex` is _poisoning_; if a panic occurs while a lock is held, the data is marked as poisoned. The poison status can be removed manually, but we don't even want to support poisoning -- Cascade aborts on panic, so a poison status cannot be observed. Handling the possibility of failure (using `.unwrap()`, `.expect()`, or propagating the error) adds boilerplate across Cascade, and different call sites handle it differently.

We can simplify things by building a simple wrapper around `Mutex` that hides the failure possibility, automatically adding a `.expect()`. This would reduce boilerplate and make things more consistent. This wrapper may have additional benefits -- for example, it can mark zones as dirty automatically, as many call sites in Cascade forget to do so.

bal-e:
This PR introduces `ZoneStateLock`, a wrapper around the `ZoneState` lock with higher-level methods. `ZoneStateLock` alleviates boilerplate and inconsistencies around how we use the `ZoneState` lock.

Previously, access to zone state was obtained via `zone.state.lock().unwrap()`. Some call sites used `.expect()` or `?` instead of `.unwrap()`, introducing inconsistencies. Most importantly: _most_ call sites did not call `.mark_dirty()`, so changes to the zone were not being persisted with regularity.

`ZoneStateLock` wraps an `RwLock` instead of a `Mutex` because some call sites only read the zone state, and it seemed better to expose that functionality rather than use write locks everywhere. Rather than a simple `.lock()` method, a write lock must be obtained with the obtuse name `.write_cleanly()`, and that method tells users to prefer `Zone::write()`. `Zone::write()` marks the state as dirty automatically so that callers don't have to remember to.

In addition, a new `OwnedZoneHandle` type is introduced. This works around some borrowing limitations of `ZoneHandle`. The new `Zone::write_handle()` method obtains a write lock over the zone state, marks it as dirty, and returns an `OwnedZoneHandle` so that handle-related methods can be called conveniently.

All call sites of `zone.state.lock()` have been updated to use the new locking methods, and most of them (particularly those that needed `ZoneHandle`) are much simpler now.

Note: this PR is very prone to (causing and incurring) merge conflicts.

Resolves #599.

---

### DIFF
--- a/src/signer/incremental.rs
+++ b/src/signer/incremental.rs
@@ -80,14 +80,7 @@ pub fn sign_incrementally(
     let keyset_state: KeySetState = serde_json::from_str(&state)
         .map_err(|e| SignerError::SigningError(format!("loading keyset state failed: {e}")))?;
 
-    let policy;
-    {
-        let zone_state = zone
-            .state
-            .lock()
-            .map_err(|e| SignerError::SigningError(format!("zone.state.lock() failed: {e}")))?;
-        policy = zone_state.policy.clone().expect("should be there");
-    };
+    let policy = zone.read().policy.clone().unwrap();
 
     let use_nsec3 = matches!(policy.signer.denial, SignerDenialPolicy::NSec3 { .. });
 
@@ -231,7 +224,7 @@ pub fn sign_incrementally(
             .map(|s| domain::base::Serial(s.into())),
     );
 
-    ws.local_state.save(&ws.center, &ws.zone)?;
+    ws.local_state.save(&ws.center, &ws.zone);
 
     Ok(())
 }
@@ -2285,10 +2278,7 @@ pub struct LocalState {
 
 impl LocalState {
     pub fn new(zone: &Arc<Zone>) -> Result<Self, SignerError> {
-        let zone_state = zone
-            .state
-            .lock()
-            .map_err(|e| SignerError::SigningError(format!("zone.state.lock() failed: {e}")))?;
+        let zone_state = zone.read();
 
         Ok(Self {
             apex_remove: zone_state.apex_remove.clone(),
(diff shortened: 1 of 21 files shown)


src/signer/incremental.rs
```diff
@@ -76,14 +76,7 @@ pub fn sign_incrementally(
...
 
-    let policy;
-    {
-        let zone_state = zone
-            .state
-            .lock()
-            .map_err(|e| SignerError::SigningError(format!("zone.state.lock() failed: {e}")))?;
-        policy = zone_state.policy.clone().expect("should be there");
-    };
+    let policy = { zone.read().policy.clone().unwrap() };
```
tertsdiepraam: The `{}` is important here right? Maybe we should have a comment saying that it shouldn't be removed?

src/zone/mod.rs
```diff
@@ -49,10 +52,8 @@ pub struct Zone {
 
     /// The state of this zone.
     ///
-    /// This uses a mutex to ensure that all parts of the zone state are
-    /// consistent with each other, and that changes to the zone happen in a
-    /// single (sequentially consistent) order.
-    pub state: Mutex<ZoneState>,
+    /// The state is locked for consistency; see [`ZoneStateLock`].
```
tertsdiepraam: Could you link to `ZoneState` here? That would be nice for goto definition from rust-analyzer.

src/zone/mod.rs
```diff
@@ -170,6 +221,75 @@ impl ZoneHandle<'_> {
     }
 }
 
+//----------- OwnedZoneHandle --------------------------------------------------
+
+/// A [`ZoneHandle`] that owns a zone state write lock.
+///
+/// This is a convenience type. By owning the write lock, it can simplify
+/// construction of `ZoneHandle`s for quick uses.
+pub struct OwnedZoneHandle<'a> {
```
tertsdiepraam, line 236: This name is very confusing to me because it doesn't actually own the zone but only a write lock to it. Maybe `WriteZoneHandle`?

src/signer/incremental.rs
bal-e: Actually no; the lock would be dropped at the end of the statement anyway. But I can at least document that.

tertsdiepraam: Yeah I guess that would be nice. Do you just put them there for extra clarity?

bal-e: Yeah, I think it's nice to explicitly see that. And `rustfmt` doesn't try to remove it.

src/zone/mod.rs
bal-e: I think the current name works well compared to `ZoneHandle` (which holds the zone state lock by reference). `WriteZoneHandle` doesn't feel informative because `ZoneHandle` also has write access. Let's merge the PR first, while the conflicts are resolved, and debate the name later.

PR #610, opened 2026-04-30
Add maintenance mode

tertsdiepraam:
Maintenance mode disables automatic loading and signing. It is the new name of the "start/stop" functionality that we talked about.

```
cascade zone maintenance enable <name>
cascade zone maintenance disable <name>
```

Ultimately, this should allow the user to still trigger reloads and resigns manually.

We thought about many different names for this:
- start/stop
- resume/pause
- etc.

The "maintenance mode" name seemed to encapsulate the "do not start new operations" meaning best, because the others imply stopping/pausing an ongoing operation immediately. Additionally, we can expand maintenance mode to allow some useful operations.

When the pipeline is in maintenance mode, a big warning will be printed about that in the zone status.

---

### DIFF
--- a/crates/cli/src/commands/zone.rs
+++ b/crates/cli/src/commands/zone.rs
@@ -559,7 +559,7 @@ impl Zone {
                                 "Cascade will no longer automatically start new loading and signing operations"
                             );
                             println!(
-                                "Run {}`cascade zone maintenance stop {name}`{} to continue automatic operation",
+                                "Run {}`cascade zone maintenance stop {name}`{} to resume automatic operation",
                                 ansi::BLUE,
                                 ansi::RESET,
                             );
--- a/crates/cli/src/commands/zone.rs
+++ b/crates/cli/src/commands/zone.rs
@@ -130,6 +130,22 @@ pub enum ZoneCommand {
         /// The zone to report the history of.
         zone: ZoneName,
     },
+
+    /// Resume a paused zone pipeline
+    #[command(name = "maintenance")]
+    Maintenance {
+        /// Enable or disable maintenance mode.
+        #[command(subcommand)]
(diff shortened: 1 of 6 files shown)


Philip-NLnetLabs:
I don't understand the name 'manual-mode'. My mental model is that manual is the opposite of automatic. But when manual mode is enabled, Cascade just doesn't do anything. There is nothing to do manually.

tertsdiepraam:
There's nothing to do _yet_. I would like to make some operations still allowed (manual reload for example). That would actually make this mode more useful than a pure stopping mechanism.

Philip-NLnetLabs:
It is not clear to me that this manual mode would give much benefit. But if we want to go there, then the PR should be more complete. The current PR goes to manual mode and then you can't do anything.

crates/cli/src/commands/zone.rs
```diff
@@ -522,6 +538,54 @@ impl Zone {
...
+                                "Maintenance mode for zone `{name}` is now {}enabled{}",
+                                ansi::BOLD,
+                                ansi::RESET
+                            );
+                            println!("");
+                            println!(
+                                "Cascade will no longer automatically start new loading and signing operations"
+                            );
+                            println!(
+                                "Run {}`cascade zone maintenance stop {name}`{} to continue automatic operation",
```
ximon18:
```suggestion
                                "Run {}`cascade zone maintenance disable {name}`{} to resume normal operation",
```

ximon18 (approved):
LGTM. Nice work @tertsdiepraam!

ximon18:
Will you also add man page and RtD content?

tertsdiepraam:
I'll do the docs separately

PR #616, opened 2026-05-01
Soft halt

tertsdiepraam:
Re-introduces soft halting based on policy settings. To achieve this, I also reworked how the review is specified in the policy files.

> [!NOTE]
> I'm aware that the documentation is not up to date. I'd like to get consensus on this new mechanism before I update that.

It works as follows (examples are for loaded review but signed works the same way):

A missing `loader.review` in the policy file means that review is turned off. A more explicit way to say the same thing is:

```toml
[loader.review]
mode = "off"
```

Then we can switch the mode to `"manual"` for CLI review:

```toml
[loader.review]
mode = "manual"
```

or we can use a script like so:

```toml
[loader.review]
mode = "script"
hook = "my_hook"
```

Both the `"manual"` and `"script"` modes can additionally have a `on-reject` field that can be set to either `"discard"` or `"halt"` (i.e. soft and hard halt, respectively, but with more descriptive names).

`"discard"` will simply throw away the rejected zone and go back to the waiting state.

```toml
[loader.review]
mode = "manual"
on-reject = "discard"
```

`"halt"` will go to a halted state like before.

```toml
[loader.review]
mode = "manual"
on-reject = "halt"
```

The default value is `"discard"`. 

I'm currently restrictive because I don't allow `hook` to be present if `mode` is not `"script"` and I don't allow `on-reject` if `mode = "off"` but we can easily change that.

---

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -816,8 +816,21 @@ pub struct KeyManagerPolicyInfo {
 
 #[derive(Deserialize, Serialize, Debug, Clone)]
 pub struct ReviewPolicyInfo {
-    pub required: bool,
-    pub cmd_hook: Option<String>,
+    pub mode: ReviewPolicyMode,
+    pub on_reject: ReviewPolicyOnReject,
+}
+
+#[derive(Deserialize, Serialize, Debug, Clone)]
+pub enum ReviewPolicyMode {
+    Off,
+    Manual,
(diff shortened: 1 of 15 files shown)


tertsdiepraam:
I'll update the documentation page on review hooks separately from this PR.

Philip-NLnetLabs (approved):

PR #617, opened 2026-05-01
Introduce (basic) `Instances`

bal-e:
This PR adds `zone/instance.rs` as the primary location for information about the (current, upcoming, and historical) instances of zones. It replaces several bits of state that were independently tracking some of this information, and offers space to store more information in the future (e.g. DNSSEC data for the current published instance for re-signing).

This PR does not affect Cascade's external behavior, beyond some minor changes in logging. Its primary purpose is to simplify the codebase. It is a stripped-down version of #528.

The following bits of state are added:
- In `zone/instance.rs`:
  - `CurrentInstance`: the state of the instance published by Cascade.
    - Includes the SOA serials of the current loaded instance and current signed instance.
  - `UpcomingInstance`: the state of an upcoming instance being built right now.
    - Includes the SOA serials of the upcoming loaded and signed instances (if they exist).
-  In the zone state machine (`zone/machine.rs`):
  - A `decided` boolean on `LoadedReview` and `SignedReview` which tracks whether a review decision has been received.

They replace the following bits of state:
- In `StorageState` (`zone/storage.rs`): `loaded_review_soa`, `signed_review_soa`, `published_loaded_soa`, `published_soa`
  - These were being used for zone status information.
- In `zone/mod.rs`: `ZoneState::last_published` and `LastPublished`
  - These were being used for zone status information.
- In `zone/mod.rs`: `ZoneState::{unsigned,signed}`, `{Unsigned,Signed}ZoneVersionState`, and `ZoneVersionReviewState`
  - These were being used for tracking the status of zone review.
  
An additional change in this PR is refactoring `ZoneServer::on_zone_review()`. This method is called from `LoadedReviewServer` and `SignedReviewServer`, and they contained TODOs for inlining `on_zone_review()` into them, as this would simplify the implementation. This PR heavily affected how zone review decisions are processed, so I had to rewrite `on_zone_review()`. I took the time to also inline it into `LoadedReviewServer` and `SignedReviewServer`; the code looks much better for it.

Open questions:

- [ ] How does this interact with zone restoration?
  - Should Cascade persist information about the current instance in its zone state file?
  - What should happen to this information if restoration fails?

Future work:

- Introduce instance IDs and use them for review. This will make review resilient against duplicate SOA serials (this is primarily a concern for loaded review).

- Track obsolete instances of zones. These instances are relevant to IXFR out, so this might serve as the right place to track information about them (including persisted diffs).

- Track abandoned instances of zones (zones that e.g. failed review). This could be nice for more zone history. We could choose to persist abandoned instances to disk for operators to review (e.g. if they use soft rejection but suspect a bug occurred), and store information about it here.

- Move some zone signer state from `ZoneState` (e.g. `min_expiration`) to `instance::SignedInstance`. This will fix a subtle bug where abandoning a signed instance can desynchronize the incremental signer (it might assume the instance was published, and this could affect re-signing schedules).

---


bal-e:
Closing in favor of #873.

PR #621, opened 2026-05-04
remove obsolete note about NetHSM

jans23:
remove obsolete note about NetHSM as discussed in #14

### DIFF
--- a/doc/manual/source/hsms.rst
+++ b/doc/manual/source/hsms.rst
@@ -66,16 +66,16 @@ to stress or performance test the interface. The tested HSMs are\:
 .. table:: Supported HSMs
    :widths: auto
 
-   =====================  =======================  =========  =================
-   HSM                    Type                     Interface  Integration guide
-   =====================  =======================  =========  =================
-   Fortanix DSM           Cloud                    KMIP       
-   Thales Cloud HSM       Cloud                    PKCS#11    :doc:`view <thales>`
-   Nitrokey NetHSM        Appliance, Docker image  PKCS#11    :doc:`view <nethsm>`
-   YubiHSM 2              USB key                  PKCS#11    
-   SoftHSM v2.6.1         Software                 PKCS#11    :doc:`view <softhsm>`
-   SmartCard-HSM          Smart Card               PKCS#11    :doc:`view <smartcard-hsm>`
-   =====================  =======================  =========  =================
+   ================  ======================  =========  =================
+   HSM               Type                    Interface  Integration guide
+   ================  ======================  =========  =================
+   Fortanix DSM      Cloud                   KMIP       
+   Thales Cloud HSM  Cldud                   PKCS#11    :doc:`view <thales>`
+   Nitrokey NetHSM   Hardware, Docker image  PKCS#11    :doc:`view <nethsm>`
+   YubiHSM 2         USB key                 PKCS#11    
+   SoftHSM v2.6.1    Software                PKCS#11    :doc:`view <softhsm>`
+   SmartCard-HSM     Smart Card              PKCS#11    :doc:`view <smartcard-hsm>`
+   ================  ======================  =========  =================
 
 .. Note:: Cascade requires TLS 1.3 for connections to the KMIP server, even
    though KMIP 1.2 requires servers to offer support for old versions of the
--- a/doc/manual/source/hsms.rst
+++ b/doc/manual/source/hsms.rst
@@ -66,16 +66,22 @@ to stress or performance test the interface. The tested HSMs are\:
 .. table:: Supported HSMs
    :widths: auto
 
-   ================  ======================  =========  =================
-   HSM               Type                    Interface  Integration guide
-   ================  ======================  =========  =================
-   Fortanix DSM      Cloud                   KMIP       
-   Thales Cloud HSM  Cldud                   PKCS#11    :doc:`view <thales>`
-   Nitrokey NetHSM   Hardware, Docker image  PKCS#11    :doc:`view <nethsm>`
-   YubiHSM 2         USB key                 PKCS#11    
-   SoftHSM v2.6.1    Software                PKCS#11    :doc:`view <softhsm>`
-   SmartCard-HSM     Smart Card              PKCS#11    :doc:`view <smartcard-hsm>`
-   ================  ======================  =========  =================
+   ====================  ======================  =========  =================
+   HSM                   Type                    Interface  Integration guide
+   ====================  ======================  =========  =================
+   Fortanix DSM          Cloud                   KMIP       
+   Thales Cloud HSM      Cldud                   PKCS#11    :doc:`view <thales>`
+   Nitrokey NetHSM [1]_  Hardware, Docker image  PKCS#11    :doc:`view <nethsm>`
+   YubiHSM 2             USB key                 PKCS#11    
+   SoftHSM v2.6.1        Software                PKCS#11    :doc:`view <softhsm>`
+   SmartCard-HSM         Smart Card              PKCS#11    :doc:`view <smartcard-hsm>`
+   ====================  ======================  =========  =================
+
+.. [1] Username and password must be specified in `p11nethsm.conf` for
+    both the `operator` and `admin` user in order to use the Nitrokey
+    NetHSM PKCS#11 module with Cascade-HSM-Bridge. See the
+    `Nitrokey Documentation <https://docs.nitrokey.com/nethsm/pkcs11-setup#users>`_
+    for more information.
 
 .. Note:: Cascade requires TLS 1.3 for connections to the KMIP server, even
    though KMIP 1.2 requires servers to offer support for old versions of the
--- a/doc/manual/source/hsms.rst
+++ b/doc/manual/source/hsms.rst
@@ -66,20 +66,16 @@ to stress or performance test the interface. The tested HSMs are\:
 .. table:: Supported HSMs
    :widths: auto
 
-   =====================  ============  =========  =================
-   HSM                    Type          Interface  Integration guide
-   =====================  ============  =========  =================
-   Fortanix DSM           Cloud         KMIP       
-   Thales Cloud HSM       Cloud         PKCS#11    :doc:`view <thales>`
-   Nitrokey NetHSM [1]_   Docker image  PKCS#11    :doc:`view <nethsm>`
-   YubiHSM 2              USB key       PKCS#11    
-   SoftHSM v2.6.1         Software      PKCS#11    :doc:`view <softhsm>`
-   SmartCard-HSM          Smart Card    PKCS#11    :doc:`view <smartcard-hsm>`
-   =====================  ============  =========  =================
-
-.. [1] Works with v1.7.2 of their PKCS#11 module. v2.0.0 and above are
-   NOT currently supported due to a `known bug in cascade-hsm-bridge
-   <https://github.com/NLnetLabs/cascade-hsm-bridge/issues/14>`_.
+   =====================  =======================  =========  =================
+   HSM                    Type                     Interface  Integration guide
+   =====================  =======================  =========  =================
+   Fortanix DSM           Cloud                    KMIP       
+   Thales Cloud HSM       Cloud                    PKCS#11    :doc:`view <thales>`
+   Nitrokey NetHSM        Appliance, Docker image  PKCS#11    :doc:`view <nethsm>`
+   YubiHSM 2              USB key                  PKCS#11    
+   SoftHSM v2.6.1         Software                 PKCS#11    :doc:`view <softhsm>`
+   SmartCard-HSM          Smart Card               PKCS#11    :doc:`view <smartcard-hsm>`
+   =====================  =======================  =========  =================
 
 .. Note:: Cascade requires TLS 1.3 for connections to the KMIP server, even
    though KMIP 1.2 requires servers to offer support for old versions of the


bal-e:
Hey, thanks for submitting a PR! Please give us time to test this ourselves before we update the documentation. We're currently busy preparing a beta release, but we hope we can get to this soon after.

bal-e:
Sorry, closing it didn't make sense. We'll leave it open until we've tested it, and then we can merge this PR.

ximon18:
So, I have been trying to verify that Cascade-HSM-Bridge currently works with NitroKey NetHSM... but I cannot get it to work.

This is my setup:
- Cascade commit 2b7a99c8 built in release mode
- Cascade-HSM-Bridge commit 1f529d10 built in release mode
- NitroKey NetHSM: I tried `nitrokey/nethsm:testing`, `nitrokey/nethsm:testing-v3.0` and `nitrokey/nethsm:testing-v4.0`.
- NitroKey NetHSM PKCS#11 module: https://github.com/Nitrokey/nethsm-pkcs11/releases/download/v2.2.0/nethsm-pkcs11-v2.2.0-x86_64-linux-glibc.so
- On Linux Fedora 43

NetHSM setup:
```
$ docker run -d --name hsm --rm -p 127.0.0.1:8443:8443 --env DEBUG_LOG=true nitrokey/nethsm:testing-v4.0

$ NETHSM_HOST="127.0.0.1:8443"
$ UNLOCK_PASS="something_between_10_and_20_characters"
$ ADMIN_PASS="something_else_between_10_and_20_characters"

$ curl -i -s --insecure -X 'POST' "https://${NETHSM_HOST}/api/v1/provision" -H 'accept: */*' -H 'Content-Type: application/json' -d '{"unlockPassphrase": "1234567890", "adminPassphrase": "'${ADMIN_PASS}'", "systemTime": "'$(date +%FT%TZ)'"}'

$ curl -i -s --user "admin:${ADMIN_PASS}" --insecure -X 'PUT' "https://${NETHSM_HOST}/api/v1/config/unattended-boot" -H 'accept: */*' -H 'Content-Type: application/json' -d '{"status": "on" }'

$ curl -i -s --user "admin:${ADMIN_PASS}" --insecure -X 'PUT' "https://${NETHSM_HOST}/api/v1/users/operator" -H 'accept: */*' -H 'Content-Type: application/json' -d '{"realName": "Operator", "role": "Operator", "passphrase": "'${ADMIN_PASS}'"

$ cat cascade-hsm-bridge.toml
version = "v1"

[pkcs11]
lib-path = "/home/ximon/nethsm-pkcs11.so"

[server]
addr = "127.0.0.1:5696"

[daemon]
log-level = "trace"
log-target = { type = "file", path = "/tmp/cascade-hsm-bridge.log" }
daemonize = true
pid-file = "/tmp/cascade-hsm-bridge.pid"

$ cat /tmp/p11nethsm.conf
# See: https://docs.nitrokey.com/nethsm/pkcs11-setup
log_level: Trace
slots:
  - label: DockerizedNetHSM
    description: NetHSM (Docker)
    operator:
      username: "operator"
    administrator:
      username: "admin"
    instances:
      - url: "https://127.0.0.1:8443/api/v1"
        danger_insecure_cert: true

$ export P11NETHSM_CONFIG_FILE=/tmp/p11nethsm.conf
$ cascade-hsm-bridge --config /tmp/cascade-hsm-bridge.toml
$ cascade --config /tmp/cascade.toml --state /tmp/cascade.state
$ cascade hsm add --insecure --username DockerizedNetHSM --password ${ADMIN_PASS} cascade-hsm-bridge 127.0.0.1
$ cascade template policy > /tmp/cascade/policies/default.toml
$ sed -i -e 's/^#hsm-server-id.\+/hsm-server-id = "cascade-hsm-bridge"/' "/etc/cascade/policies/default.toml"
$ cascade policy reload
$ cascade zone add --policy default --source /tmp/example.com example.com
```

Note that this relies on the admin and operator users having the same password and NOT specifying passwords in p11nethsm.conf, as we can only tell Cascade-HSM-Bridge one password to use.

With this Cascade-HSM-Bridge this gives me an error that I am not logged in. If I modify `ensure_logged_in()` to NOT test session state but just always login then that error goes away.

However then I am left with error CKR_FUNCTION_FAILED when calling `C_GenerateKeyPair()`. If I use pkcs11spy.so from the OpenSC project I see the following in the logs:

```
16: C_GenerateKeyPair
P:3740101; T:0x139917296293568 2026-06-01 13:34:14.187
[in] hSession = 0x1
[in] pMechanism->type = CKM_EC_KEY_PAIR_GEN
[in] pMechanism->pParameter = NULL
[in] pPublicKeyTemplate[9]:
    CKA_CLASS             CKO_PUBLIC_KEY
    CKA_ID                00007f408c006a20 / 20
    00000000  63 F2 0E 78 D1 EF E8 90 A5 51 1F 90 53 A0 84 2A  c..x.....Q..S..*
    00000010  62 05 72 AF                                      b.r.
    CKA_VERIFY            True
    CKA_ENCRYPT           False
    CKA_WRAP              False
    CKA_TOKEN             True
    CKA_EC_PARAMS         00007f408c003c90 / 10
    00000000  06 08 2A 86 48 CE 3D 03 01 07                    ..*.H.=...
    CKA_KEY_TYPE          CKK_EC
    CKA_LABEL             00007f408c0061d0 / 52
    32334445 494B3641 4E4D4E37 32414B36 41394849 44414D4E 39314538 34524E43
     2 3 D E  I K 6 A  N M N 7  2 A K 6  A 9 H I  D A M N  9 1 E 8  4 R N C38464641 39364635 46484C30 47463737 4A504447
     8 F F A  9 6 F 5  F H L 0  G F 7 7  J P D G
[in] pPrivateKeyTemplate[9]:
    CKA_CLASS             CKO_PRIVATE_KEY
    CKA_ID                00007f408c004f30 / 20
    00000000  63 F2 0E 78 D1 EF E8 90 A5 51 1F 90 53 A0 84 2A  c..x.....Q..S..*
    00000010  62 05 72 AF                                      b.r.
    CKA_SIGN              True
    CKA_DECRYPT           False
    CKA_TOKEN             True
    CKA_PRIVATE           True
    CKA_SENSITIVE         True
    CKA_KEY_TYPE          CKK_EC
    CKA_LABEL             00007f408c005ed0 / 52
    50394846 484E4544 554B554C 4F513244 48544E53 534D3334 39385454 53544452
     P 9 H F  H N E D  U K U L  O Q 2 D  H T N S  S M 3 4  9 8 T T  S T D R37384253 52433153 48363344 4D51424F 51453847
     7 8 B S  R C 1 S  H 6 3 D  M Q B O  Q E 8 G
Returned:  6 CKR_FUNCTION_FAILED
```

There is nothing useful in the output of `docker logs hsm`.

So, at the moment I cannot confirm that Cascade works with NitroKey NetHSM at all.

Have you been able to make it work? It used to work with an older version of the PKCS#11 module (as [documented by JP Mens](https://cascade.docs.nlnetlabs.nl/en/latest/nethsm.html))until NitroKey introduced an incompatibility with the PKCS#11 specification regarding allowed CKA_ID values, but they claim to have fixed that, so that shouldn't be the issue.

Note that I can still make Cascade work with SoftHSMv2 and with a YubiHSM, I'm only having this issue with NitroKey NetHSM (Docker testing image version).

ximon18:
I think Cascade-HSM-Bridge might be suppressing the logs of the NetHSM PKCS#11 module as that is also written in Rust and when I use `pkcs11-tool` with the module I see lots of Rust style log output but I don't see that in the Cascade-HSM-Bridge logging.

ximon18:
Hmm, I just got it working but not sure what I did.

The changes I made were:
- Revert back to `main` of Cascade-HSM-Bridge, specifically commit 998a000f9fa.
- Specify passwords explicitly in `p11nethsm.conf`.
- Set `daemonization = false` in the Cascade-HSM-Bridge configuration file.

ximon18:
Hmm, now it works with `daemonization = true` too.

jans23:
> However then I am left with error CKR_FUNCTION_FAILED when calling C_GenerateKeyPair(). 

This may be caused by using the Administrator instead of Operator user.

ximon18:
For Krill we have [an option to control whether or not Krill attempts to login to the HSM](https://github.com/NLnetLabs/krill/blob/609ba24efa09842d65f70ccf9699a4b59d42a716/defaults/krill-hsm.conf#L208). For Cascade-HSM-Bridge it is a bit "smarter", it doesn't login unless the operation is expected to need it, and we only ever in code login using the User role, not the SO role. It may be that we need more flexibility in the Casacde-HSM-Bridge configuration to cater for PKCS#11 modules that have different requirements/expectations of how they should be used.

jans23:
We are looking at adding support for Labels to NetHSM. AFAIK Cascade uses labels but does it need to *search* for labels too?

ximon18:
Hi @jans23,

Thank you very much for asking!

I'll get back to you as soon as I can, but that will likely be next week as I am on holiday today, apologies for the delay.

Ximon

ximon18:
> > However then I am left with error CKR_FUNCTION_FAILED when calling C_GenerateKeyPair().
> 
> This may be caused by using the Administrator instead of Operator user.

Thanks @jans23,

Indeed this appears to be the cause of the problem and I note that it is mentioned in the [Nitrokey Operation documentation](https://docs.nitrokey.com/nethsm/operation#generate-key).

The work-around with Cascade is to set the usernames and passwords for both the `operator` and `admin` roles in the `p11nethsm.conf` file so that the PKCS#11 module uses the appropriate credentials with NetHSM automatically.

What I don't understand is why NetHSM requires admin / SO rights to generate a key pair while other HSMs I have tested do not, reserving SO rights for things like PIN management. E.g.

- [Thales](https://thalesdocs.com/gphsm/ptk/protectserver3/docs/ps_ptk_docs/ptkc_programming/pkcs11_intro/index.html): _"The security officer’s only role is to initialize a token and set the normal user's access PIN."_
- [YubiHSM](https://docs.yubico.com/hardware/yubihsm-2/hsm-2-user-guide/hsm2-tools-pkcs11.html#logging-in): _"The concept of a Security Officer (SO) is not supported by the device, and the PIN management functions are not implemented, neither for user nor for SO."_
- [Fortanix DSM](https://support.fortanix.com/docs/fortanix-dsm-clients-pkcs11-supported-functions-and-mechanisms): _"Nobody can log in as a Security Officer (SO). Fortanix returns CKR_PIN_INCORRECT when someone tries to log in as an SO."_
- [AWS CloudHSM](https://docs.aws.amazon.com/cloudhsm/latest/userguide/pkcs11-pin.html): _"When you use the PKCS#11 library, your application runs as a particular [crypto user (CU)](https://docs.aws.amazon.com/cloudhsm/latest/userguide/manage-hsm-users.html) in your HSMs in AWS CloudHSM"_ and from [this page](https://docs.aws.amazon.com/cloudhsm/latest/userguide/understanding-users-cmu.html) one can see that the Crypto User (CU) can generate keys, while user management is restricted to the Crypto Officer (CO) user type.

Currently Cascade-HSM-Bridge does not support using the security officer (CKU_SO) user type to login and so is incompatible with NetHSM unless the use of the correct NetHSM user identity is left to the PKCS#11 module to handle.

ximon18:
> Hi @jans23,
> 
> Thank you very much for asking!
> 
> I'll get back to you as soon as I can, but that will likely be next week as I am on holiday today, apologies for the delay.
> 
> Ximon

@jans23: I haven't forgotten this, but it will take me some time to give an answer I am completely sure of and first I need to finish something else.

jans23:
> The work-around with Cascade is to set the usernames and passwords for both the operator and admin roles in the p11nethsm.conf

Why is this a workaround and what would be the expected solution? From our perspective it is expected to set the credentials in p11nethsm.conf.

> why NetHSM requires admin / SO rights to generate a key pair

We don't have a sophisticated user role system yet and wanted to prevent that an overtaken user could maliciously delete keys. Ideally we would support more roles and a more flexible permission system but don't have such yet.

Can you deal with this and find a solution in Cascade?

ximon18:
@jans23: Yes, the solution in Cascade is to document how to make it work with NetHSM, as it **does** work, if `p11nethsm.conf` is configured correctly.

> Why is this a workaround and what would be the expected solution?

I would expect the application that calls the `C_Login()` function to be the one to provide the user type and PIN, not for that to he magically handled behind the scenes. Most PKCS#11 using applications will I assume need to support taking authentication details from the operator and passing them to the PKCS#11 module. 

I find it then confusing for the operator to have to configure both layers separately with credentials when what the first layer does with them is irrelevant because of magic at the second layer.

I don't recall the NetHSM PKCS#11 module advertising via `GetTokenInfo()` that it doesn't require login, that might make it easier to code the PKCS#11 host application if it could support login-less mode and thus not require the operator to supply both it and the PKCS#11 module with credentials.

I don't have the code in front of me so apologies if I am misunderstanding how to use the NetHSM PKCS#11 module or misremembering parts of the PKCS#11 specifications.

sosthene-nitrokey:
The nethsm PKCS#11 module does advertise that `CKF_LOGIN_REQUIRED` **only** if there is no configured password in the config file (note that it will not advertise `CKF_LOGIN_REQUIRED`  it if the operator is configured but not the administrator).

ximon18:
Slightly related to this, see this post I just made on our community site: https://community.nlnetlabs.nl/t/running-cascade-using-containers/

doc/manual/source/hsms.rst
```diff
@@ -66,20 +66,16 @@ to stress or performance test the interface. The tested HSMs are\:
...
-   =====================  ============  =========  =================
-   Fortanix DSM           Cloud         KMIP       
-   Thales Cloud HSM       Cloud         PKCS#11    :doc:`view <thales>`
-   Nitrokey NetHSM [1]_   Docker image  PKCS#11    :doc:`view <nethsm>`
-   YubiHSM 2              USB key       PKCS#11    
-   SoftHSM v2.6.1         Software      PKCS#11    :doc:`view <softhsm>`
-   SmartCard-HSM          Smart Card    PKCS#11    :doc:`view <smartcard-hsm>`
-   =====================  ============  =========  =================
-
-.. [1] Works with v1.7.2 of their PKCS#11 module. v2.0.0 and above are
```
ximon18, line 80: @jans23: Maybe we should keep the note but change the text to something like:

```
.. [1] Username and password must be specified in `p11nethsm.conf` for
    both the `operator` and `admin` user in order to use the Nitrokey
    NetHSM PKCS#11 module with Cascade-HSM-Bridge.
```

What do you think?

jans23: Sounds good.

ximon18: As @jans23: this PR is from your own forked repo I can't edit it or make suggested edits either it seems. If you make the changes, maybe even with a link to Nitrokey docs, I can merge it. Something like this perhaps?

```
.. [1] Username and password must be specified in `p11nethsm.conf` for
    both the `operator` and `admin` user in order to use the Nitrokey
    NetHSM PKCS#11 module with Cascade-HSM-Bridge. See the
    `Nitrokey Documentation <https://docs.nitrokey.com/nethsm/pkcs11-setup#users>`_
    for more information.
```

jans23: Added.

ximon18 (approved):

ximon18:
> We are looking at adding support for Labels to NetHSM. AFAIK Cascade uses labels but does it need to _search_ for labels too?

Hi @jans23,

I was finally able to make time to check.

No, Cascade-HSM-Bridge does not search for keys by label, it only ever looks up objects using `C_FindObjectsXxx()` by their PKCS#11 `CKA_ID` and `CK_OBJECT_CLASS` (`CKO_PRIVATE_KEY` or `CKO_PUBLIC_KEY`).

Ximon

jans23:
Thank you for the feedback. We do support these look ups.

PR #623, opened 2026-05-04
Add signature-refresh-interval and key-roll-time.

Philip-NLnetLabs:
Add signature-refresh-interval and key-roll-time to template.

### DIFF
--- a/etc/policy.template.toml
+++ b/etc/policy.template.toml
@@ -338,6 +338,7 @@ signature-remain-time = "1w"
 #
 # signature-refresh-interval should be a lot smaller than
 # signature-remain-time to make sure that signatures are refreshed in time.
+# If this is not the case then in extreme cases, signatures could expire.
 #
 # An integer value is interpreted as seconds. A string is interpreted as time
 # string with a number followed by a unit (i.e. "s", "m", "h", "d", or "w").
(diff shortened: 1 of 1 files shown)


etc/policy.template.toml
```diff
@@ -329,6 +329,23 @@ signature-lifetime = "2w"
 # string with a number followed by a unit (i.e. "s", "m", "h", "d", or "w").
 signature-remain-time = "1w"
 
+# To prevent signatures from expiring, Cascade will periodically refresh
```
ximon18: This lacks a top single line description of what the value actually represents.

Also, does it relate to signature-remain-time, and if so how, it is unclear from this description how it works.

E.g. is the fraction of signatures that will be renewed derived from these two settings, i.e. if remain time is 24h and refresh interval is 12h does that mean 50% of signatures will be renewed every 12h?

Also, what happens if this value is larger than remain time?

Philip-NLnetLabs: I added more text.

ximon18: What happens if the `signature-refresh-interval` value is NOT small enough? Will some signatures fail to be refreshed? Will a full zone sign occur? Will an error occur?

ximon18: Also, so what then is `key-roll-time` for? I don't get it.

ximon18 (approved):

PR #626, opened 2026-05-05
Persist and restore TSIG keys for zone sources

### LINKED ISSUE #301: TSIG support

bal-e:
A re-implementation of #590. GH seems confused about that PR, so it's hard to see it's original diff, but I think this is the core of it.

---

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -76,10 +76,19 @@ fn main() -> ExitCode {
     // Load the global state file or build one from scratch.
     let mut zones = Default::default();
     let mut policies = Default::default();
-    let mut state = match center::State::init_from_file(&config, &mut zones, &mut policies) {
+    let state = match center::State::init_from_file(&config, &mut zones, &mut policies) {
         Ok(mut state) => {
-            // TODO: Restore the TSIG key store here, so that keys are available
-            // to the zones and policies being restored.
+            // Load the TSIG store file.
+            match state.tsig_store.init_from_file(&config) {
+                Ok(()) => debug!("Loaded the TSIG store"),
+                Err(err) if err.kind() == io::ErrorKind::NotFound => {
+                    debug!("No TSIG store found; will create one");
+                }
+                Err(err) => {
+                    error!("Failed to load the TSIG store: {err}");
+                    return ExitCode::FAILURE;
+                }
+            }
 
             // Restore pending zones.
             for name in zones {
@@ -87,10 +96,11 @@ fn main() -> ExitCode {
                     !state.zones.contains(&name),
                     "Zone '{name}' was encountered twice"
                 );
-                let zone = match Zone::restore(&config, name, &mut state.policies) {
-                    Ok(zone) => zone,
-                    Err(_) => return ExitCode::FAILURE,
-                };
+                let zone =
+                    match Zone::restore(&config, name, &mut state.policies, &state.tsig_store) {
+                        Ok(zone) => zone,
+                        Err(_) => return ExitCode::FAILURE,
+                    };
                 state.zones.insert(ZoneByName(Arc::new(zone)));
             }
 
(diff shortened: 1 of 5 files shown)


ximon18:
Manual testing with the `.ch` zone shows this working for me, the TSIG key is again used after a restart to query the source nameserver which without this PR is not the case.

src/main.rs
```diff
@@ -76,21 +76,31 @@ fn main() -> ExitCode {
...
     let mut zones = Default::default();
     let mut policies = Default::default();
-    let mut state = match center::State::init_from_file(&config, &mut zones, &mut policies) {
+    let state = match center::State::init_from_file(&config, &mut zones, &mut policies) {
         Ok(mut state) => {
-            // TODO: Restore the TSIG key store here, so that keys are available
-            // to the zones and policies being restored.
+            // Load the TSIG store file.
+            match state.tsig_store.init_from_file(&config) {
+                Ok(()) => debug!("Loaded the TSIG store"),
```
ximon18: Inconsistent logging:
- `State::init_from_file()` loads the file then logs at _info_ level **and** notes the path that was loaded from.
- `TsigStore::init_from_file()` doesn't log, instead `main` does, and that log message doesn't mention the path that was loaded from.

Also, though not caused by this PR, neither for main state nor TSIG store state do we log the path we are trying to load from before loading, nor on error, so in the unhappy case the path used is not mentioned at all - unless I'm missing something.

src/main.rs
```diff
@@ -137,6 +147,18 @@ fn main() -> ExitCode {
 
             let mut state = center::State::default();
 
+            // Load the TSIG store file.
```
ximon18, line 167: Why would there be a TSIG store file if there was no state file?

src/zone/state/mod.rs
```diff
@@ -146,3 +151,65 @@ impl Spec {
...
+}
+
+impl fmt::Display for LoadError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Read { path, error } => {
+                write!(f, "could not read the zone state file '{path}': {error}")
+            }
+            Self::MissingSourceTsigKey(error) => {
+                write!(f, "could not load the zone source setting: {error}")
```
ximon18, line 191: How does one "load" a "setting"? Should this text not say something about a missing TSIG key?

ximon18 (approved):
LGTM. Some minor comments.

src/main.rs
bal-e: Maybe the user deleted the state file (or lost it) and doesn't want to re-enter their TSIG keys?

src/zone/state/mod.rs
bal-e: This can be phrased differently, but I was referring to the loading of the zone from the zone state file. The underlying `{error}` does explicitly say the TSIG key is missing.

ximon18: Right, I was confused by the name of the enum variant then.

src/main.rs
ximon18: Okay, but we don't generally support restoring from a partial/corrupted state, e.g. once the state file is lost any zone state files become orphans, those zones are no longer reflected in main state IIRC. This seems a bit inconsistent.

src/main.rs
bal-e: You're right; I've added a commit which logs every time.

PR #631, opened 2026-05-06
Save last serial and key tags in zone state.

Philip-NLnetLabs:
The incremental signer introduces a few more state variables. Update them from the full zone signer.

---

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -428,8 +428,8 @@ fn check_dnst_version(config: &Config) -> bool {
     // Change this string and the match pattern to whatever version we require in the future
     let required_version = ">0.2.0";
     let res = match (major, minor, patch) {
-        // major = 0; minor >= 1; patch = *
-        (0, 1.., ..) => true,
+        // major = 0; minor >= 2; patch = *
+        (0, 2.., ..) => true,
         _ => false,
     };
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -426,7 +426,7 @@ fn check_dnst_version(config: &Config) -> bool {
     };
 
     // Change this string and the match pattern to whatever version we require in the future
-    let required_version = ">0.1.0";
+    let required_version = ">0.2.0";
     let res = match (major, minor, patch) {
         // major = 0; minor >= 1; patch = *
         (0, 1.., ..) => true,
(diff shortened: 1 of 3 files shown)


src/units/zone_signer.rs
```diff
@@ -362,26 +364,22 @@ impl ZoneSigner {
...
-                    serial = previous_serial.inc(1);
-                }
-                serial
+                // Always increment the serial number, ignore the serial
+                // number in the unsigned zone.
+                let previous_serial = if let Some(serial) = previous_serial {
+                    serial
+                } else {
+                    Serial::from(0)
+                };
```
bal-e: [Minor] This can be simplified to `previous_serial.unwrap_or(Serial::from(0))`.

src/units/zone_signer.rs
```diff
@@ -394,23 +392,25 @@ impl ZoneSigner {
...
+        let serial = NewBaseSerial::from(serial.into_int());
         let new_soa = {
             let mut soa = loaded.soa().clone();
             soa.rdata.serial = serial;
             soa
         };
 
         info!(
-            "[ZS]: Serials for zone '{zone_name}': last signed={last_signed_serial:?}, current={loaded_serial}, serial policy={}, new={serial}",
+            "[ZS]: Serials for zone '{zone_name}': last signed={previous_serial:?}, current={loaded_serial}, serial policy={}, new={serial}",
```
bal-e, line 408: I think this changes the meaning of this field. `last_signed_serial` _used_ to be (IIRC) the serial of the currently published zone. But `previous_serial` (AFAICT) refers to the latest used serial, even if the instance it was used for was abandoned. Do we want to change the `last signed=` bit to account for that?

src/main.rs
bal-e, line 434: This looks like a bug; you changed the string but not the match pattern.

bal-e (commented):
Mostly good, just one bug.

src/units/zone_signer.rs
Philip-NLnetLabs: No, serials should be unique. It could be confusing if the signed review script sees the same serial twice with different content.

src/units/zone_signer.rs
Philip-NLnetLabs: done.

src/main.rs
Philip-NLnetLabs: fixed.

ximon18 (approved):
@Philip-NLnetLabs and I went through this together, LGTM.

PR #633, opened 2026-05-06
Report an error for violation of serial keep policy

### LINKED ISSUE #476: zone reload command without changing the SOA number can put cascade server into weird state.
When enabled serial-policy = "keep" for policy, Cascaded can went into error when the reload command is ran without change the SOA.

Success Case: 
1. update the SOA on zone file, 
2. run "cascade zone reload $zone"
3. zone updated properly.

Failed Case. 
1. run "cascade zone reload $zone" without updating the SOA
2. zone "locked" with error and cannot be update again unless restart cascaded 

expected behavior: 
Cascaded should be able to reload the zone again once the SOA number updated properly, and the http-remote control should still accept reload command. 

log 

```
root@CIRA-20240001:/var/lib/cascade/zone-state# cascade zone status example.test
Status report for zone 'example.test' using policy 'ecdsa-policy'
✔ Waited for a new version of the example.test zone
✔ Loaded version 6
  Loaded at 2026-02-23T10:47:45 (3s ago)
  Loaded 170 B and 3 records from the filesystem in 0 seconds
✔ Auto approving signing of version 6, no checks enabled in policy.
✔ Approval received to sign version 6, signing requested
✔ Signed version 6 as version 6
  Signing requested at 2026-02-23T10:47:45 (3s ago)
  Signing started at 2026-02-23T10:47:45 (3s ago)
  Signing finished at 2026-02-23T10:47:45 (3s ago)
  Collected 3 records in 0s, sorted in 0s
  Generated 3 NSEC(3) records in 0s
  Generated 6 signatures in 0s (6 sig/s)
  Inserted signatures in 0s (6 sig/s)
  Took 0s in total, using 2 threads
  Current action: Finished
✔ Auto approving publication of version 6, no checks enabled in policy.
✔ Published version 6
  Published zone available on 127.0.0.1:4543
root@CIRA-20240001:/var/lib/cascade/zone-state# cascade zone reload example.test
Success: Sent zone reload command for example.test
root@CIRA-20240001:/var/lib/cascade/zone-state# cascade zone status example.test
Status report for zone 'example.test' using policy 'ecdsa-policy'
✔ Waited for a new version of the example.test zone
✔ Loaded version 6
  Loaded at 2026-02-23T10:47:50 (2s ago)
  Loaded 170 B and 3 records from the filesystem in 0 seconds
✔ Auto approving signing of version 6, no checks enabled in policy.
✔ Approval received to sign version 6, signing requested
✔ Signed version 6 as version 6
  Signing requested at 2026-02-23T10:47:50 (8s ago)
  Signing started at 2026-02-23T10:47:50 (8s ago)
  Signing finished at 2026-02-23T10:47:50 (8s ago)
  Collected 3 records in 0s, sorted in 0s
  Generated 3 NSEC(3) records in 0s
  Generated 6 signatures in 0s (6 sig/s)
  Inserted signatures in 0s (6 sig/s)
  Took 0s in total, using 2 threads
  Current action: Finished
✔ Auto approving publication of version 6, no checks enabled in policy.
✔ Published version 6
  Published zone available on 127.0.0.1:4543
**x The pipeline for this zone is hard halted due to a serious error:
x Serial policy is Keep but upstream serial did not increase**


root@CIRA-20240001:/var/lib/cascade/zone-state# cascade zone reload example.test
2026-02-23T15:48:08.074451Z ERROR cascade: Error: Failed to reload zone: the zone has been halted (reason: Serial policy is Keep but upstream serial did not increase)
```

our example policy file:

```
root@CIRA-20240001:/var/lib/cascade/zone-state# cat /etc/cascade/policies/ecdsa-policy.toml
version = "v1"

[key-manager.generation]
# Use the existing SoftHSM-backed KMIP server
hsm-server-id = "kmip2pkcs11"

# ECDSA P-256
algorithm = "ECDSAP256SHA256"

[signer]
signature-inception-offset = "1d"
signature-lifetime = "2w"
signature-remain-time = "1w"
serial-policy = "keep"

[signer.denial]
type = "nsec3"

# NSEC3 parameters

# THIS enables opt-out
opt-out = true

```

tertsdiepraam:
Closes #476

Tests have been run by Philip. They are passing.

### DIFF
--- a/src/signer/mod.rs
+++ b/src/signer/mod.rs
@@ -102,7 +102,20 @@ async fn sign(
             // Something in status would be good.
             handle.abandon_signing(builder);
             status.status.finish(true);
+
             status.current_action = "Resign failed due to Keep policy".to_string();
+
+            let error =
+                "serial policy is \"keep\" but the serial of the loaded zone did not increase";
+
+            error!("Signing failed: {error}");
+            handle.state.record_event(
+                HistoricalEvent::SigningFailed {
+                    trigger: trigger.into(),
+                    reason: error.to_string(),
+                },
+                None, // TODO
+            );
         }
         Err(error) => {
             error!("Signing failed: {error}");
--- a/src/signer/mod.rs
+++ b/src/signer/mod.rs
@@ -23,7 +23,6 @@ use std::{
 };
 
 use cascade_zonedata::SignedZoneBuilder;
-use domain::rdata::dnssec::Timestamp;
 use tracing::{debug, error};
 
 use crate::units::zone_signer::SignerError;
@@ -106,23 +105,9 @@ async fn sign(
 
             status.current_action = "Resign failed due to Keep policy".to_string();
 
-            // TODO: Ideally, we wouldn't trigger resigning if the policy is keep,
-            // but currently we do, so we have to use a heuristic to determine whether
-            // this was a resign triggered by Cascade or a reload triggered by the user
-            // that erroneously did not have an incremented SOA.
-            //
-            // The heuristic is that we try to compute whether Cascade would
-            // have triggered a resign based on the min_expiration and the
-            // policy.
-            let Some(min_expiration) = handle.state.min_expiration else {
-                return;
-            };
-            let sig_remain_time = handle.state.policy.as_ref().unwrap().signer.sig_remain_time;
-            let now = Timestamp::now().into_int();
-
-            if min_expiration.into_int().saturating_sub(sig_remain_time) > now {
-                // In this branch, Cascade wouldn't have triggered a resign, so
-                // it must be an issue with the loaded zone.
+            // If the sign operation was triggered by a load, the user forgot to increase the
+            // serial of the zone, so we should tell them about that by emitting an error.
+            if trigger == SigningTrigger::Load {
                 let error =
                     "serial policy is \"keep\" but the serial of the loaded zone did not increase";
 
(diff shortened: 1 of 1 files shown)


Philip-NLnetLabs:
This would be fine if the signing request is due to a new version of the zone. But signing requests also result from key management and timers. And I think we don't want an error for those cases.

ximon18:
> This would be fine if the signing request is due to a new version of the zone. But signing requests also result from key management and timers. And I think we don't want an error for those cases.

I mistakenly set a zone to keep policy while also giving it a fast incremental resign rate which leads to Cascade repeatedly trying and failing to sign the zone due to keep policy violation, except that violation doesn't show up anywhere so I didn't realize what was going on. This is admittedly user error on my part, but this doesn't seem a useful state for Cascade to silently be in.

Philip-NLnetLabs:
If we make it not silent then we basically kill Keep because it will generate an error every time the refresh timer expires. Something at debug level is fine of course.

tertsdiepraam:
@Philip-NLnetLabs could you explain how you want this to work? I think the signer should not even start any resigning with keep. So we should add that to this PR maybe? Is there an easy way to do that?

Philip-NLnetLabs:
I think I know what we can do. Take min_expiration (in zone state) and subtract sig_remain_time from policy. If the result is less than now, then issue an error. If the result is greater than now, then issue a debug log message.

src/signer/mod.rs
```diff
@@ -102,7 +103,40 @@ async fn sign(
...
+            // have triggered a resign based on the min_expiration and the
+            // policy.
+            let Some(min_expiration) = handle.state.min_expiration else {
+                return;
+            };
+            let sig_remain_time = handle.state.policy.as_ref().unwrap().signer.sig_remain_time;
+            let now = Timestamp::now().into_int();
+
+            if min_expiration.into_int().saturating_sub(sig_remain_time) > now {
+                // In this branch, Cascade wouldn't have triggered a resign, so
```
Philip-NLnetLabs: This assumption is incorrect. Why not just implement what I wrote before.

Philip-NLnetLabs (approved):

PR #634, opened 2026-05-07
Add auto-remove-delay policy option.

Philip-NLnetLabs:
Auto remove delay was added to keyset a while ago but never made it to Cascade.
---

### DIFF
--- a/src/policy/file/v1.rs
+++ b/src/policy/file/v1.rs
@@ -57,7 +57,7 @@ const SIGNATURE_REFRESH_INTERVAL: u32 = 12 * 3600;
 // sign more records.
 const KEY_ROLL_TIME: u32 = 24 * 3600;
 
-// When auto remove is enable, remove old keys after one week.
+// When auto remove is enabled, remove old keys after one week.
 const AUTO_REMOVE_DELAY: u32 = 7 * 24 * 3600;
 
 //----------- Spec -------------------------------------------------------------
--- a/src/policy/file/v1.rs
+++ b/src/policy/file/v1.rs
@@ -1,5 +1,6 @@
 //! Version 1 of the policy file.
 
+use std::time::Duration;
 use std::{
     fmt::{self, Display},
     net::{IpAddr, SocketAddr},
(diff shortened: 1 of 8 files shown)


src/zone/state/v1.rs
```diff
@@ -245,9 +246,12 @@ pub struct KeyManagerPolicySpec {
     /// The TTL to use when creating DNSKEY/CDS/CDNSKEY records.
     default_ttl: Ttl,
 
-    /// Automatically remove keys that are no long in use.
+    /// Automatically remove keys that are no lerong in use.
```
Philip-NLnetLabs:
```suggestion
    /// Automatically remove keys that are no longer in use.
```

src/policy/file/v1.rs
```diff
@@ -56,6 +57,9 @@ const SIGNATURE_REFRESH_INTERVAL: u32 = 12 * 3600;
 // sign more records.
 const KEY_ROLL_TIME: u32 = 24 * 3600;
 
+// When auto remove is enable, remove old keys after one week.
```
ximon18:
```suggestion
// When auto remove is enabled, remove old keys after one week.
```

src/state/v1.rs
```diff
@@ -244,6 +244,9 @@ pub struct KeyManagerPolicySpec {
     /// Automatically remove keys that are no long in use.
     auto_remove: bool,
 
+    /// Remove old keys after this amount of time.
+    auto_remove_delay: u64,
```
ximon18, line 248: Why is this u64 when the `AUTO_REMOVE_DELAY` constant is u32 and here also for example `cds_inception_offset` is u32?

src/zone/state/v1.rs
```diff
@@ -245,9 +246,12 @@ pub struct KeyManagerPolicySpec {
     /// The TTL to use when creating DNSKEY/CDS/CDNSKEY records.
     default_ttl: Ttl,
 
-    /// Automatically remove keys that are no long in use.
+    /// Automatically remove keys that are no longer in use.
     auto_remove: bool,
 
+    /// Remove old keys after this amount of time.
+    auto_remove_delay: u64,
```
ximon18, line 253: Same comment about u32 as above.

ximon18 (approved):
LGTM. Some minor feedback.

src/state/v1.rs
Philip-NLnetLabs: The whole thing is a mess. I think these types needs to be u64 because that is what Duration uses. But I don't want to change everything else at the moment.

PR #637, opened 2026-05-08
Fix the position of Tsig in the service layers.

Philip-NLnetLabs:


ximon18 (approved):
The `downstream-tsig` system test should be failing because of the problem that this PR fixes, but due to a bug in the test that failure went unnoticed. PR #638 fixes the `downstream-tsig` system test so that it correctly fails. I tested that this PR then makes the system test work again by fixing the underlying problem.

PR #638, opened 2026-05-08
Check dig result properly

ximon18:
TSIG response signing is actually failing but wasn't detected by the `downstream-tsig` system test due to wrongly assuming that `dig` would exit with a non-zero exit code if there is a problem with the response. However, dig will exit with code 0 if it receives any DNS response, success or failure.

This PR makes the dig check stricter. The dig check was also failing to use the TSIG key when it should which this PR also fixes.

This test detects the TSIG failure but doesn't fix it. PR #637 has been tested with this PR and causes the TSIG error reported by `dig` to disappear.

---

### DIFF
--- a/integration-tests/tests/downstream-tsig/action.yml
+++ b/integration-tests/tests/downstream-tsig/action.yml
@@ -45,10 +45,13 @@ runs:
         log-level: ${{ inputs.log-level }}
 
     - name: Determine the downstrem secondary zone config to use
-      id: set-zone-name
+      id: set-vars
       run: |
+        TSIG_SPEC="hmac-sha256:tsig-key:COzoVsYQmXeXiyq1Quhp0bbVnMyxjPxsaGSoIWR98i0="
+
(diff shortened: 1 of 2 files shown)


integration-tests/tests/downstream-tsig/action.yml
```diff
@@ -108,37 +113,47 @@ runs:
...
+        ZONE_NAME: ${{ steps.set-vars.outputs.ZONE_NAME }}
+        TSIG_SPEC: ${{ steps.set-vars.outputs.TSIG_SPEC }}
       run: |
-        dig @127.0.0.1 -p 4542 ${{ steps.set-zone-name.outputs.ZONE_NAME }} SOA &> soa-query.log
+        EXTRA_ARGS=
+        if [[ "${TSIG_SPEC}" != "" ]]; then
+          EXTRA_ARGS="-y ${TSIG_SPEC}"
+        fi
+        dig @127.0.0.1 -p 4542 ${EXTRA_ARGS} ${ZONE_NAME} SOA &> soa-query.log
+        ! grep -Fq "Transfer failed" soa-query.log
```
tertsdiepraam, line 143: TIL what that `!` does 😄

tertsdiepraam (approved):

PR #639, opened 2026-05-10
Document incremental signing

bal-e:
This is based on my understanding of the incremental signer. I couldn't get my editor to highlight rST properly, so I'm not sure I followed the syntax correctly; feel free to make changes.

### DIFF
--- a/doc/manual/source/incremental-signing.rst
+++ b/doc/manual/source/incremental-signing.rst
@@ -1,0 +1,105 @@
+Incremental Signing
+===================
+
+.. note:: This page documents how Cascade's incremental signer works, so that
+   you can understand what to expect from its output, and how the associated
+   settings will affect it. 
+   
+   Note that Cascade does *not* use a jitter-based mechanism (as one would
+   find in OpenDNSSEC) for spreading out the re-signing of records. Read on to
+   understand Cascade's "re-signing window" mechanism.
+
+Context
+-------
+
+Cascade can sign your zone in two modes: full signing and incremental signing.
+Full signing unconditionally signs every record in the loaded zone, even if
+some of those records had valid signatures. Incremental signing will examine
+previously generated signatures and only sign records that are missing
+signatures or whose signatures need updating. Incremental signing is more
+complicated than full signing, but it has two benefits:
+
+1. It greatly reduces the number of signatures that need to be generated. This
+   is important for performance, especially if an :term:`HSM <Hardware Security
+   Module (HSM)>` is used.
+
+2. It greatly reduces the number of records that change in the signed zone.
+   With full signing, every change to the loaded zone would result in brand
+   new signatures for *every* record. With incremental signing, small changes
+   in the loaded zone will translate to small changes in the signed zone. This
+   reduces the size of :term:`IXFRs <Incremental Zone Transfers (IXFRs)>`,
+   improving network latency and bandwidth for propagating the zone.
+
+Cascade uses full signing for the first time it signs a zone, as it does not
+know of any previous signatures. After the initial signing operation, a signed
+copy of the zone will always be available, so Cascade will only use incremental
+signing. In the future, we may add support for forcing the use of the full
+signer even if a signed copy of the zone is available.
+
+Key Rollovers
+-------------
+
+Cascade has a key manager, which can direct the signer to change the set of
+keys it signs the zone with. With full signing, this is relatively simple:
+all signatures will be re-generated using the new set of signing keys. But for
+the reasons discussed above, this is suboptimal. Thus, the incremental signer
+supports performing key rollovers over a period of time, where it will gradually
+update the signatures of all the records in the zone. Re-signing is performed
+such that the zone is always DNSSEC valid; for most key rollovers, this implies
+signing all records with an additional key, and once the new signatures have
+propagated, removing signatures with a previous key.
+
+The incremental signer follows the same procedure whether a key rollover is
+occurring or not. Internally, it will first determine which signatures need
+updating, and then calculate when they should be updated. If a key rollover
+is occurring, it will mark signatures signed with the previous set of keys to
+be updated.
+
+Algorithm
+---------
+
+The incremental signer has to satisfy the following needs:
+
+1. If records in the loaded zone have changed, new signatures need to be
+   generated for them, and the NSEC/NSEC3 chain must be updated.
+
+2. If signatures are nearing (a particular point before) expiry, they will be
+   regenerated.
+
+3. If a key rollover is ongoing (as discussed above), new signatures need to be
+   generated if new keys are being added, and old signatures need to be removed
+   if old keys are being removed.
+
+These three needs are handled independently.
+
+First of all, all changes in the loaded zone are accounted for, from NSEC/NSEC3
+chain updates to generating new signatures.
+
+Then: every signature that needs re-generating (because it is sufficiently close
+to expiry) will be considered. Signatures that are _too_ close to expiry will be
+re-generated immediately. The rest will be sorted into a _re-signing schedule_,
+which is a deterministic way to spread out incremental signing work. They will
+be sorted by ascending order of expiry, and then by DNSSEC canonical order. The
+first ``N`` such signatures will be re-generated, where ``N`` is chosen so that
+re-signing will complete within a certain period of time.
+
+Finally, the ongoing key rollover (if any) is handled. If the key rollover has
+taken longer than planned, it will be completed immediately. Otherwise, another
+re-signing schedule will be constructed. The signatures pending updates will
+be sorted in DNSSEC canonical order and the first ``N`` such signatures will
+be re-generated.
+
+Configuration
+-------------
+
+The incremental signer can be configured using the following policy settings:
+
+* `signer.signature-refresh-interval`: The interval at which the incremental
+  signer will re-generate signatures pending expiry and update signatures for
+  key rollovers.
+
+* `signer.key-roll-time`: The total amount of time a key rollover
+  should take. Signature updates (which happen periodically at interval
+  `signer.signature-refresh-interval`) will be spread out across this. If the
+  key rollover takes longer than this, all remaining signatures will be updated
+  at the next invocation of the incremental signer.
--- a/doc/manual/source/incremental-signing.rst
+++ b/doc/manual/source/incremental-signing.rst
@@ -1,59 +1,60 @@
 Incremental Signing
 ===================
 
-.. note:: This page documents how Cascade's incremental signer works, so that
-   you can understand what to expect from its output, and how the associated
-   settings will affect it. 
+This page documents how Cascade's incremental signer works, so that you can
+understand what to expect from its output, and how the associated settings will
+affect it.
    
-   Note that Cascade does *not* use a jitter-based mechanism (as one would
-   find in OpenDNSSEC) for spreading out the re-signing of records. Read on to
-   understand Cascade's "re-signing window" mechanism.
+Note that Cascade does *not* use a jitter-based mechanism (as one would find
+in OpenDNSSEC) for spreading out the re-signing of records. It uses a different
+mechanism to achieve this, a "re-signing schedule"; read on to learn more.
 
 Context
 -------
 
-Cascade can sign your zone in two modes: full signing and incremental signing.
-Full signing unconditionally signs every record in the loaded zone, even if
-some of those records had valid signatures. Incremental signing will examine
-previously generated signatures and only sign records that are missing
-signatures or whose signatures need updating. Incremental signing is more
-complicated than full signing, but it has two benefits:
+Cascade can sign zones in either of two modes: full signing and incremental
+signing. It will automatically choose one. Full signing unconditionally signs
+every record in the loaded zone, even if some of those records had valid
+signatures. Incremental signing will examine previously generated signatures
+and only sign records that are missing signatures or whose signatures need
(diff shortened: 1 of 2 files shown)


doc/manual/source/incremental-signing.rst
```diff
@@ -0,0 +1,105 @@
...
+Key Rollovers
+-------------
+
+Cascade has a key manager, which can direct the signer to change the set of
+keys it signs the zone with. With full signing, this is relatively simple:
+all signatures will be re-generated using the new set of signing keys. But for
+the reasons discussed above, this is suboptimal. Thus, the incremental signer
+supports performing key rollovers over a period of time, where it will gradually
+update the signatures of all the records in the zone. Re-signing is performed
+such that the zone is always DNSSEC valid; for most key rollovers, this implies
```
Philip-NLnetLabs, line 48: For a typical ZSK roll, you first introduce the new key and then replace the signatures. What you describe is also possible but less efficient. Algorithm rolls do require double signatures.

doc/manual/source/incremental-signing.rst
```diff
@@ -0,0 +1,105 @@
...
+signing all records with an additional key, and once the new signatures have
+propagated, removing signatures with a previous key.
+
+The incremental signer follows the same procedure whether a key rollover is
```
Philip-NLnetLabs: There are two steps. The first is to update signatures that should be refreshed. The second is to update signatures during a key roll. The second step is skipped if there is no key roll. Of course new signatures are generated with the new key. So the first step helps during a key roll.

doc/manual/source/incremental-signing.rst
```diff
@@ -0,0 +1,105 @@
...
+to expiry) will be considered. Signatures that are *too* close to expiry will be
+re-generated immediately. The rest will be sorted into a *re-signing schedule*,
+which is a deterministic way to spread out incremental signing work. They will
+be sorted by ascending order of expiry, and then by DNSSEC canonical order. The
+first ``N`` such signatures will be re-generated, where ``N`` is chosen so that
+re-signing will complete within a certain period of time.
+
+Finally, the ongoing key rollover (if any) is handled. If the key rollover has
+taken longer than planned, it will be completed immediately. Otherwise, another
+re-signing schedule will be constructed. The signatures pending updates will
```
Philip-NLnetLabs: This is a different N. All signatures will be considered not just the ones that need update. When signing records that already have signatures from the right key will be skip but still count towards N.

doc/manual/source/incremental-signing.rst
```diff
@@ -0,0 +1,105 @@
+Incremental Signing
+===================
+
+.. note:: This page documents how Cascade's incremental signer works, so that
```
ximon18: Two things here:
1. I don't think a note is needed here, this can just be normal text.
2. The text about jitter is confusing as it can be interpreted to mean that Cascade does not spread out signing over time but it does, just not based on jitter.

doc/manual/source/incremental-signing.rst
```diff
@@ -0,0 +1,105 @@
...
+   settings will affect it. 
+   
+   Note that Cascade does *not* use a jitter-based mechanism (as one would
+   find in OpenDNSSEC) for spreading out the re-signing of records. Read on to
+   understand Cascade's "re-signing window" mechanism.
+
+Context
+-------
+
+Cascade can sign your zone in two modes: full signing and incremental signing.
```
ximon18: I also find this confusing. To me it reads that as a user you need to more and care about these two modes and select which one you want to use which is not the case. Maybe say Cascade has two different modes of signing and chooses automatically which one needs to be used when.

Or don't talk about modes but instead say that when possible Cascade will sign only part of the zone leaving the other parts till later, instead of signing the whole zone at once, and that this is referred to as incremental signing.

doc/manual/source/incremental-signing.rst
```diff
@@ -0,0 +1,105 @@
...
+Full signing unconditionally signs every record in the loaded zone, even if
+some of those records had valid signatures. Incremental signing will examine
+previously generated signatures and only sign records that are missing
+signatures or whose signatures need updating. Incremental signing is more
+complicated than full signing, but it has two benefits:
+
+1. It greatly reduces the number of signatures that need to be generated. This
```
ximon18: Is this true? The same number of signatures always need to be generated, it's just a question of when and avoiding unnecessarily re-generating signatures.

doc/manual/source/incremental-signing.rst
```diff
@@ -0,0 +1,105 @@
...
+   is important for performance, especially if an :term:`HSM <Hardware Security
+   Module (HSM)>` is used.
+
+2. It greatly reduces the number of records that change in the signed zone.
+   With full signing, every change to the loaded zone would result in brand
```
ximon18:
```suggestion
   With full signing, every change to the loaded zone would result in
```

doc/manual/source/incremental-signing.rst
```diff
@@ -0,0 +1,105 @@
...
+   new signatures for *every* record. With incremental signing, small changes
+   in the loaded zone will translate to small changes in the signed zone. This
+   reduces the size of :term:`IXFRs <Incremental Zone Transfers (IXFRs)>`,
+   improving network latency and bandwidth for propagating the zone.
```
ximon18: I don't think you mean that Cascade is able to improve the latency and bandwidth of the network as Cascade has no control over such things. It is able to make better use of the connection making it less sensitive to the latency and bandwidth properties of the network.

doc/manual/source/incremental-signing.rst
```diff
@@ -0,0 +1,105 @@
...
+
+Cascade uses full signing for the first time it signs a zone, as it does not
```
ximon18, line 34: This is presumably not true for passthru with an already signed zone as Cascade **is** then aware of existing signatures.

doc/manual/source/incremental-signing.rst
```diff
@@ -0,0 +1,105 @@
...
+know of any previous signatures. After the initial signing operation, a signed
+copy of the zone will always be available, so Cascade will only use incremental
+signing. In the future, we may add support for forcing the use of the full
```
ximon18:
```suggestion
signing from that point on. In the future, we may add support for forcing the use of the full
```
However, I wonder, could policy change require complete re-signing of an already signed zone?

doc/manual/source/incremental-signing.rst
```diff
@@ -0,0 +1,105 @@
...
+signer even if a signed copy of the zone is available.
```
ximon18: I would delete the in future sentence.

doc/manual/source/incremental-signing.rst
```diff
@@ -0,0 +1,105 @@
...
+
+Key Rollovers
+-------------
+
+Cascade has a key manager, which can direct the signer to change the set of
```
ximon18, line 42:
```suggestion
The Cascade key manager can direct the signer to change the set of
```

doc/manual/source/incremental-signing.rst
```diff
@@ -0,0 +1,105 @@
...
+keys it signs the zone with. With full signing, this is relatively simple:
+all signatures will be re-generated using the new set of signing keys. But for
```
ximon18:
```suggestion
all signatures will be re-generated using the new set of signing keys. For
```
I was taught never start a sentence with "but" so I'm wondering if this correct, but honestly I have seen it done by others so am unsure.

doc/manual/source/incremental-signing.rst
```diff
@@ -0,0 +1,105 @@
...
+supports performing key rollovers over a period of time, where it will gradually
+update the signatures of all the records in the zone. Re-signing is performed
+such that the zone is always DNSSEC valid; for most key rollovers, this implies
+signing all records with an additional key, and once the new signatures have
+propagated, removing signatures with a previous key.
+
+The incremental signer follows the same procedure whether a key rollover is
+occurring or not. Internally, it will first determine which signatures need
+updating, and then calculate when they should be updated. If a key rollover
+is occurring, it will mark signatures signed with the previous set of keys to
```
ximon18:
```suggestion
is occurring, it will mark signatures signed with the previous set of keys to
```

Does it actually "mark" signatures? I assume instead it keeps track of them auto marking/modifying them in any way.

doc/manual/source/incremental-signing.rst
```diff
@@ -0,0 +1,105 @@
...
+First of all, all changes in the loaded zone are accounted for, from NSEC/NSEC3
+chain updates to generating new signatures.
+
+Then: every signature that needs re-generating (because it is sufficiently close
+to expiry) will be considered. Signatures that are *too* close to expiry will be
+re-generated immediately. The rest will be sorted into a *re-signing schedule*,
+which is a deterministic way to spread out incremental signing work. They will
+be sorted by ascending order of expiry, and then by DNSSEC canonical order. The
+first ``N`` such signatures will be re-generated, where ``N`` is chosen so that
+re-signing will complete within a certain period of time.
```
ximon18:
```suggestion
re-signing of the entire zone will complete within a certain period of time.
```

doc/manual/source/incremental-signing.rst
Philip-NLnetLabs: I think what is meant here is that always signing the full zone requires more signatures to be generated than only signing what changed (plus what is needed to avoid expiration and for key rolls).

doc/manual/source/incremental-signing.rst
Philip-NLnetLabs: I think what is meant here is that small IXFRs propagate faster and require less bandwidth than the bigger changes you would get from a full zone signer. Though this feels bit like a sale pitch.

doc/manual/source/incremental-signing.rst
Philip-NLnetLabs: We don't have pass-through at the moment. It is hard to say how it will work until we get it.

doc/manual/source/incremental-signing.rst
Philip-NLnetLabs: Technically, the answer is yes. If you change the policy to switch between NSEC and NSEC3 and set the key-roll-time to zero and trigger a key roll, then the entire zone will be signed. But that feels a bit contrived.

doc/manual/source/incremental-signing.rst
Philip-NLnetLabs: Indeed there is no marking.

doc/manual/source/incremental-signing.rst
bal-e: I'll change the text to mention that the mode is chosen automatically.

doc/manual/source/incremental-signing.rst
bal-e: I'll clarify this a bit.

doc/manual/source/incremental-signing.rst
bal-e: I'll just remove that note.

doc/manual/source/incremental-signing.rst
bal-e: We've previously discussed that the way Cascade works today often differs from how we'd like it to work ideally, and that it can make sense to mention both. But discussion about forcing full-re-signing is still ongoing, so I'll remove the sentence.

doc/manual/source/incremental-signing.rst
bal-e: Yep, we'll just have to update this page once we get to it.

doc/manual/source/incremental-signing.rst
bal-e: The sentence already starts with "After", I don't think we need to repeat ourselves.

doc/manual/source/incremental-signing.rst
bal-e: I think it's better to introduce the concept of the key manager being a dedicated component within Cascade explicitly.

doc/manual/source/incremental-signing.rst
bal-e: The behavior I was referring to over here is <https://github.com/NLnetLabs/cascade/blob/47d9b7cb8437a4d34f7955e7289c0768f1c2e930/src/signer/incremental.rs#L520-L524>. It's not quite explicit in the code, but the key tags are only being copied and checked to determine whether the RRset needs resigning. While the signatures are not literally marked, I think it's an appropriate simplification that correctly communicates the behavior for people unfamiliar with the code.

doc/manual/source/incremental-signing.rst
bal-e: Ah, right! I've updated the text.

doc/manual/source/incremental-signing.rst
bal-e: Right, I've rewritten the paragraph to make that more explicit.

doc/manual/source/incremental-signing.rst
bal-e: Ah, right! Fixing.

doc/manual/source/incremental-signing.rst
Philip-NLnetLabs: May use select instead of mark?

bal-e: That's perfectly fine with me, will do.

bal-e: I ended up rewriting the paragraph anyway, take a look.

Philip-NLnetLabs (approved):

PR #644, opened 2026-05-11
Ignore NOTIFY if the zone was sourced from a file.

ximon18:
As discussed internally.

---

### DIFF
--- a/src/units/zone_server.rs
+++ b/src/units/zone_server.rs
@@ -745,10 +745,36 @@ impl Notifiable for LoaderNotifier {
             // able to learn the used TSIG key.
             let center = &self.center;
             if let Some(zone) = crate::center::get_zone(center, apex_name) {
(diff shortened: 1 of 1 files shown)


src/units/zone_server.rs
```diff
@@ -745,10 +745,36 @@ impl Notifiable for LoaderNotifier {
...
+                    }
+
+                    crate::loader::Source::Zonefile { .. } => {
+                        warn!(
+                            "Ignoring NOTIFY for zone '{apex_name}': zone source is not an upstream nameserver"
+                        );
+                    }
+
+                    crate::loader::Source::None => {
+                        // Nothing to do.
```
tertsdiepraam: I wonder if a warning here would be useful too? Either way is fine though.

tertsdiepraam (approved):

src/units/zone_server.rs
ximon18: I added a warning.

PR #646, opened 2026-05-11
Recompute the resign timer when policy changes.

Philip-NLnetLabs:
Recompute the resign timer when policy changes.

---

### DIFF
--- a/src/units/zone_signer.rs
+++ b/src/units/zone_signer.rs
@@ -275,6 +275,13 @@ impl ZoneSigner {
         let _ = self.next_resign_time_tx.send(self.next_resign_time(center));
     }
 
+    pub fn on_zone_policy_changed(&self) {
+        // Just recompute the resign timer. In the future we may want to
+        // react to changes in policy, for example, whether NSEC is used
+        // or NSEC3.
+        let _ = self.next_resign_time_tx.send(Some(Instant::now()));
+    }
+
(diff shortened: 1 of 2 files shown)


src/units/zone_signer.rs
```diff
@@ -275,6 +275,13 @@ impl ZoneSigner {
         let _ = self.next_resign_time_tx.send(self.next_resign_time(center));
     }
 
+    pub fn on_zone_policy_changed(&self) {
+        // Just recompute the resign timer. In the future we may want to
+        // react to changes in policy, for example, whether NSEC is used
+        // or NSEC3.
+        let _ = self.next_resign_time_tx.send(Some(Instant::now()));
```
ximon18, line 282: When I read "recompute" I expected something more than "now". Is that correct?

ximon18: Would some policy changes require notifying `dnst keyset` of changes? If so, would that result in `dnst keyset` deciding re-signing needs to happen now, and thus is an alternative way to do this be to nudge the Cascade key manager to invoke `dnst keyset` to find out if re-signing is needed?

Philip-NLnetLabs: Yes, that is correct. Sending now lets the timer expire immediately. As a side effect the proper timeout gets computed.

That might work, but I doubt that there is a reliable way to get the key manager to trigger resigning.

ximon18 (approved):

PR #656, opened 2026-05-13
Signing policy check and signature refresh improvements.

Philip-NLnetLabs:
Consistency check for signature lifetime time, remain time and fresh interval. Improve signature refresh calculations.

---

### DIFF
--- a/src/policy/mod.rs
+++ b/src/policy/mod.rs
@@ -227,6 +227,79 @@ fn check_policy(policy: &PolicyVersion, tsig_store: &TsigStore) -> Result<(), Po
             .get(tsig_name)
             .ok_or(PolicyReloadError::NoSuchTsigKey(tsig_name.clone()))?;
     }
+
+    // Check signer policy.
+
+    // sig_validity_time
+    //
+    // The maximum sig_validity_time is determined by what we can put in
+    // the expiration time. Expiration time is effectively a 32-bit signed
+    // value. So sig_validity_time has to be less then 0x8000_0000. To
+    // give ourselves some headroom, set the limit to 0x4000_0000.
+    if policy.signer.sig_validity_time >= 0x4000_0000 {
+        return Err(PolicyReloadError::BadValue(format!(
+            "signature-lifetime {} too big (>= 0x4000_0000)",
+            policy.signer.sig_validity_time
+        )));
+    }
+
+    // The minimum value of sig_validity_time is bounded by sig_remain_time
+    // and signature_refresh_interval. We get to this later.
+
+    // sig_remain_time
+    //
+    // The effective lifetime of a signature is
+    // sig_validity_time - sig_remain_time. This needs to be greater than
+    // zero. So the maximum value of sig_remain_time is bounded by
+    // sig_validity_time. We will check this later.
+    //
+    // Ideally, sig_remain_time should be larger than the maximum TTL
+    // to make sure that old signature are removed from caches before
+    // they expire. We don't have a maximum TTL value. So what the signer
+    // does is add the TTL of an RRset to sig_remain_time to determine
+    // if a signature needs to be refreshed. For this reason, the lower bound
+    // of sig_remain_time is zero. However, this does not leave any margin
+    // for error.
+
+    // signature_refresh_interval
+    //
+    // The maximum is again bounded by sig_validity_time.
+    //
+    // Each signature_refresh_interval seconds, the signer will generate a
+    // new version of the zone with some refreshed signatures. For this reason,
+    // signature_refresh_interval should not be too low. Enforce a lower
+    // bound of 60 seconds to avoid accidentally generating new zone versions
+    // at a high rate.
+    if policy.signer.signature_refresh_interval < 60 {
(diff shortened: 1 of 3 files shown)


src/policy/mod.rs
```diff
@@ -227,6 +227,85 @@ fn check_policy(policy: &PolicyVersion, tsig_store: &TsigStore) -> Result<(), Po
...
+
+    // sig_validity_time
+    //
+    // The maximum sig_validity_time is determined by what we can put in
+    // the expiration time. Expiration time is effectively a 32-bit signed
+    // value. So sig_validity_time has to be less then 0x8000_0000. To
+    // give ourselves some headroom, set the limit to 0x4000_0000.
+    if policy.signer.sig_validity_time >= 0x4000_0000 {
+        return Err(PolicyReloadError::BadValue(format!(
+            "signature-lifetime {} too big (>= 0x4000_0000)",
```
ximon18, line 241: While technically correct I don't think it's very friendly to present the operator with a hex value as a limit.

Also, there's something I don't understand here. 0x4000_0000 is 1,073,741,824 seconds which is 17,895,697 minutes which is 298,621 hours which is 12,427 days. ~That's an odd and low number is it not?~ _Sorry, I was confused by Dutch locale using period where in English locale it would be a comma..._ That's a very high value, 34 years, why would we want to let people configure something that high, would it make sense to pick a lower round number as a limit such as 10 years or something?

And lastly, why don't I see `sig-validity-time` or `sig-remain-time` in the policy template?

src/policy/mod.rs
```diff
@@ -227,6 +227,85 @@ fn check_policy(policy: &PolicyVersion, tsig_store: &TsigStore) -> Result<(), Po
...
+
+    // sig_remain_time
+    //
+    // The effective lifetime of a signature is
+    // sig_validity_time - sig_remain_time. This needs to be greater than
+    // zero. So the maximum value of sig_remain_time is bounded by
+    // sig_validity_time. We will check this later.
+    //
+    // Ideally, sig_remain_time should be larger than the maximum TTL
+    // to make sure that old signature are removed from caches before
```
ximon18:
```suggestion
    // to make sure that old signatures are removed from caches before
```

src/policy/mod.rs
```diff
@@ -227,6 +227,85 @@ fn check_policy(policy: &PolicyVersion, tsig_store: &TsigStore) -> Result<(), Po
...
+    // signature_refresh_interval
+    //
+    // The maximum is again bounded by sig_validity_time.
+    //
+    // Each signature_refresh_interval seconds, the signer will generate a
+    // new version of the zone with some refreshed signatures. For this reason,
+    // signature_refresh_interval should not be too low. Enforce a lower
+    // bound of 60 seconds to avoid accidentally generating new zone versions
+    // at a high rate.
+    if policy.signer.signature_refresh_interval < 60 {
```
ximon18, line 273: I note this also prevents testing and demonstrating with values lower than this.

src/policy/mod.rs
```diff
@@ -227,6 +227,85 @@ fn check_policy(policy: &PolicyVersion, tsig_store: &TsigStore) -> Result<(), Po
...
+        return Err(PolicyReloadError::BadValue(format!(
+            "signature-refresh-interval {} too small (< 60)",
+            policy.signer.signature_refresh_interval
+        )));
+    }
+
+    // Check if everything fits together. The effective lifetime of a
+    // signature is sig_validity_time - sig_remain_time. This needs to be
+    // greater than zero. We need to take TTL into account. Assume a reasonable
+    // TTL of one hour (3600 seconds). This could cause signing a zone to
```
ximon18: How would this cause signing a zone to be aborted?

src/policy/mod.rs
```diff
@@ -227,6 +227,85 @@ fn check_policy(policy: &PolicyVersion, tsig_store: &TsigStore) -> Result<(), Po
...
+    // be aborted, but there is not much we can do. So now we have
+    // sig_validity_time - sig_remain_time - 3600 > 0.
+    // We sign every signature_refresh_interval so we need to take that into
+    // account as. Which gives:
```
ximon18:
```suggestion
    // account. Which gives:
```

ximon18 (approved):
Very nice code comments/documentation, thanks @Philip-NLnetLabs.

src/policy/mod.rs
Philip-NLnetLabs: Maybe people want signatures that don't expire. I don't see a reason to limit it to 10 years. You will only see this limit if you manage to input something bizar (or want to find out what the maximum is).

Philip-NLnetLabs: They are called signature-lifetime and signature-remain-time. Yes, it is confusing.

src/policy/mod.rs
Philip-NLnetLabs: Testing should be done using fake time. Maybe demonstrations as well.

src/policy/mod.rs
Philip-NLnetLabs: Ah, I need to get back to this. I now change to code to log an error and continue.

PR #657, opened 2026-05-14
Purging of excess diffs.

### LINKED ISSUE #825: Restoring with a rejected signed zone in the history causes loss of the published zone.
With Cascade v0.1.0-beta2:

```
$ cascaded --config ./cascade.conf --state /tmp/cascade/cascade.db &
$ cascade template policy > /tmp/cascade/policies/default.toml
$ vi /tmp/cascade/policies/default.toml  # set signer.review to "manual"
$ cascade policy reload
$ cascade zone add --source /tmp/example.zone --policy default example.com
$ cascade zone reject --signed example.com 2026061800
$ cascade zone reload example.com
$ cascade zone approve --signed example.com 2026061801
$ pkill cascaded

$ cascade --config ./cascade.conf --state /tmp/cascade/cascade.db &
$ dig +noall +answer +onesoa @127.0.0.1 -p 2005 AXFR arya.com
; Transfer failed

$ grep WARN /tmp/cascade.log
2026-06-18T11:19:19.187225Z  WARN stream{conn=2}: cascaded::server::service::compat: Returning SERVFAIL for AXFR request from client '127.0.0.1' for zone 'example.com': zone is empty
```

This is due to two things:

1. At signed review rejection the internal copy of the loaded zone was discarded but the persisted data for the loaded zone was not deleted and not forgotten from state. This leads to two copies of the same loaded zone with one being considered the snapshot and the other a diff against it, which it is not. On process restart the restore fails and both the persisted data and any in-memory zone contents are discarded.
2. We don't automatically do a reload of a zone, an operator would have to issue `cascade zone reload example.com` to resolve the situation.

ximon18:
Builds on #804.

# The problem

In #550 when a zone is first loaded and is approved it is persisted to disk as a loaded snapshot. Once it is signed and approved a signed snapshot is written to disk (as an AXFR payload in wire format). On process restart the snapshots are restored allowing the published zone to be served as it was before shutdown. Every time thereafter that the zone changes, either due to changes to the loaded zone via reloaded from disk or incoming XFR, or due to changes in the signed zone due to signing loaded record changes and/or (incremental) re-signing to account for DNSKEY changes and/or to avoid signature expiration, that results in loaded and/or signed diffs.

Those diffs are persisted to disk (as IXFR payloads in wire format) in order to be replayed on top of the zone snapshot after process restart, and stored in memory to be served as IXFR out diffs.

The persistence of diffs in #550 is unbounded, both in-memory and on-disk.

# Challenges to overcome

Excess IXFR in-memory diffs can just be forgotten, but excess on-disk diffs cannot just be deleted because they form part of the snapshot+diff history of the zone that is replayed on process restart to restore the published zone. Removing a diff would break restoration as the next available diff would not apply correctly to the snapshot.

# What this PR provides

This PR sets an upper limit on the number of diffs to keep, adds `PersistentDiffManager`, `Compacter` and `IxfrZoneDiffs`:

- `PersistentDiffManager` manages the set of paths to persistent diffs on disk. It tracks the `next_idx` to use to generate a unique filename for the next diff to write (this could perhaps be avoided by using UUIDs or something). It also keeps track of the `restore_base_idx` which indicates the offset into the set of persisted diffs at which replay during restore should restart.

- `Compacter` replaces the persisted snapshots for a zone with new versions containing records from the in-memory copy of the currently published zone, as this represents the existing snapshot with all existing diffs applied, and updates `restore_base_idx` accordingly. Writing out the entire zone to disk is a time consuming operation for a large zone and so we look for zones needing compaction once a minute. I'm not at all sure what strategy should be used here, this is just an initial attempt to avoid persisting to disk on every single change but instead to batch the changes together less frequently.

- `IxfrZoneDiffs` stores diffs for use by `Service` when responding to IXFR requests to the publication server. It also provides the logic to look up the diffs needed to answer a query, and "trimming" functionality to keep in-memory diff sizes within configurable limits.

# New policy settings

- `server.outbound.max-diffs`
- `server.outbound.max-diffs-size`

# Features NOT included in this PR

- There is no limit on bytes used in memory or on disk.
- There is no support for RFC 1995 purging of diffs based on total IXFR response size or SOA expire period, though there is something related which is purging based on the ratio of diff size to last published zone size.
- There is no support for RFC 1995 condensation of multiple diffs.

# Known issues

- Code quality: All manipulation of the filesystem for persistence zone snapshots/diffs should be done in one place but spread across multiple places. Partly this is because `PersistentState` and `PersistentDiffManager` have no reference to `Center` from which they can gain the `zone-state` path config value) nor can it be passed in via a constructor as the parent hierarchy uses `Default`. Likewise the persistent state relates to a single zone but has no concept of which zone that is as when it is created, as well as the Default problem mentioned, it is created as part of the zone so it can't take a reference to something which isn't constructed yet.

# Tasks remaining

- [x] What happens if compaction is still running when compaction is attempted again? The lock held on zone state prevents concurrent changes to the set of diff paths or the diffs themselves while compaction, persistence or restoration is in progress.
- [x] Fix #825. Ideally the fix would be backported but the code is different in earlier branches so copying the code unmodified not possible.
- [x] Manually test with incremental signing.
- [x] Add support for #830
- [ ] Add unit tests
- [x] Add architectural documentation (see `src/persistence/mod.rs`)
- [x] Do we maybe need some end user facing documentation about persistence e.g. that `zone status` may show the zone as restoring or that `policy show` prints the purging related policy settings, or something in the high level zone transfers page about limiting of diffs kept?

---


cascade.conf
tertsdiepraam, line 1: I guess this got committed by accident?

crates/cli/src/commands/policy.rs
```diff
@@ -286,13 +286,16 @@ fn print_server_policy(
...
+                max_diffs,
+                max_diffs_size,
             },
     }: &ServerPolicyInfo,
 ) {
     println!("  server:");
     println!("    outbound:");
     print_nameserver_comms_policy("provide XFR to", provide_xfr_to);
     print_nameserver_comms_policy("send NOTIFY to", send_notify_to);
+    println!("      max diffs: {max_diffs} ({max_diffs_size}%)");
```
tertsdiepraam: I'm not sure what the percentage here is meant to represent. I can read it later, but I find the representation here confusing.

doc/manual/source/man/cascaded-policy.toml.rst
```diff
@@ -597,6 +597,50 @@ The ``[server.outbound]`` section.
...
+   does not currently implement this behaviour. Cascade also currently has no
+   support for RFC 1995 section 6 condensation of multiple versions.
+   
+   The default is to limits diffs to 5 per zone. If set to 0, storage of diffs
+   will be disabled.
+
+.. option:: max-diffs-size = 20
+
+   The maximum size allowed for in-memory diffs for a single zone, defined as
+   a parcentage of the size of the last published version of the zone. If
```
tertsdiepraam:
```suggestion
   a percentage of the size of the last published version of the zone. If
```

doc/manual/source/man/cascaded-policy.toml.rst
```diff
@@ -597,6 +597,50 @@ The ``[server.outbound]`` section.
...
+   storing a new diff would exceed the limit, diffs will be discarded to make
+   space starting with the oldest diffs first.
+
+   This setting has no impact on the number of diffs persisted to disk. See
+   max-diffs to control that.
+
+   The default is 20%. If set to 0, or a value low enough such that the set of
+   diffs for a zone always exceed the limit, then no diffs will be stored in
+   memory and IXFR requests will be responded to with an AXFR instead.
```
tertsdiepraam, line 646: Maybe suggest using `max-diffs` for that use-case?

doc/manual/source/zone-transfers.rst
```diff
@@ -42,25 +42,70 @@ Cascade can be instructed to authenticate the upstream nameserver by use of a
...
+
+Diffs are persisted to disk so that if Cascade is restarted that it is still
+able to respond to IXFR requests from downstream nameservers rather than
+forcing them to costly fallback to AXFR transfers.
+
+Cascade has two policy settings which limit the amount of memory and disk
+space used per zone to store diffs:
+
+- ``server.outbound.max_diffs``
+- ``server.outbound.max_diffs_size``
```
tertsdiepraam: I like to add a toml example for these kinds of things because the conversion from this description to toml can sometimes be confusing for people who don't know toml that well.

```toml
[server.outbound]
max-diffs = 5
max-diffs-size = 20
```

doc/manual/source/zone-transfers.rst
```diff
@@ -42,25 +42,70 @@ Cascade can be instructed to authenticate the upstream nameserver by use of a
...
+
+``max_diffs`` specifies the maximum numer of IXFR diffs per zone that Cascade
+may keep in memory and on disk.
+
+``max_diffs_size`` specifies the maximum size of all diffs combined that may
+be stored in memory per zone and is defined in relation to the size (number of
+records) of the published zone.
```
tertsdiepraam: "in relation to the size" is a bit ambiguous.

doc/manual/source/zone-transfers.rst
```diff
@@ -42,25 +42,70 @@ Cascade can be instructed to authenticate the upstream nameserver by use of a
...
+forcing them to costly fallback to AXFR transfers.
+
+Cascade has two policy settings which limit the amount of memory and disk
+space used per zone to store diffs:
+
+- ``server.outbound.max_diffs``
+- ``server.outbound.max_diffs_size``
+
+``max_diffs`` specifies the maximum numer of IXFR diffs per zone that Cascade
+may keep in memory and on disk.
```
tertsdiepraam, line 95: You're using underscores here but hyphen in the policy template.

etc/policy.template.toml
```diff
@@ -478,3 +478,42 @@ mode = "off"
...
+# maintenance mode.
+# 
+# Note that while RFC 1955 section 5 suggests that diffs be purged based on
+# total IXFR response size and/or if older than the SOA expire period, Cascade
+# does not currently implement this behaviour. Cascade also currently has no
+# support for RFC 1995 section 6 condensation of multiple versions.
+#max-diffs = 5
+
+# The maximum size allowed for in-memory diffs for a single zone, defined as a
+# parcentage of the size of the last published version of the zone. If storing
```
tertsdiepraam:
```suggestion
# percentage of the size of the last published version of the zone. If storing
```

src/persistence/zone.rs
```diff
@@ -271,18 +266,450 @@ pub struct PersistenceState {
...
+        trace!(
+            "Checking if compaction is needed for zone '{}': {num_signed_diffs} > {max_diffs}",
+            zone.name
+        );
+        if num_signed_diffs > max_diffs {
+            debug!(
+                "Compacting persisted diffs for zone '{}' with {} diffs > {} max diffs",
+                zone.name, num_signed_diffs, max_diffs
+            );
+            let num_diffs_to_remove = num_signed_diffs.abs_diff(max_diffs);
```
tertsdiepraam: Why `abs_diff` if you know that `num_signed_diffs > num_diffs`.

If you want to be extra safe regarding panics you might be able to do this:

```rust
if let Some(x) = num_signed_diffs.checked_sub(max_diffs) && x > 0 {
    // ...
}
```
But I think using `-` here is fine.

More personal preference: to quickly see that there is no `else` branch I'd prefer

```rust
if num_signed_difs <= max_diffs {
    return;
}
```
or `let-else` instead of `if-let` with my other suggestion.

src/persistence/zone.rs
```diff
@@ -271,18 +266,450 @@ pub struct PersistenceState {
...
+                                    "Failed to remove persisted zone data file '{}' while compacting zone '{}': {err}",
+                                    zone.name, diff_info.path.display()
+                                );
+                            }
+                        }
+                        keep
+                    });
+
+                // Remove the corresponding loaded diffs.
+                for loaded_serial in loaded_serials_to_remove.into_iter() {
```
tertsdiepraam: I think `into_iter` can be removed.

src/persistence/zone.rs
```diff
@@ -271,18 +266,450 @@ pub struct PersistenceState {
...
+                zone.name, num_signed_diffs, max_diffs
+            );
+            let num_diffs_to_remove = num_signed_diffs.abs_diff(max_diffs);
+            let loaded_snapshot_path = &loaded_snapshot_path.unwrap().path;
+            let signed_snapshot_path = &signed_snapshot_path.unwrap().path;
+
+            // Get access to the published records for the zone, so that we
+            // can write new loaded and signed snapshot files to disk.
+            if let Ok(viewer) = viewer.try_read()
+                && let Some(reader) = viewer.read()
```
tertsdiepraam: Here I'd also prefer `let Ok(viewer) = viewer.try_read() else { return; }`. But feel free to leave it like this.

src/persistence/zone.rs
```diff
@@ -271,18 +266,450 @@ pub struct PersistenceState {
...
+                    None,
+                    reader.soa().clone(),
+                    [].iter(),
+                    reader.generated_records().iter(),
+                );
+
+                // Now that we have re-written the snapshots using the latest
+                // published version of the zone we don't need any of the
+                // on-disk persisted diffs that were previously applied on top
+                // of the old snapshot to re-create the zone.
```
tertsdiepraam: What happens if Cascade is quit or crashes here? Maybe we don't care, I'm just wondering.

src/persistence/zone.rs
```diff
@@ -271,18 +266,450 @@ pub struct PersistenceState {
...
+                let mut idx = 0;
+                let mut loaded_serials_to_remove = vec![];
+                state
+                    .persistence
+                    .signed_diffs
+                    .diff_infos
+                    .retain(|diff_info| {
+                        // Keep only the snapshot and diffs newer than the ones
+                        // to remove.
+                        let keep = idx == 0 || idx > num_diffs_to_remove;
```
tertsdiepraam, line 432: Hmmm, so it's the snapshot as the first and then old to new in the `BTreeSet`. I guess ideally the snapshot would be a separate field, but this works.

I think you can make this a bit more intuitive with `drain` so you don't need the keep variable:

```rust
for diff_info in state
    .persistence
    .signed_diffs
    .diff_infos
    .drain(1..=num_diffs_to_remove) {
    // ...
}
```

src/persistence/zone.rs
```diff
@@ -271,18 +266,450 @@ pub struct PersistenceState {
...
+            .zone_state_dir
+            .join(format!("{zone_name}.{data_file_type}.{}", self.next_idx))
+            .into_std_path_buf();
+        let file_info = PersistedDiffFileInfo::new(path.clone(), loaded_serial, signed_serial);
+
+        assert!(self.diff_infos.insert(file_info));
+
+        // replace with strict_add() once our MSRV reaches 1.91.0.
+        assert_ne!(self.next_idx, usize::MAX);
+        self.next_idx += 1;
```
tertsdiepraam: or  `self.next_idx.checked_add(1).unwrap()` but this is fine too.

src/persistence/zone.rs
```diff
@@ -271,18 +266,450 @@ pub struct PersistenceState {
...
+pub enum PersistedDiffRecordSource {
+    Loaded,
+    Signed,
+}
+
+//----------- PersistedDiffManager -------------------------------------------
+
+/// Metadata about a related collection of persisted zone data files.
+#[derive(Clone, Debug)]
+pub struct PersistedDiffManager {
```
tertsdiepraam, line 567: I like this struct being responsible for the file management!

src/persistence/zone.rs
```diff
@@ -271,18 +266,450 @@ pub struct PersistenceState {
...
+    next_idx: usize,
+
+    /// The index of the first diff_info to apply to the snapshot when restoring.
+    ///
+    /// After compaction the on-disk diffs that existed must no longer be applied
+    /// to the base snapshot as the new snapshot includes them, but we should
+    /// still track their paths so that we can load them for use in responding to
+    /// IXFR client requests. So we need to remember which index to start applying
+    /// diffs to the snapshot from.
+    restore_base_idx: usize,
```
tertsdiepraam: I think this struct could use a couple of improvements. It took me a while to figure out that there are diffs "before" and "after" the snapshot. It's quite confusing that the snapshot is the first "diff" even though it was created after some of the diffs.

The `idx` also has several different meanings now: the index in the filename and the index into the btreeset. This might be improved by calling one the "id" and the other the "idx". However, I'd also add an `id` field to each `PersistedDiffFileInfo` to make make some operations easier.

If I'd design it, it would probably look something like this:

```rust
pub struct PersistedDiffManager {
    record_source: PersistedRecordSource,
    snapshot: PathBuf, // or a similar type
    before: Vec<PersistedDiffFileInfo>,
    after: Vec<PersistedDiffFileInfo>,
}

pub struct PersistedDiffFileInfo {
    id: usize,
    path: PathBuf,
    loaded_serial: Option<Serial>,
    signed_serial: Option<Serial>,
}
```

You then don't need `next_id` because that's just the next id after the last. It's also more clear that the snapshot is treated differently from the diffs.

If you don't want to make those changes then we at least need some documentation somewhere on how the whole system works. For instance, I don't think this matches the description you currently give in `src/persistence/mod.rs`.

Another alternative is doing compaction differently without the data from the publication server, just doing actual compaction from the diffs in the files. Then we don't need diffs before the snapshot at all.

src/persistence/persist.rs
```diff
@@ -164,93 +138,62 @@ pub fn persist_signed(
...
+        destination,
+        diff.removed_soa.clone(),
+        diff.added_soa.clone().unwrap(),
+        diff.removed_records.iter(),
+        diff.added_records.iter(),
+    );
+}
+
+// TODO: It would be nice to take the records by reference.
+pub fn persist_to_file_from_parts<
```
tertsdiepraam, line 164: I thought at first that "parts" meant "diffs". I think this could use maybe another name. Maybe:

- `persist_to_file` -> `persist_diff_to_file`
- `persist_to_file_from_parts` -> `persist_to_file`

Just an idea, maybe you have a better solution.

src/persistence/persist.rs
```diff
@@ -303,43 +244,162 @@ fn persist_to_file(destination: &Path, diff: Arc<DiffData>) {
...
+    let diffs = &mut zone.write(center).storage.diffs;
+    if let Some(loaded_diff) = loaded_diff {
+        diffs.store_loaded_diff(loaded_diff.clone());
+    }
+    diffs.store_signed_diff(loaded_serial, signed_diff.clone());
+}
+
+//------------ discard_excess_diffs() ----------------------------------------
+
+pub fn discard_excess_diffs(center: &Arc<Center>, zone: &Arc<Zone>) {
```
tertsdiepraam: This function (and some others) doesn't seem to impact persistence, so maybe it's in the wrong place? The name currently reads to me as "remove on disk diffs".

tertsdiepraam (changes requested):
There are some typos and small code suggestions but mainly I'd change how the the `PersistedDiffManager` is structured, because I think there's some structure there that's not really reflected in the fields that it has. It took me a while to figure out that the code is actually handling that correctly.

I think the mechanism is good and implemented pretty nicely given the complexity of the task, but I'd like to see it improved in that regard.

crates/cli/src/commands/policy.rs
ximon18: Would this be better?
```suggestion
    println!("      max diffs: {max_diffs} totaling less than {max_diffs_size}% of the published record count");
```

doc/manual/source/man/cascaded-policy.toml.rst
ximon18: I clarified the docs to note that `max-diffs` and `max-diffs-size` are not separate limits, both are always applied. Does that influence your suggestion?

doc/manual/source/zone-transfers.rst
ximon18: Isn't that already provided by the commented out default values?

doc/manual/source/zone-transfers.rst
ximon18: I clarified this. Please take a look at the changes.

src/persistence/zone.rs
ximon18: That is indeed stupid ;-) Thanks for spotting it. I have also applied your personal preference suggestion.

src/persistence/zone.rs
ximon18: Done.

src/persistence/zone.rs
ximon18: It's a good point. The steps the algorithm follows are:

1. Atomically write the updated loaded snapshot.
2. Atomically write the updated signed snapshot.
3. Remove no longer needed signed diff paths from state in memory and the files on disk, one at a time.
4. Remove no longer needed corresponding loaded diff paths from state in memory and the files on disk. one at a time.
5. Update the restore base idx values in state in memory, so that during restore any diffs now included in the snapshot are not applied on top of it.
6. Drop the state lock causing updated state in memory to be written to disk after 5 seconds.

If cascade is terminated anywhere in those steps the best case is that on restore a diff that should be skipped would be applied or may fail to load if deleted, which would cause all zone data to be discarded until next reload, the last published version of the zone would no longer be served and related IXFR diffs would be lost. Downstream secondaries would be unable to fetch the zone until next reload, sign and publication.

Worst case, some diffs that should not be applied may apply without error (e.g. a delete of a non-existing record would fail, but addition of a non-existing record could succeed) leading to corruption.

Since there are multiple steps involved, parts can be made atomic but not all, not as it is. One solution could be to keep persistent zone record state files in a subdirectory per zone and during compaction first prepare the new directory, then update state in-memory and on disk to refer to it, then delete the old state. This would use more disk space as it would require duplicating diff files temporarily.

src/persistence/zone.rs
ximon18: I think of it as the snapshot being just the first diff (against nothing) so its diffs stored old to new.

Hmm, I tried to use `drain()` but couldn't work out how to do it, thanks, let me try that.

ximon18: Hmm:

```
no method named `drain` found for struct `BTreeSet<T, A>` in the current scope
method not found in `BTreeSet<PersistedDiffFileInfo>`
```

Maybe that is why I didn't use drain.

src/persistence/zone.rs
ximon18: Thanks, I like that.

src/persistence/zone.rs
ximon18: > I think this struct could use a couple of improvements. It took me a while to figure out that there are diffs "before" and "after" the snapshot. It's quite confusing that the snapshot is the first "diff" even though it was created after some of the diffs.

There is always a single snapshot, then a sequence of diffs. Pre-compaction none of the diffs are present in the snapshot and some of them may be used for IXFR. Post-compaction some of the diffs may be present in the snapshot and so are loaded during restore for IXFR purposes but not applied to the snapshot. In both cases all diffs are loaded during restore and the diffs form a contiguous sequence. To me at least it's not confusing and I don't think of it as "before" and "after", but clearly it is not as obvious as I expected.

> The idx also has several different meanings now: the index in the filename and the index into the btreeset. 

It's true that there are two fields called idx and they are used differently and should be named differently, I just wasn't being very creative and lacked inspiration for a better name. For the `next_idx` I also considered whether maybe we should just use a UUID or something as it's only a means of ensuring the files have unique names and using a UUID would get rid of the need for the field entirely. It would however make the files on disk less obviously a sequence, one would have to use the persisted JSON zone state fields to know which order the files belong in as a sequence.

> You then don't need next_id because that's just the next id after the last. It's also more clear that the snapshot is treated differently from the diffs.

I'm not sure about this idea. The `id` in `PersistdDiffFileInfo` serves no purpose in every instance of the struct except the last one, which I find confusing as it is present but unused in the other instances.

> Another alternative is doing compaction differently without the data from the publication server, just doing actual compaction from the diffs in the files. Then we don't need diffs before the snapshot at all.

To compact without the publication server one would need to re-load the snapshot on disk and apply the diffs to it. That could be done but would require a second copy of the zone to be constructed in memory. You would still need the "before" diffs (as you call them) as they may still be needed in order to respond to IXFR requests from clients with an older version of the zone.

> If I'd design it, it would probably look something like this:

I considered doing that but found the overhead of the additional `Vec` undesirable given that they form a sequence and we only need an offset into it. 

I also had `Vec` instead of `BTreeSet` before but found the indexing into the vec error prone when referencing the appropriate loaded diff that relates to a given signed diff and preferred the explicit lookup by key into the `BTreeSet`. A vec also doesn't prevent the same entry being present twice by accident.

> If you don't want to make those changes then we at least need some documentation somewhere on how the whole system works. For instance, I don't think this matches the description you currently give in src/persistence/mod.rs.

I'm open to changes. I'm not convinced or fond of your proposed changes ;-) but I am only one maintainer of this code and if others prefer a different structure/naming/algorithm I will of course change it to what suits the majority of maintainers best.

src/persistence/persist.rs
ximon18: The naming was inspired by `Vec::from_parts()` in the Rust standard library.

The names are definitely lacking (as is RustDoc on the functions!) as they don't reference the wire XFR format used or as you note their intended use with a "diff".

Another approach could be to add a `Persist` trait with a `persist(&self, destination: &Path)` function and impl the trait for `DiffData`, then have it call the actual helper function, and call the helper function something like `fn persist_diff_records()` or such?

src/persistence/persist.rs
ximon18: It is called in `persistence.rs` by `persist_signed()` -> `store_for_ixfr_out()`. It has no impact on files on disk, it causes IXFR in-memory diffs to be discarded. Ideally `IxfrZoneDiffs::trim()` would do all of the logic itself but it has no access to `Center` or `Zone`. This wrapper function only exists to execute logic that requires the missing context before delegating to `trim()` to do the actual work.

crates/cli/src/commands/policy.rs
tertsdiepraam: I think that's better yes!

doc/manual/source/man/cascaded-policy.toml.rst
tertsdiepraam: I was thinking more that `max-diffs = 0` would make more sense than `max-diff-size = 0`, though I didn't quite understand that `max-diffs-size` only applies to in-memory.

doc/manual/source/zone-transfers.rst
tertsdiepraam: In the template you mean? Maybe yeah, though I think it could also fit here.

src/persistence/zone.rs
tertsdiepraam: Oh right! A `VecDeque` could work for this use-case but otherwise I guess this is fine.

src/persistence/persist.rs
tertsdiepraam: Alright!

src/persistence/zone.rs
tertsdiepraam: I understand your concerns. It might just need some more docs (I haven't checked your changes yet, maybe that's what you've done).

I think you should do what you feel is best. If you want we could brainstorm together on Thursday.

src/persistence/persist.rs
tertsdiepraam: I think the `from_parts` naming works best when constructing something, like with `Vec::from_parts`. A trait feels a bit heavy for this small naming problem. I'd rather have the current names I think.

doc/manual/source/zone-transfers.rst
tertsdiepraam: Ah sorry I wasn't clear enough. I like the clarification of "number of records", but I think it should also use the word "percentage".

src/persistence/mod.rs
tertsdiepraam: I think this needs a clear mention of "each diff newer than the snapshot".

I'm also now thinking that a lot of my confusion comes from the snapshot being `{zone_name}.loaded.0`. That implies it coming before the rest. Treating that separately from the other diffs might be nice, without changing any of the other data structures.

src/persistence/zone.rs
```diff
@@ -301,35 +296,463 @@ fn clear_persisted_zone_data(center: &Center, state: &mut ZoneState) {
...
+
+    /// The collection of persisted data file paths in this set.
+    diff_infos: BTreeSet<PersistedDiffFileInfo>,
```
tertsdiepraam, line 596: Maybe mention here that the snapshot is the first.

You could also try making this a `VecDeque`, because it feels like a very `VecDeque`-shaped problem, but I do see some of the reasons for `BTreeSet`.

src/persistence/zone.rs
```diff
@@ -301,35 +296,463 @@ fn clear_persisted_zone_data(center: &Center, state: &mut ZoneState) {
...
+                            "Failed to remove persisted zone data file '{}' while compacting zone '{}': {err}",
+                            zone.name,
+                            found_item.path.display()
+                        );
+                    }
+                }
+            }
+
+            state.persistence.loaded_diffs.restore_base_idx = state.persistence.loaded_diffs.len();
+            state.persistence.signed_diffs.restore_base_idx = state.persistence.signed_diffs.len();
```
tertsdiepraam: Maybe a comment here explaining what this is doing? That would help correlate the persistence with the restoration I think.

tertsdiepraam (commented):
Some comment suggestions to help explain how the restoration works.

src/persistence/zone.rs
ximon18: I'm resistant to anything accessed by index (i.e. a VecDeque as opposed to a BTreeMap) as I had hard to find bugs in earlier attempts when working that way.

tertsdiepraam: But how much are you accessing by index? Isn't it mostly pushing to the back and popping from the front?

doc/manual/source/man/cascaded-policy.toml.rst
ximon18: I reworked the text describing `max-diffs` and `max-diffs-size` a bit, please take a look at the changes.

doc/manual/source/zone-transfers.rst
ximon18: Ah, I see!

ximon18: I updated `zone-transfers.rst`, please take a look.

src/persistence/zone.rs
ximon18: Loaded diff has to be looked up by index to find the one that relates to a particular signed diff. I'm not willing at this point to change this from a BTreeSet to a Vec-like type as I already had trouble with that approach but I can add a TODO that with care this could be done with a Vec-like type instead.

src/persistence/zone.rs
ximon18: I added a TODO note for now about this.

doc/manual/source/zone-transfers.rst
ximon18: The text now includes the word "percentage".

src/persistence/zone.rs
ximon18: > Loaded diff has to be looked up by index to find the one that relates to a particular signed diff. I'm not willing at this point to change this from a BTreeSet to a Vec-like type as I already had trouble with that approach but I can add a TODO that with care this could be done with a Vec-like type instead.

Looking at the current code this doesn't justify the use of `BTreeSet` as the key currently in use is a path, to find an item by serial number the set has to be searched which would be the same for a Vec-like type. I was forgetting that earlier versions of the code did no such search but instead assumed an index based lookup could be done. It was the logic to calculate the index that was not easy to read and prone to mistakes (which could be solved through a better implementation). Looking up by serial was not done yet at that point as the serial relationships were not persisted in state in that version of the code.

ximon18: I tried briefly switching to a Vec-like data type but realized I like having a type that ensures it is a set. I also tried splitting out the initial diff to a separate snapshot field but found it made the code cumbersome, or at least it's not quick to follow up on all the changes it implies.

src/persistence/zone.rs
ximon18: I tried re-working this several times this morning and at other moments but the changes ripple out such that the work would take longer than I have available at present to complete. Instead I added some architectural TODO notes to the persistent module RustDocs to capture areas that could be improved later.

src/persistence/persist.rs
ximon18: I renamed `from_parts()` to `for_existing_diffs()`.

src/persistence/mod.rs
ximon18: I made some edits to the text there, please take a look.

src/persistence/zone.rs
ximon18: I made some edits to the RustDocs for PersistedDiffManager fields. Please take a look.

src/persistence/zone.rs
ximon18: Done.

etc/policy.template.toml
```diff
@@ -480,40 +480,46 @@ mode = "off"
...
-# occurs. Discarding of persisted diffs will be skipped for zones that are in
-# maintenance mode.
+# Diffs persisted to disk will also be deleted if above the specified
+# max-diffs limit, but not immediately, and if not needed to restore the
+# zone on Cascade restart. Deletion of diffs on disk (so-called 'compaction')
+# requires re-writing the entire zone to disk which for large zones could take
+# time and for rapidly changing zones is thus not desirable to do on every
+# change to the zone, and therefore is done periodically rather than on every
+# change. Diffs that exceed the max-diffs limit but which are still required
+# to restore the zone (because compactiion has not occurred yet) will remain
```
tertsdiepraam:
```suggestion
# to restore the zone (because compaction has not occurred yet) will remain
```

src/persistence/zone.rs
```diff
@@ -612,7 +612,7 @@ impl PersistedDiffManager {
         let file_info = PersistedDiffFileInfo::new(path.clone(), loaded_serial, signed_serial);
 
         assert!(self.diff_infos.insert(file_info));
-        self.next_idx.checked_add(1).unwrap();
+        self.next_idx = self.next_idx.checked_add(1).unwrap();
```
tertsdiepraam: Oh right! That's an interesting mistake. Totally understandable.

src/persistence/zone.rs
```diff
@@ -559,9 +572,19 @@ pub struct PersistedDiffManager {
...
+    ///
+    /// The first entry always refers to a persisted snapshot of the zone records
+    /// at a point in time.
+    ///
+    /// Subsequent entries refer to persisted diffs that should be applied in
+    /// sequence to the snapshot upon restore. However, if the [`Compacter`] has
+    /// replaced the snapshot content such that it includes some of the persisted
+    /// diffs, entries [1..restore_base_idx] may only be needed to respond to IXFR
+    /// requests (see [`Self::gete_diffs()`] and [`Service::ixfr()`]) and must not
+    /// be used for restoration.
```
tertsdiepraam: Nice!

tertsdiepraam (approved):

PR #660, opened 2026-05-18
Improve zone history for incremental signing.

### LINKED ISSUE #642: zone history is wrong for incremental signing.
I see the following entry in zone history:
2026-05-11T08:28:46Z -          Signing succeeded (triggered by loading a new instance)

The SOA serial is missing and the signing was not triggered by loading a new instance.

Philip-NLnetLabs:
Add serial, improve message. Remove entry about loading.

Fixes #642 

---

### DIFF
--- a/src/zone/machine.rs
+++ b/src/zone/machine.rs
@@ -157,6 +157,9 @@ impl<'a> ZoneHandle<'a> {
 
         transition.move_to(ZoneStateMachine::Signing(waiting.start_resign()));
 
+        self.state
+            .record_event(HistoricalEvent::StartedResign, None);
+
         Some(builder)
     }
 }
--- a/src/zone/machine.rs
+++ b/src/zone/machine.rs
@@ -3,7 +3,6 @@ use cascade_zonedata::{LoadedZoneBuilder, LoadedZoneBuilt, SignedZoneBuilder};
 use tracing::{info, trace};
 
 use crate::{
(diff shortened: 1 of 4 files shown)


src/zone/machine.rs
```diff
@@ -288,15 +288,6 @@ impl<'a> ZoneHandle<'a> {
     }
 
     pub(crate) fn finish_signing(&mut self, built: cascade_zonedata::SignedZoneBuilt) {
-        self.state.record_event(
```
ximon18, line 291: Why is this now logged in `sign_incrementally()`? Shouldn't it also be logged for full zone signing? Why isn't `finish_signing()` logically the right place to record that signing succeeded?

ximon18 (commented):
Including the serial in the history event and changing the wording is fine, I'm just unsure about the change of location in the code where signing succeeded is now logged.

src/zone/machine.rs
Philip-NLnetLabs: It already gets logged by the full zone signer. Without this PR there are two log entries when the full zone signer is used. finish_signing does not have easy access to the trigger of the SOA serrial.

ximon18: Ah I didn't realize it got logged somewhere by the full zone signer.

ximon18 (approved):

PR #671, opened 2026-05-22
Disable signing for non-apex DNSKEY/CDS/CDNSKEY in the incremental signer.

Philip-NLnetLabs:
Disable signing for non-apex DNSKEY/CDS/CDNSKEY in the incremental signer. This avoid confusion between the incremental signer and the full zone signer. On the whole this is not good because it breaks RFC 9615 (Automatic DNSSEC Bootstrapping Using Authenticated Signals from the Zone's Operator).

---


ximon18 (approved):
This change is simple enough but I wouldn't know how to test it or why the full zone signer is "incompatible" with the incremental signer as noted by the new comment in the code.

PR #674, opened 2026-05-27
Restore policies before zones

### LINKED ISSUE #673: Global policy not present when zones are restored.
After
```
cargo run --bin cascade zone add --source example.com.in --policy default example.com
```

stop cascaded and start cascaded again.

There is a log message:
```
2026-05-26T14:55:36.974439Z  WARN restore{name=example.com}: cascaded::zone::state: The policy 'default' used by the zone is not present in the global state; restoring it into the global state
```

This is caused by code in main.rs
```
            // Restore pending zones.
            for name in zones {
                assert!(
                    !state.zones.contains(&name),
                    "Zone '{name}' was encountered twice"
                );
                let zone =
                    match Zone::restore(&config, name, &mut state.policies, &state.tsig_store) {
                        Ok(zone) => zone,
                        Err(_) => return ExitCode::FAILURE,
                    };
                state.zones.insert(ZoneByName(Arc::new(zone)));
            }

            // Restore pending policies.
            state
                .policies
                .extend(policies.into_iter().map(|(name, spec)| {
                    let policy = spec.parse(&name);
                    (name, policy)
                }));

```
zones get restored before policies, but in the restore of  a zone there is a check that the policy exists in global state.

bal-e:
Resolves #673.

---

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -101,6 +101,14 @@ fn main() -> ExitCode {
                 }
             }
 
+            // Restore pending policies.
+            state
+                .policies
+                .extend(policies.into_iter().map(|(name, spec)| {
+                    let policy = spec.parse(&name);
+                    (name, policy)
+                }));
+
             // Restore pending zones.
             for name in zones {
                 assert!(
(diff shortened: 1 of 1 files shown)


Philip-NLnetLabs (approved):
LGTM

PR #678, opened 2026-05-27
Retain a new loaded instance upon signed review failure

bal-e:
If a signed instance is soft-rejected during review, we will retain any newly loaded instance it accompanies and try signing that instance again. If three consecutive signing attempts fail, we abandon the newly-loaded instance and enqueue a refresh in the zone loader.

Partly works around a bug in the zone signer, caused by rejecting the first signed instance generated by Cascade for a newly-loaded zone. We still need to fix this properly.

Affects the zone state machine pipeline: `SignedReview` used to move to `Waiting` on soft rejection, now it can move to `Waiting` or `Signing` depending on the retry counter.

---


src/zone/storage.rs
```diff
@@ -713,6 +769,56 @@ impl StorageZoneHandle<'_> {
...
+                    transition.move_to(ZoneDataStorage::Signing(s));
+                    builder
+                }
+
+                _ => unreachable!(
+                    "'ZoneDataStorage::CleaningSigned' is the only state where a 'SignedZoneCleaner' is available"
+                ),
+            };
+
+            // TODO: Re-start signing.
```
tertsdiepraam: Isn't this what the line below it is doing?

tertsdiepraam (commented):
I suppose this is gonna need some docs. Am I correct that this counter doesn't do anything for halted? That makes sense. Looks good though!

What was the bug we were working around here?

src/zone/storage.rs
bal-e: Oh yeah, I'll remove the comment.

bal-e:
> I suppose this is gonna need some docs.

You mean in the man pages / ReadTheDocs? AFAICT we don't have a single good place to document this right now. Maybe it should be part of a RtD page about the pipeline / state machine.

bal-e:
The `review-signed-zone` integration test is failing, but we have internally discussed a solution to that. However, this PR is related, so I don't want to ignore the failure. ~Let me make a new PR and explicitly block this on it.~ Blocked on #795.

bal-e:
After internal discussion, we've realized we're not sure whether this behavior is warranted. Closing until we have more experience with signing failures and signed review failures.

PR #681, opened 2026-05-29
The integration test used as a live demo script at the OARC 46 event.

ximon18:
This PR contains the integration test script that was used to run the live demo at [OARC 46](https://indico.dns-oarc.net/event/56/contributions/1213/).

I'm not sure if we should keep it or not. Opinions welcome.

### DIFF
--- a/integration-tests/tests/oarc46-demo/README.md
+++ b/integration-tests/tests/oarc46-demo/README.md
@@ -1,0 +1,5 @@
+The integration test in this directory was executed as a live demo at OARC 46
+in Edinburgh on May 17 2026.
+
+See https://indico.dns-oarc.net/event/56/contributions/1213/ for more
+information.
--- a/integration-tests/tests/oarc46-demo/action.yml
+++ b/integration-tests/tests/oarc46-demo/action.yml
@@ -1,0 +1,192 @@
+
+# Making reusable composite actions documented at
+# https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action#creating-a-composite-action-within-the-same-repository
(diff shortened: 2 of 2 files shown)


tertsdiepraam (commented):
> I'm not sure if we should keep it or not. Opinions welcome.

I'm not sure either. Maybe it makes sense as a separate bash script that you can run that we execute as part of the tests? As a pure test, it would need to do something that the other tests don't test.

Another question I have is how this works with `wait_for_key_press`. Does it wait for that?

I'm not approving because I still can't run the systems tests, so I'm leaving that to someone else.

ximon18:
Hi @tertsdiepraam 

This should never be included as part of the normal test suite or run by anybody, it is purely a historical record relating to OARC 46.

The integration test framework was used to run the demo because it sets up two NSD instances and Cascade in a way that is easy to repeat and required no manual local setup.

The wait for key press during the actual demo worked via coordination with my phone via HTTP requests to an intermediate web service running on my work machine, with the state of the service also being updated via HTTP request by a background loop running inside the demo script. I then used a button on my phone during the demo to make a HTTP request to the web service to advance the demo at will. The script still has the HTTP request code but commented out and instead would now progress automatically via timed sleeps as the service is no longer running on my work machine.

One can run the demo script if desired, but I'm not sure why you would. I just wondered if for viewers of the live (or later when published the recorded) demo would be interested to see the actual commands the demo used, reading through it at their own pace.

We can also just discard it.

ximon18:
Since there doesn't seem to be consensus about keeping this test script I will abandon this PR.

PR #682, opened 2026-05-29
Fix override not starting signing

tertsdiepraam:
Override seems to be broken since we started doing persistence as the persistence was not taking into account that signing could be triggered from the halted state. To fix that, we now trigger persistence on override as well. To allow for that, I had to move the transition to the signing state to *before* the persistence instead of after. I believe this is still correct and maybe even nicer because it guards against approving or overriding twice.

---

### DIFF
--- a/src/zone/machine.rs
+++ b/src/zone/machine.rs
@@ -344,6 +344,13 @@ impl<'a> ZoneHandle<'a> {
             None, // TODO
         );
 
+        // Move to the 'Waiting' state.
+        let (transition, state) = self.state.machine.transition();
+        let ZoneStateMachine::SignedReview(signed) = state else {
+            panic!("The zone must be in signer review")
+        };
+        transition.move_to(ZoneStateMachine::Waiting(signed.approve()));
+
         // Persist the signed instance while remaining in the 'SignedReview'
         // state. Once persistence is complete, 'switch()' will be called, which
         // will transition to the 'Waiting' state.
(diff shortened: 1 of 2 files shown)


src/zone/machine.rs
```diff
@@ -204,6 +204,12 @@ impl<'a> ZoneHandle<'a> {
             None, // TODO
         );
 
+        let (transition, state) = self.state.machine.transition();
+        let ZoneStateMachine::LoadedReview(loaded) = state else {
+            panic!("cannot soft reject loaded in this state");
```
bal-e: The panic message needs to be updated.

src/zone/machine.rs
```diff
@@ -339,6 +344,13 @@ impl<'a> ZoneHandle<'a> {
...
         );
 
+        // Move to the 'Waiting' state.
+        let (transition, state) = self.state.machine.transition();
+        let ZoneStateMachine::SignedReview(signed) = state else {
+            panic!("The zone must be in signer review")
+        };
+        transition.move_to(ZoneStateMachine::Waiting(signed.approve()));
+
         // Persist the signed instance while remaining in the 'SignedReview'
```
bal-e: This comment needs to be updated.

src/zone/machine.rs
```diff
@@ -462,7 +449,11 @@ impl<'a> ZoneHandle<'a> {
 
         transition.move_to(ZoneStateMachine::Signing(halt.override_rejection()));
 
-        self.storage().accept_loaded();
+        // Persist the loaded instance while remaining in the 'LoadedReview'
```
bal-e: This comment needs to be updated; we are in the `Signing` state here.

src/zone/machine.rs
```diff
@@ -480,7 +471,11 @@ impl<'a> ZoneHandle<'a> {
 
         transition.move_to(ZoneStateMachine::Waiting(halt_signed.override_rejection()));
 
-        self.storage().accept_signed();
+        // Persist the signed instance while remaining in the 'SignedReview'
```
bal-e: This comment needs to be updated; we are in the `SignedReview` state here.

bal-e (changes requested):
Looks good wrt. functionality, but the comments in the code need to be updated.

Also: Do you want to add some `#[must_use]` annotations on the related functions? Or do that in a separate PR?

tertsdiepraam:
Fixed the comments and the panic message. I added the `#[must_use]` on the type instead of the functions.

bal-e (approved):
Nice!

PR #683, opened 2026-05-29
record timestamp and number of records on publication

tertsdiepraam:
Record the timestamp and number of records in the signed zone on publication and show that in the zone status:

```
zone:   example.com
policy: foo
source: /home/terts/data/zones/example.txt

review
  loaded: manual
  signed: manual

last published
  loaded serial: 2001062501
  signed serial: 2026052902
  timestamp:     2026-05-29T12:40:47.746456816Z
  size:          44 records

status: idle

Published zone available at:
  - 127.0.0.1:4542
  - [::1]:4542
```

Doesn't include the size in bytes yet because that's harder to compute, but this is a good start and the number of records is a nice sanity check.

---

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -466,10 +466,9 @@ pub struct ZoneStatus {
 pub struct LastPublishedZone {
     pub loaded_serial: Serial,
     pub signed_serial: Serial,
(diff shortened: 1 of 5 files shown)


bal-e (approved):
Nice!

PR #686, opened 2026-06-01
Add an integration test that signs using SoftHSMv2.

ximon18:
A simple integration-test that tests signing using a HSM via a specific version of Cascade-HSM-Bridge (as it is pending a new release) and using SoftHSMv2.


integration-tests/cascade-test-image/Dockerfile
```diff
@@ -54,6 +57,18 @@ RUN <<EOF
...
 
+RUN <<EOF
+  # Install cascade-hsm-bridge
+  cargo install --debug --git https://github.com/nlnetlabs/cascade-hsm-bridge --root /root/cargo-debug --locked
+  cargo install --git https://github.com/nlnetlabs/cascade-hsm-bridge --root /root/cargo-release --locked
+EOF
+
+RUN <<EOF
+  # Install and initialize SoftHSMv2
+  apt-get install -y libsofthsm2
```
bal-e, line 68: Would it be better to this install along with all the other `apt-get` ones?

bal-e (approved):
Looks good. Isn't the cascade-hsm-bridge release done now, so we can rely on that over Git?

integration-tests/cascade-test-image/Dockerfile
ximon18: I didn't do that as I wanted to keep the SoftHSM steps together but honestly I'm fine either way.

ximon18:
> Looks good. Isn't the cascade-hsm-bridge release done now, so we can rely on that over Git?

One issue with this is the way @mozzieongit set this up it installs both a debug and a release version of tools so I continued that pattern which requires building from sources. Also, if we want to run integration tests with newer-not-yet-released versions of the bridge then one would need to install via git, not installation by DEB (which I assume you were referring to).

In an ideal world we wouldn't have this test at all in this form, instead we would keep bridge testing in the bridge repo and test against a KMIP mock in the Cascade repo, also so that we can test handling of unhappy flows such as unexpected connection errors and KMIP protocol responses.

Having an actual end-to-end test with a "real" HSM and the bridge still has value though, and I wouldn't want to find out issues only after doing a release of the bridge, i.e. install via DEB in this test.

integration-tests/cascade-test-image/Dockerfile
bal-e: I think it would be better to keep the `apt-get` steps together because the order and grouping of these steps affects how much work is re-done when context changes for the Docker build.

bal-e:
I see! I'm happy to leave it as-is then.

integration-tests/cascade-test-image/Dockerfile
ximon18: Regarding Docker build layer caching the apt install steps are somewhat ironic in that they are grouped together inside a single RUN statement and so changing or adding one apt install line will cause the entire group to be re-run.

PR #691, opened 2026-06-02
Update docs

Philip-NLnetLabs:
Update documentation for the beta release.

### DIFF
--- a/doc/manual/source/review-hooks.rst
+++ b/doc/manual/source/review-hooks.rst
@@ -144,9 +144,6 @@ script\ [1]_ as ``/usr/local/bin/cascade-review.sh``:
             dnssec-verify -q -o "${CASCADE_ZONE}" /dev/stdin
         fi
 
-.. versionadded:: 0.1.0-alpha2
-   Environment variables ``CASCADE_SERVER_IP`` and ``CASCADE_SERVER_PORT``.
-
 Next, we update the zone's policy to use the review script for both stages:
 
 .. code:: toml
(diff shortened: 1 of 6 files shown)


doc/manual/source/review-hooks.rst
```diff
@@ -144,9 +144,6 @@ script\ [1]_ as ``/usr/local/bin/cascade-review.sh``:
             dnssec-verify -q -o "${CASCADE_ZONE}" /dev/stdin
         fi
 
-.. versionadded:: 0.1.0-alpha2
```
ximon18, line 147: Why has this been removed?

ximon18 (approved):
Nice.

doc/manual/source/review-hooks.rst
Philip-NLnetLabs: We work towards a release. I doubt anyone would like to see documentation littered with historical details of when features where added during alpha and beta release. This is also alpha2, it is no relevant to anybody upgrading to the beta.

It make more sense to do this for release versions.

ximon18: I agree, but was thinking during alpha and beta releases they are useful to people testing and that we would remove them all on first production release.

Philip-NLnetLabs: I come across it, found and found it weirdly cluttering the text. I think if it would have been more recent (or more important) I would have left it.

ximon18: Alpha2 is indeed quite irrelevant now, we don't want anybody testing or providing feedback on that, nor do we provide an upgrade path during the pre-production releases. So maybe changes since alpha5 are worth calling out but anything older is not.

PR #692, opened 2026-06-03
Improve match statements around review server

tertsdiepraam:
This PR tries to improve the readability of a match expression in `src/units/zone_server.rs`. It also helps reduce the number of panics. Finally, the phrasing was kind of outdated (e.g. referred to `required`) so I updated that too.

---

### DIFF
--- a/src/units/zone_server.rs
+++ b/src/units/zone_server.rs
@@ -297,25 +297,41 @@ impl ZoneServer {
         };
 
         let zone_name = &zone.name;
-        if let ReviewMode::Off = review.mode {
-            // Approve immediately.
-            match self.source {
-                Source::Unsigned => {
-                    info!("[{unit_name}]: Adding '{zone_name}' to the set of signable zones.");
-                    self.on_unsigned_zone_approved(center, zone, zone_serial);
+        match review.mode {
+            ReviewMode::Off => {
+                // Approve immediately.
+                match self.source {
+                    Source::Unsigned => {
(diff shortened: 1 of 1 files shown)


src/units/zone_server.rs
```diff
@@ -352,35 +368,17 @@ impl ZoneServer {
...
-            info!(
-                "[{unit_name}]: Approve with command: cascade zone approve --{zone_type} {zone_name} {zone_serial}"
-            );
-            info!(
-                "[{unit_name}]: Reject with command: cascade zone reject --{zone_type} {zone_name} {zone_serial}"
+        let ReviewMode::Script { hook } = review.mode else {
+            // The only other case is ReviewMode::Manual, in which case we don't
+            // need to do anything here anymore.
+            return None;
+        };
```
bal-e, line 373: Might be good to make this a `match` statement so compilation fails (and we realize we need to update this) if new modes are added?

bal-e (approved):
Happy to see this code get simpler :)

src/units/zone_server.rs
tertsdiepraam: Given the context that matching on all of the branches already happens above I think this is fine? Note that it would add an `unreachable` for `ReviewMode::Off` too (really wishing that we'd have `Option<ReviewMode>` with `None` being `Off` right now).

tertsdiepraam:
Merge went wrong from GitHub's perspective so I force pushed a rebase of the same thing

Issue #702, opened 2026-06-06
Slightly confusing syntax (to me) when using TSIG on incoming XFR


jpmens:
cascade 0.1.0-beta2-dev at 2f9fee20df

I saw the policy configures the TSIG key to use during XFR, and that is clear enough:

```toml
# The set of nameservers to accept zone transfer requests from.
# 
# Each nameserver must be specified as a string in the form:
#
# `"<IP>[^<TSIG_KEY_NAME>]"`
#
# If a TSIG key name is specified, a key by that name must exist in the
# Cascade TSIG key store and will be used to authenticate communication with
# the nameserver.
#
# If not specified, zone transfer requests will be accepted from any
# nameserver.
accept-xfr-from = [ "192.168.1.140^c01" ]
```

While reading the comments in the policy, where it says "The set of nameservers to accept zone transfer requests from", I was surprised to not detect the use of a TCP port number.

Later, when adding the zone to cascade, I am not permitted to specify the TSIG key, however I do specify a source port:

```console
$ cascade zone add --source 192.168.1.140:5034 --policy k01 jp.aa 
```

I don't, however, know how to phrase this better, so you'll do well ignoring it. :-)

Philip-NLnetLabs:
This part of policy configures XFR out, so to nameservers that fetch the signed zone. The port of the publication server is specified in the config file (because it applies to all zones).

The port in the --source argument is where to fetch the unsigned zone from.

jpmens:
I was slowly getting there ... thanks, @Philip-NLnetLabs. In that case I would like to propose a change in wording:

replace `to accept zone transfer requests from` by `to provide zone transfers to`

Philip-NLnetLabs:
Yes, good suggestion. This text clearly needs some work. It may be a good idea to also explicitly refer to the signed zone. To further reduce confusion.

ximon18:
`provide` is also the term used by NSD, e.g. see https://nsd.docs.nlnetlabs.nl/en/latest/manpages/nsd.conf.html#provide.

ximon18:
We can also rename the setting from `accept-xfr-from` to `provide-xfr-to`. That would be a breaking change as we are not supporting backward compatibility until we start publishing what we consider to be production ready versions of Cascdade.

[...]

PR #704, opened 2026-06-06
fixes a few typos and I take the liberty of using the term "provides"

jpmens:
when referring to zone transfers made available to secondaries. We just mentioned this in #702 (IIRC that's also the terminology used by NSD)

Provide your description here

---

### DIFF
--- a/doc/manual/source/zone-transfers.rst
+++ b/doc/manual/source/zone-transfers.rst
@@ -2,20 +2,20 @@ Zone Transfers
 ==============
 
 Cascade is expected to be deployed between a hidden upstream nameserver and
-public downstream nameservers. The hidden upstream serves the unsigned zone,
-Cascade signs it, and serves it to downstream nameservers for publication
+public downstream nameserver. The hidden upstream serves the unsigned zone,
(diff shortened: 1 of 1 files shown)


doc/manual/source/zone-transfers.rst
```diff
@@ -2,20 +2,20 @@ Zone Transfers
 ==============
 
 Cascade is expected to be deployed between a hidden upstream nameserver and
-public downstream nameservers. The hidden upstream serves the unsigned zone,
-Cascade signs it, and serves it to downstream nameservers for publication
+public downstream nameserver. The hidden upstream serves the unsigned zone,
```
bal-e: I think this should remain "nameservers" -- Cascade supports notifying and transferring to multiple downstream nameservers for a single zone.

bal-e (commented):
Thanks for helping out @jpmens :) Just one minor note, otherwise LGTM.

doc/manual/source/zone-transfers.rst
jpmens: fixed

bal-e (approved):
LGTM!

PR #708, opened 2026-06-06
required=true no longer seems to exist, hence remove its mention from…

jpmens:
comments in the templated policy mention a `required` option which no longer appears to exist.


tertsdiepraam (approved):

tertsdiepraam:
Thank you!

PR #711, opened 2026-06-08
Corrections to the "Key Management" documentation page.

jpmens:
to reflect that TSIG auth for RRSIG check currently exists (untested, but the policy says it's there :-)

Provide your description here

### DIFF
--- a/doc/manual/source/key-management.rst
+++ b/doc/manual/source/key-management.rst
@@ -5,7 +5,7 @@ Key Management
    the implementation in OpenDNSSEC and for example BIND. 
    
    The goal is that key rolls should always go through the same sequence of 
-   steps. As much as possible, we strive get all key rolls in a single mold.
+   steps. As much as possible, we strive to get all key rolls in a single mold.
    They will always follow the same pattern, while the details of a 
    :term:`KSK <Key signing key (KSK)>` and :term:`ZSK <Zone signing key 
    (ZSK)>` roll will be different.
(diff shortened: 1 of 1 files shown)


doc/manual/source/key-management.rst
```diff
@@ -125,9 +125,8 @@ do this, the key manager needs network access to those nameservers.
    :ref:`disable (part of) this functionality <automation-control>`. 
 
 To check the signatures in the zone, the key manager will issue an AXFR
-request to the primary nameserver listed in the SOA record of the zone. In
-the future we plan to make it possible to configure which nameserver should
-be used and which TSIG keys should be used for authentication.
+request to the primary nameserver listed in the SOA record of the zone to
```
ximon18: @jpmens: Did you mean to say "or" here?
```suggestion
request to the primary nameserver listed in the SOA record of the zone or to
```

doc/manual/source/key-management.rst
```diff
@@ -125,9 +125,8 @@ do this, the key manager needs network access to those nameservers.
...
+a configurable nameserver with configurable TSIG keys for authentication.
```
ximon18: Perhaps this would be better?
```suggestion
request to the primary nameserver listed in the SOA record of the zone, or
if configured to a set of one or more nameservers specified in policy along with optional TSIG keys for authentication.
```

jpmens: github doesn't permit me to apply this suggestion of  yours ... 
I think it's done

ximon18: Ah, that's because it conflicts with the first suggestion I made which you accepted. I think we can just accept the change to use "or" that you accepted, the additional information is documented elsewhere such as with the policy field `publication-nameservers`.

ximon18 (approved):
LGTM, thanks @jpmens!

PR #714, opened 2026-06-08
Note in CLI help that source to zone add can include a TSIG key name. (resolves #712)

### LINKED ISSUE #712: `zone add --help` doesn't mention TSIG

The `--source <SOURCE>` parameter is missing a mention of an optional `^<TSIG key name>`, as I discover from #703

ximon18:
Before:

```
$ cascade zone add --help
Usage: cascade zone add [OPTIONS] --source <SOURCE> --policy <POLICY> <NAME>
...
      --source <SOURCE>
          The zone source can be an IP address (with or without port, defaults to
          port 53) or a file path
```

After:

```
$ cascade zone add -h
Usage: cascade zone add [OPTIONS] --source <SOURCE> --policy <POLICY> <NAME>
...
      --source <SOURCE>
          The source to obtain the zone content from: IP:[PORT][^TSIG_KEY_NAME] (port
          defaults to 53) or the path to a zone file locally available to the
          `cascaded` daemon
```

The new text both includes mention of the TSIG key support and more closely matches the [ReadTheDocs zone add documentation](https://cascade.docs.nlnetlabs.nl/en/latest/man/cascade-zone.html#options-for-zone-add) noting the important point that the path if provided cannot be local to the client host but most be available on the server host.

@jpmens: Would this resolve #712 for you?

### DIFF
--- a/crates/cli/src/commands/zone.rs
+++ b/crates/cli/src/commands/zone.rs
@@ -23,8 +23,8 @@ pub enum ZoneCommand {
     Add {
         name: ZoneName,
 
-        /// The zone source can be an IP address (with or without port,
-        /// defaults to port 53) or a file path.
+        /// The source to obtain the zone content from:
+        /// IP:[PORT][^TSIG_KEY_NAME] or a file path.
         // TODO: allow supplying different tcp and/or udp port?
         #[arg(long = "source")]
         source: ZoneSource,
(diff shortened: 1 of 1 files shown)


jpmens:
@ximon18 definitely, thank you. That looks very good to me.

PR #718, opened 2026-06-08
Refactor signing code out of `zone_signer.rs`

bal-e:
While working on #630 I noticed key loading functionality was distributed throughout `zone_signer.rs` and thought it nice to move it to a dedicated file. Along the way, I cleaned up some error messages, and turned some panics into signing failures (related to the client credentials store).

This PR also includes #865. Together, they move the vast majority of functionality out of `zone_signer.rs`.

---


src/units/zone_signer.rs
```diff
@@ -487,18 +435,7 @@ impl ZoneSigner {
         debug!("Loading dnst keyset signing keys");
         status.write().unwrap().current_action = "Loading signing keys".to_string();
         // Load the signing keys indicated by the keyset state.
-        let signing_keys = load_keys(self, center, zone_name.clone(), &state, status.clone())?;
-
-        debug!("{} signing keys loaded", signing_keys.len());
-
-        // TODO: If signing is disabled for a zone should we then allow the
-        // unsigned zone to propagate through the pipeline?
-        if signing_keys.is_empty() {
```
ximon18, line 375: Has the functionality changed here? It looks like a check for empty signing keys has been removed as I don't see that check in the new `load()` function,

bal-e: Oh, whoops. This check was duplicated here and in the previous `load_keys()`, so I had assumed I had copied it into `ZoneSigningKeys::load()`. Good catch, will fix!

src/units/zone_signer.rs
```diff
@@ -1035,206 +934,6 @@ pub fn load_binary_file(path: &Path) -> Vec<u8> {
...
-                                read_timeout: Some(read_timeout),
-                                write_timeout: Some(write_timeout),
-                                max_response_bytes: Some(max_response_bytes),
-                            };
-
-                            let cloned_status = status.clone();
-                            let cloned_server_id = server_id.clone();
-                            tokio::task::spawn(async move {
-                                cloned_status.write().unwrap().current_action =
-                                    format!("Connecting to KMIP server '{cloned_server_id}");
```
ximon18, line 1172: This status write and the one below appear to have been lost, while the one about "Loading signing keys" was preserved. This seems inconsistent. I'm not sure where the status of connecting to the KMIP server would actually appear in Cascade now compared to when this code was originally written, does it appear anywhere? I wonder if it makes sense to keep some status writes and not others. Potentially slow/long running actions like connecting to a KMIP server and waiting for it to time out could be useful to tell the user to prevent them wondering what Cascade is doing.

src/signer/keys.rs
```diff
@@ -0,0 +1,633 @@
...
+
+                let mut username = None;
+                let mut password = None;
+                if has_credentials {
+                    let creds_path = &center.config.kmip_credentials_store_path;
+                    let creds_file = KmipClientCredentialsFile::new(
+                        creds_path.as_std_path(),
+                        KmipServerCredentialsFileMode::ReadOnly,
+                    )
+                    .unwrap();
```
ximon18: You converted a panic above about being unable to open or parse a KMIP server state file above, but you left this unwrap which opens and reads from the credentials state file, was that intentional?

ximon18 (approved):
I like this a lot. I just noted a couple of minor things I wanted to raise, see my comments, but they are not blocking.

src/signer/keys.rs
bal-e: Ooh, good catch! I'll fix this too.

src/units/zone_signer.rs
bal-e: That's fair, I'll add it back.

ximon18 (approved):

PR #719, opened 2026-06-08
Extend `cascade tsig remove` error to report the users of the key. (resolves #705)

### LINKED ISSUE #705: Would `cascade tsig remove` be able to list which zones the TSIG is in use for?

When I attempt to remove a TSIG key I, correctly, am informed the key is in use and cannot be removed.

```console
$ cascade tsig remove c01
2026-06-06T11:58:26.440665Z ERROR cascade: Failed to remove TSIG key: the TSIG key cannot be removed as it is in use
```

If it isn't more than a few lines of code, what would slightly improve usability is to be informed on which zones the key is in use.

**Ceci n'est pas une bogue**, as Stéphane would rightly say! 😊

ximon18:
As requested by @jpmens via #705.

~The code feels a bit verbose and clunky, suggestions for improvement welcome.~

Attempts to remove a TSIG key in use will now report, comma-separated, the names of the zones and/or policies using the key.

---


crates/api/src/lib.rs
```diff
@@ -260,16 +260,56 @@ pub enum TsigRemoveError {
...
+    // The key is in use for a zone for some other reason.
+    InUseByZoneOther {
+        using_zones: Vec<ZoneName>,
+    },
+
+    // The key is in use by one or more policies to authenticate communication
+    // with downstream nameservers.
+    InUseByPolicy {
+        using_policies: Vec<Box<str>>,
+    },
```
bal-e: This feels weird to me because, conceptually, multiple such errors can happen simultaneously; a key may be used by the zone source of `a.example.org` while being used by policy `bar`, and it would be nice to report both. Why not keep the single `InUse` variant and keep all three `Vec`s (or better, a `Vec` of `enum`s) in there?

bal-e: (likewise in `src/tsig/mod.rs`)

ximon18: We can do it that way. I didn't do that as it requires more work (currently the code aborts early, it would have to accumulate information for error reporting and complete all checks before returning).

bal-e: But the current way, IMO, risks giving the operator more confusing information. Maybe give it a try -- if it turns out to be prohibitively complex, stick with the current behavior and we can do something about it later.

ximon18: I will try changing it.

ximon18: Done. Now outputs each known reference to the TSIG key, one per line, to the CLI user, e.g.:
 
```
    $ cascade tsig remove my_key
    2026-06-15T10:23:19.191825Z ERROR cascade: Failed to remove TSIG key:
    key is still in use by:
      - The source of zone 'some-zone-name'
      - Policy 'some-policy-name'
```

ximon18: Note: I have not added any tests for this new functionality yet. Given the dependence of `remove_key()` on `Center` I would need a mock `Center` to be able to unit test it. I could extend a system test to test as I did manually, e.g.

- remove a key that doesn't exist
- add a key
- remove and re-add the key
- add a zone that does not use the key
- remove and re-add the key
- add a zone that uses the key as a source
- removing the key should fail and report one usage
- edit the policy to reference the key
- reload policy
- removing the key should fail and report two usages
- etc

But it's not a tiny amount of work.

bal-e: Nice! The lack of testing is understandable, system tests don't feel like the right tool for this.

src/tsig/mod.rs
```diff
@@ -291,57 +293,81 @@ pub fn remove_key(center: &Arc<Center>, name: &tsig::KeyName) -> Result<(), Remo
...
+                    || p.server
+                        .outbound
+                        .send_notify_to
+                        .iter()
+                        .any(|acl| acl.tsig_key_name.as_ref() == Some(name))
+            })
+            .map(UsageReference::Policy),
+    );
+
+    if !refs.is_empty() {
```
bal-e, line 369: What if `refs` is empty but `unknown_refs` is not? That is possible, right?

src/tsig/mod.rs
```diff
@@ -291,57 +293,81 @@ pub fn remove_key(center: &Arc<Center>, name: &tsig::KeyName) -> Result<(), Remo
...
+        .filter(|z| {
+            matches!(
+                z.0.state.read().loader.source,
+                crate::loader::Source::Server { tsig_key: Some(ref key), .. }
+                    if name == key.name()
+            )
+        })
+        .inspect(|&referenced_zone| {
+            // Lookup the zone by name because:
+            //   a) if using ZoneByName Clippy complains about a mutable key
```
bal-e: The Clippy lint is a frustrating false positive. I'll look into fixing it for later.

bal-e (commented):
Looking awesome! Just one comment about unknown references.

src/tsig/mod.rs
ximon18: Good catch. Fixed.

src/tsig/mod.rs
ximon18: Fixing the above feedback was easier using `ZoneByName` so I added an allow for the mutable key warning by Clippy.

bal-e (approved):
Awesome, LGTM :)

PR #722, opened 2026-06-08
Rename `accept-xfr-from` to `provide-xfr-to`. (resolves #702)

### LINKED ISSUE #702: Slightly confusing syntax (to me) when using TSIG on incoming XFR

ximon18:
As requested by @jpmens via #702.

### DIFF
--- a/crates/cli/src/commands/policy.rs
+++ b/crates/cli/src/commands/policy.rs
@@ -164,7 +164,7 @@ fn print_policy(p: &PolicyInfo) {
     println!("  server:");
     println!("    outbound:");
     println!("      accept XFR from:");
-    print_nameserver_comms_policy(&p.server.outbound.accept_xfr_from);
+    print_nameserver_comms_policy(&p.server.outbound.provide_xfr_to);
     println!("      send NOTIFY to:");
     print_nameserver_comms_policy(&p.server.outbound.send_notify_to);
 }
--- a/crates/cli/src/commands/policy.rs
+++ b/crates/cli/src/commands/policy.rs
@@ -163,7 +163,7 @@ fn print_policy(p: &PolicyInfo) {
     print_review(&p.signer.review);
     println!("  server:");
     println!("    outbound:");
-    println!("      accept XFR from:");
+    println!("      provide XFR to:");
     print_nameserver_comms_policy(&p.server.outbound.provide_xfr_to);
     println!("      send NOTIFY to:");
     print_nameserver_comms_policy(&p.server.outbound.send_notify_to);
--- a/doc/manual/source/man/cascaded-policy.toml.rst
+++ b/doc/manual/source/man/cascaded-policy.toml.rst
@@ -587,12 +587,12 @@ The ``[server.outbound]`` section.
    Cascade TSIG key store and will be used to authenticate communication with
    the nameserver.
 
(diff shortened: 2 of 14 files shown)


crates/cli/src/commands/policy.rs
```diff
@@ -164,7 +164,7 @@ fn print_policy(p: &PolicyInfo) {
     println!("  server:");
     println!("    outbound:");
     println!("      accept XFR from:");
```
bal-e: Should we update this line too?

doc/manual/source/man/cascaded-policy.toml.rst
```diff
@@ -587,12 +587,12 @@ The ``[server.outbound]`` section.
...
-.. option:: accept-xfr-from = []
+.. option:: provide-xfr-to = []
 
-   The set of nameservers to accept zone transfer requests from.
+   The set of nameservers to provide zone transfers to.
 
-   If no nameservers are specified, zone transfer requests will be accepted
-   from any nameserver.
+   If no nameservers are specified, zone transfer requests will be provided to
+   any nameserver.
```
bal-e: "requests will be provided to" doesn't sound right, this sentence might need a bit more work.

etc/policy.template.toml
```diff
@@ -462,6 +462,8 @@ send-notify-to = []
 # Cascade TSIG key store and will be used to authenticate communication with
 # the nameserver.
 #
-# If not specified, zone transfer requests will be accepted from any
-# nameserver.
-accept-xfr-from = []
+# Zone transfer will be provided to any nameserver that initiates a transfer
+# request from one of the specified IP addresses.
```
bal-e: "Zone transfer" => "Zone transfers"?

bal-e (changes requested):
Looks good, just some minor concerns :)

crates/cli/src/commands/policy.rs
ximon18: Good catch. I'm surprised my `accept.xfr.from` regex didn't catch that.

doc/manual/source/man/cascaded-policy.toml.rst
ximon18: This is indeed wrong. Will fix.

ximon18:
I re-ran the system tests and they all pass. I will merge this.

Issue #740, opened 2026-06-09
Use `Scheduler` for tracking zone re-signing


bal-e:
At the moment, `zone_signer.rs` tracks a "next resign time"; whenever that time is reached, it scans all zones looking for those with expired signatures. After re-signing, it updates the "next resign time" by scanning all the zones again. This is a lot of unnecessary work, and it hinders us from giving information to the user like "here are the next N re-signing events that will happen". To fix this, the zone signer's scheduling code should be rewritten to use `common/scheduler.rs`, which was developed for the zone loader and has already been factored out into a standalone module.

bal-e:
I've tried doing this a few times; annoyingly, it is blocked by #741.

bal-e:
https://github.com/NLnetLabs/cascade/pull/784/changes#r3394074802 -- #784 is going to resolve one bug by introducing another, and this PR is needed to fix the new bug.

Edit: #784 has been closed in favor of solving this issue properly. Which means that this is now tied to #730.

Issue #731, opened 2026-06-09
cascaded silently dies: resource problem?


jpmens:
Am I trying too much on this 6GB RAM machine?

I add `.ch` as a test.

```
$ cascade zone status ch
zone:   ch
policy: 13
source: 86.119.81.74:53 with TSIG key 'tsig-zonedata-ch-public-21-01'

review
  loaded: off
  signed: off

last published
  <no versions published yet>

status: loading
  ● load
  |   serial: <not yet known>
  |   start time: 2026-06-09T14:57:55 (1m 46s ago)
  |   progress: 663 MB
  |
  ○ review loaded zone (disabled)
  ○ sign
  ○ review signed zone (disabled)
  ○ publish
```
  

  
I am slightly confused that loading the zone has restarted?
  
```
$ cascade zone status ch
zone:   ch
policy: 13
source: 86.119.81.74:53 with TSIG key 'tsig-zonedata-ch-public-21-01'

review
  loaded: off
  signed: off

last published
  <no versions published yet>

status: loading
  ● load
  |   serial: <not yet known>
  |   start time: 2026-06-09T15:00:16 (29s ago)
  |   progress: 182 MB
  |
  ○ review loaded zone (disabled)
  ○ sign
  ○ review signed zone (disabled)
  ○ publish
```

A bit later

```
$ cascade zone status ch
zone:   ch
policy: 13
source: 86.119.81.74:53 with TSIG key 'tsig-zonedata-ch-public-21-01'

review
  loaded: off
  signed: off

last published
  <no versions published yet>

status: loading
  ● load
  |   serial: <not yet known>
  |   start time: 2026-06-09T15:01:41 (1m 54s ago)
  |   progress: 710 MB
  |
  ○ review loaded zone (disabled)
  ○ sign
  ○ review signed zone (disabled)
  ○ publish

  ```
  
  Hoping to see "sign" light up in status:
  
```
  $ cascade zone status ch
2026-06-09T15:02:27.049371Z ERROR cascade: HTTP connection failed: client error (Connect): tcp connect error: Connection refused (os error 111)
```

```
$ grep ERROR /tmp/cascaded.log
2026-06-09T14:47:26.065935Z ERROR cascaded::daemon: thread 'main' (ProcessId(82128), ThreadId(1)) panicked at src/units/zone_signer.rs:832:22: called `Option::unwrap()` on a `None` value
[...]

Issue #730, opened 2026-06-09
ERROR cascaded::daemon: thread 'main': panicked at src/units/zone_signer.rs:832:22: called `Option::unwrap()` on a `None` value


jpmens:
Really sorry about this ...

For five minutes now I've been adding and removing TSIG keys, adding and removing zones and then I get an ECONNREFUSED on the `cascade` utility. I assume it's due to a mixup of TSIG keys while attempting to source .CH

cascade 0.1.0-beta2-dev at 2106dd7712

The server log shows:

```
2026-06-09T14:38:41.078489Z  INFO cascaded::signer::zone: Enqueuing a re-sign operation
2026-06-09T14:38:41.078522Z  INFO cascaded::units::zone_signer: [ZS]: Waiting to enqueue signing operation for zone 'e3.aa'.
2026-06-09T14:38:41.078533Z  INFO cascaded::units::zone_signer: [ZS]: Waiting to start signing operation for zone 'e3.aa': 0 signing operations are in progress and 0 operations are queued ahead of us.
2026-06-09T14:39:05.455954Z  INFO cascaded::loader: Refreshing Name(ch.)
2026-06-09T14:39:05.555167Z ERROR cascaded::loader: Could not load the zone: the AXFR failed: the server's response was semantically incorrect: Not a valid XFR response zone=ch
2026-06-09T14:39:30.439008Z  INFO cascaded::center: Removed zone 'ch'
2026-06-09T14:40:59.436937Z  INFO cascaded::center: Added TSIG key 'tsig-zonedata-ch-public-21-01'
2026-06-09T14:41:09.920035Z  INFO cascaded::loader::zone: Setting source of zone 'ch' from 'none' to '86.119.81.74:53 with TSIG key 'tsig-zonedata-ch-public-21-01''
2026-06-09T14:41:09.920072Z  INFO cascaded::center: Added zone 'ch'
2026-06-09T14:41:09.920119Z  INFO cascaded::loader: Refreshing Name(ch.)
2026-06-09T14:41:10.014257Z ERROR cascaded::loader: Could not load the zone: the AXFR failed: could not communicate with the server: authentication error: unknown key on server zone=ch
2026-06-09T14:41:44.070892Z  INFO cascaded::center: Removed zone 'ch'
[...]

PR #784, opened 2026-06-10
Simplify re-signing search

bal-e:
The zone signer performs a search across all zones to find those that need re-signing. This search assumed that zones had policies, leading to race conditions with zone deletion (#730, #731). It also involved a `resign_busy` field in `Center` that could be replaced with some simple checks in `ZoneState`. While larger rewrites of this code are planned (#740), this PR simplifies the re-signing code, gracefully handles zones without policies, and removes the `resign_busy` field.

This may help with #730 and #731 but is not guaranteed to resolve them.

---

### DIFF
--- a/src/units/zone_signer.rs
+++ b/src/units/zone_signer.rs
@@ -57,6 +57,7 @@ use crate::units::key_manager::{
     KmipClientCredentialsFile, KmipServerCredentialsFileMode, mk_dnst_keyset_state_file_path,
 };
 use crate::util::AbortOnDrop;
+use crate::zone::machine::ZoneStateMachine;
 use crate::zone::{HistoricalEvent, Zone, ZoneByName};
 
 // TODO: This comment is outdated in several ways.
@@ -761,6 +762,12 @@ impl ZoneSigner {
                 trace!("Ignoring {zone_name}, it has no policy");
                 continue;
             };
+
+            // If the zone is undergoing signing already, don't factor it in.
+            if matches!(zone_state.machine, ZoneStateMachine::Signing(_)) {
+                trace!("Ignoring {zone_name}, it is already (re-)signing");
+                continue;
+            }
 
             // Compute when the zone's signatures need to be refreshed next.
             //
--- a/src/units/zone_signer.rs
+++ b/src/units/zone_signer.rs
@@ -57,7 +57,6 @@ use crate::units::key_manager::{
     KmipClientCredentialsFile, KmipServerCredentialsFileMode, mk_dnst_keyset_state_file_path,
 };
 use crate::util::AbortOnDrop;
-use crate::zone::machine::ZoneStateMachine;
 use crate::zone::{HistoricalEvent, Zone, ZoneByName};
 
 // TODO: This comment is outdated in several ways.
@@ -763,13 +762,6 @@ impl ZoneSigner {
                 continue;
             };
 
-            // If the zone is undergoing some operation already, don't factor it
-            // into the signing time. It will be (or is already being) resigned.
-            if !matches!(zone_state.machine, ZoneStateMachine::Waiting(_)) {
-                trace!("Ignoring {zone_name}, it is busy");
-                continue;
-            }
-
             // Compute when the zone's signatures need to be refreshed next.
             //
             // TODO: Stop storing durations and timestamps as 'u32's.
@@ -822,13 +814,6 @@ impl ZoneSigner {
                 trace!("Ignoring {zone_name}, it has no policy");
                 continue;
             };
-
-            // If the zone is undergoing some operation already, it doesn't
-            // need to be scheduled for re-signing.
-            if !matches!(zone_state.machine, ZoneStateMachine::Waiting(_)) {
-                trace!("Ignoring {zone_name}, it is busy");
-                continue;
-            }
 
             // Compute when the zone's signatures need to be refreshed next.
             //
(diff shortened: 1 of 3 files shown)


bal-e:
Sorry about the force-pushes, accidentally got unrelated code in there.

src/units/zone_signer.rs
```diff
@@ -812,57 +814,48 @@ impl ZoneSigner {
...
+            let zone_state = zone.read();
+
+            let Some(policy) = &zone_state.policy else {
+                trace!("Ignoring {zone_name}, it has no policy");
+                continue;
+            };
+
+            // If the zone is undergoing some operation already, don't factor it
+            // into the signing time. It will be (or is already being) resigned.
+            if !matches!(zone_state.machine, ZoneStateMachine::Waiting(_)) {
```
Philip-NLnetLabs: Are all states other than Waiting related to signing? For example, what if the state is Loading?

src/units/zone_signer.rs
```diff
@@ -812,57 +814,48 @@ impl ZoneSigner {
...
-                        trace!("[ZS]: resign: zone {zone_name} is busy");
-                        continue;
-                    }
-
-                    // Zone has been resigned. Remove this entry.
-                    resign_busy.remove(zone_name);
-                }
-            }
-
+            // TODO: This does *not* look very Y2038-safe.
```
Philip-NLnetLabs: At this point, curr_refresh_time should be UnixTime. Did I get it wrong?

src/units/zone_signer.rs
bal-e: If the state is loading, then it will be signed afterwards, so we don't need to enqueue it for re-signing.

src/units/zone_signer.rs
bal-e: Oh, you're right, I didn't closely enough at the types involved. I'll remove the comments. Maybe `UnixTime` should offer some direct conversion to and from `SystemTime` to make it more obvious that it is doing the right thing (or even store a `SystemTime` internally ... but IIRC there were problems with that).

src/units/zone_signer.rs
Philip-NLnetLabs: What if the load fails?

src/units/zone_signer.rs
Philip-NLnetLabs: The point of UnixTime is that it's internal representation is unix time, unlike the undefined representation of system time. I agree that adding conversion to SystemTime in domain is better. Maybe you can add a TODO for that.

src/units/zone_signer.rs
bal-e: Hmm, good point. The same logic applies to the original check (which checked if re-signing was already happening). I think this check can be removed entirely, I will update the code.

src/units/zone_signer.rs
```diff
@@ -746,57 +747,41 @@ impl ZoneSigner {
...
+            // Compute when the zone's signatures need to be refreshed next.
+            //
+            // TODO: Stop storing durations and timestamps as 'u32's.
+            let last_signature_refresh = zone_state.last_signature_refresh.clone();
+            let signature_refresh_interval = policy.signer.signature_refresh_interval;
             let curr_refresh_time = last_signature_refresh.clone()
                 + Duration::from_secs(signature_refresh_interval as u64);
 
-            // Start a new block to make sure the mutex is released.
-            {
```
Philip-NLnetLabs, line 775: I think something got lost here. The code used to compute the new timeout value while ignoring zones where the timer expired and didn't get resigned yet. I think the new code just computes the minimum among all zones.

bal-e: Right ...

I think that is not a problem for correctness (because zones are only re-signed if they need it and are idle).

But, if a zone's timer has expired and it is being re-signed, then the computed timeout will always say "re-signing is needed immediately" so it might cause a continuous loop.

Am I understanding correctly?

Philip-NLnetLabs: Yes, I think that will be the problem.

bal-e: Will fix.

bal-e: Hm, I thought about it further and I think filtering causes a new problem: suppose that a single zone is present, its timer has expired, and it is being re-signed. If we filter it out here, the `min_time` becomes `None`, so the timer will only be reactivated when `next_resign_time` is updated. But if re-signing / signed review fails, `next_resign_time()` will not be called, and Cascade will forget to re-sign the zone entirely.

Here's my proposal: I'll add filtering logic over here, so that zones that are not idle are ignored. Then, I'll add code to update `next_resign_time` whenever a zone becomes idle. There will be more work done on this code (#740, #617) and we can reconvene afterwards to decide whether it needs further improvement.

Philip-NLnetLabs: We can't do better in the current scheme. The reason is that the timer is based on the last resign. If signing fails then the timer will expire immediately. In that case, no timer is better than a loop. We need to move to separate timers, the current system can't be fixed properly.

bal-e: Okay, I'll add the filtering logic here and add a note that #740 will fix a known bug.

src/units/zone_signer.rs
```diff
@@ -746,57 +748,47 @@ impl ZoneSigner {
...
-                    .signature_refresh_interval;
+            let zone_state = zone.read();
+
+            let Some(policy) = &zone_state.policy else {
+                trace!("Ignoring {zone_name}, it has no policy");
+                continue;
+            };
+
+            // If the zone is undergoing signing already, don't factor it in.
+            if matches!(zone_state.machine, ZoneStateMachine::Signing(_)) {
```
Philip-NLnetLabs, line 767: What is the zone in another state? For example loading a new zone or reviewing?

bal-e:
Following further discussion with Philip, closing in favor of #740.

PR #794, opened 2026-06-10
Fix size printing

### LINKED ISSUE #629: Investigate number formatting in status
Extracted from https://github.com/NLnetLabs/cascade/issues/177.

@jpmens wrote:

> This is most definitely not a bug but actually something that amused me; I knew I had chopped the (generated) zone in vi at 5000 lines (with a formatted SOA record), so 4k looked strange:
> 

Image

> Maybe 4.9K ? :-) (ed: The fractional value issue has been split out to https://github.com/NLnetLabs/cascade/issues/460)

This is probably still an issue after the status rework.

tertsdiepraam:
Closes #629

### DIFF
--- a/crates/cli/src/commands/zone.rs
+++ b/crates/cli/src/commands/zone.rs
@@ -983,15 +983,16 @@ impl std::fmt::Display for Icon {
 /// Shows one decimal point if the size is small enough for that to make sense
 /// (e.g. below 10KB).
 fn format_size(v: usize, spacer: &str, suffix: &str) -> String {
+    let v = v as f64;
+
+    const M: f64 = 1_000_000.0;
+    const K: f64 = 1_000.0;
+
     match v {
-        n if n > 10_000_000 => format!("{}{spacer}M{suffix}", n / 1_000_000),
-        n if n > 1_000_000 => format!(
-            "{}.{}{spacer}M{suffix}",
-            n / 1_000_000,
-            (n % 1_000_000) / 100_000
-        ),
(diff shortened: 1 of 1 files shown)


crates/cli/src/commands/zone.rs
```diff
@@ -978,14 +978,40 @@ impl std::fmt::Display for Icon {
...
 }
 
+#[cfg(test)]
+mod tests {
+    use super::format_size;
+
+    #[test]
+    fn test_format_size() {
+        assert_eq!(format_size(945, " ", "B"), "945 B");
+        assert_eq!(format_size(9450, " ", "B"), "9.4 KB");
```
bal-e, line 1007: This should really round to 9.5 -- perhaps you can use floating-point printing instead? It might also help elide the division+modulo logic.

bal-e (commented):
Looks great overall, just a tiny nit.

crates/cli/src/commands/zone.rs
tertsdiepraam: Did what you asked but it won't actually help in this very specific case:

> When truncating these values, Rust uses [round half-to-even](https://en.wikipedia.org/wiki/Rounding#Rounding_half_to_even), which is the default rounding mode in IEEE 754.

https://doc.rust-lang.org/stable/std/fmt/index.html

But the general case of rounding up still holds.

bal-e: Woah, TIL!

bal-e (approved):
Nice, LGTM :)

PR #796, opened 2026-06-10
Store a copy of SOA records with regular records

bal-e:
For @Philip-NLnetLabs -- makes it easier to access the data within the `zonedata` by reference.

---

### DIFF
--- a/crates/zonedata/src/reader.rs
+++ b/crates/zonedata/src/reader.rs
@@ -54,52 +54,27 @@ impl<'d> LoadedZoneReader<'d> {
 
     /// Regular records in the zone.
     ///
-    /// Records are sorted in DNSSEC canonical order. The SOA record **is not**
+    /// Records are sorted in DNSSEC canonical order. The SOA record **is**
     /// included.
     pub const fn regular_records(&self) -> &'d [RegularRecord] {
         self.instance.records.as_slice()
-    }
-
-    /// All records in the zone.
-    ///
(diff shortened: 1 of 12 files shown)


crates/zonedata/src/reader.rs
```diff
@@ -182,18 +157,18 @@ impl<'d> SignedZoneReader<'d> {
     /// Records are sorted in DNSSEC canonical order. Only records also present
     /// in the signed instance are included (the loaded SOA record, and loaded
     /// DNSKEY, RRSIG, CDS, CDNSKEY, ZONEMD records are excluded).
-    pub fn loaded_records(&self) -> impl Iterator<Item = RegularRecord> + Send + use<'d> {
-        LoadedZoneReader::new(self.loaded_instance).unsigned_records()
+    pub fn loaded_records(&self) -> impl Iterator<Item = &'d RegularRecord> + Send + use<'d> {
+        let soa = self.soa();
+        LoadedZoneReader::new(self.loaded_instance)
+            .unsigned_records()
+            .filter(|&r| r.rname != soa.rname || r.rtype != soa.rtype)
```
ximon18, line 164: Should that be &&?

Philip-NLnetLabs (approved):

crates/zonedata/src/reader.rs
bal-e: I think it's a matter of personal preference. `&` vs `&&` doesn't affect the logic, and I like to keep one layer of indirection present because it often helps with lifetime issues.

Issue #785, opened 2026-06-10
Getting `zone status` directly after `zone add` seems to hang


jpmens:
I've been meaning to mention this for days.

When I add a new zone with `cascade zone add` I typically fire off a `cascade zone status` for it directly after in order to verify the system has what I hope it has.

18 times out of 20, the `zone status` command "hangs" for many seconds, until I interrupt it (`SIGINT`), and issue it again, typically having to do that three or four times before I get an actual output.

This occurs on trivially short zones (like the ones I've been using over the weekend) as well as with larger ones.

Ideas on what I can do to provide more solid information?

I've only experienced these "hangs" of `zone status` directly after `zone add`, never at a later time.

ximon18:
Sounds like lock contention, i.e. a lock is being held while the status reporting is also trying to acquire the same lock. We will investigate.

bal-e:
With `ZoneStateLock` now integrated, this becomes quite trivial (at least, to identify whether `ZoneState` is the issue); we can just add some logging calls to its lock method. It's harder to add logging for its unlock method, but we can at least report how long locking blocks for.

jpmens:
The phenomenon is not new: I've been wanting to report it since I prepared for the NLUUG Cascade talk early May. Not 100% sure, but I believe I was running `cascade-0.1.0-alpha5-at-35ed2c2192` at the time.

PR #804, opened 2026-06-11
Persist zone fixes and improvements

ximon18:
This is a breaking change as it changes the layout of state.

---

This PR is part fix, part cleanup, part preparation for purging of diffs.

It:
- Fixes one minor bug (using hard-coded paths for restore instead of the paths persisted to state)
- Addresses the oversight that there was an existing persistence data structure which I previously failed to add the extra fields for persistence to.
- Moves in-memory IXFR diff management into its own type, making the persist and restore code more focused and moving verbose logging code out of the way of the main logic path making it easier to see what is happening.
- Switches from an array index based storage of in-memory diffs which could be prone to hard to spot bugs caused by storing or fetching loaded/signed diffs from the wrong indices, to an ordered BTreeMap based approach storing diffs by SOA serial and looking them up as needed.
- Store loaded SOA serials with persisted signed diff paths so that on restore we know which loaded SOA serial the restored signed diff relates to, instead of hoping that the last one stored for the zone is the related one.
- Logs not only the clients SOA serial on IXFR request but also the SOA serial of the zone at that point to make it easier from the logs to see if the correct diffs are being selected.

Note: The use of a BTreeMap is a bit overkill at this point, but is intended to be used in diff purging to be able to make it easy to identify and remove diffs by age. Simpler more efficient data structures could be used for the small number of diffs expected to be held in memory but the use of maps makes it easier to read in the code what is happening and the performance costs associated with the use of a map are not on the hot path, I don't think it will add a meaningful overhead at a critical point. However, it is of course up for discussion how "best" to do this.

---


src/persistence/mod.rs
```diff
@@ -45,8 +45,8 @@
 //!     `StorageState::diffs` so that it can be served in response to an IXFR
 //!     request from a downstream nameserver.
 //!   - The path that the diff file was written to is appended to
-//!     `ZoneState::persisted_loaded_diff_paths` or
-//!     `ZoneState::persisted_signed_diff_paths` and the zone state is
+//!     `ZoneState::persistence.loaded_diff_paths` or
+//!     `ZoneState::persistence.signed_diff_paths` and the zone state is
```
bal-e: You could make these intra-doc links e.g. ```[`PersistenceState::loaded_diff_paths`]``` so that we notice if they go stale :)

src/persistence/persist.rs
```diff
@@ -79,42 +81,16 @@ pub fn persist_loaded(
...
-                .as_ref()
-                .map(|soa_rr| soa_rr.rdata.serial),
-            loaded_only_diff
-                .0
-                .added_soa
-                .as_ref()
-                .map(|soa_rr| soa_rr.rdata.serial),
-        );
-        handle.state.storage.diffs.push(loaded_only_diff);
+        if loaded_diff.removed_soa.is_some() && loaded_diff.removed_soa != loaded_diff.added_soa {
```
bal-e, line 89: Why is a change in the SOA record required?
1. Is it because diffs are identified by their SOA serial number? If so, would this change once instance IDs are introduced (and can we leave a TODO comment here so we remember to fix that)?
2. Alternatively: is this check specifically looking for a change in the SOA serial number? If so, maybe the condition should be more specific. What if a field other than the SOA serial changes?

src/persistence/restore.rs
```diff
@@ -40,61 +40,72 @@ pub fn restore_loaded(
...
     restorer: &mut LoadedZoneRestorer,
 ) -> io::Result<bool> {
-    let mut state = zone.write(center);
-    if state.persisted_loaded_diff_paths.is_empty() {
-        return io::Result::Ok(false);
+    // Remove any existing diffs.
+    {
+        let mut state = zone.write(center);
+        state.storage.diffs.clear();
     }
```
bal-e: Instead of locking the zone state here: could this operation be moved into the caller of `restore_loaded`, i.e. `ZonePersistenceHandle::start_restore()`? It already has access to a locked `ZoneState`.

src/persistence/restore.rs
```diff
@@ -40,61 +40,72 @@ pub fn restore_loaded(
...
+    trace!(
+        "Restoring from paths: {:?}",
+        state.persistence.loaded_diff_paths
+    );
 
     // Determine the paths to read from. Each zone is persisted as an AXFR
     // plus zero or more IXFRs. The restorer takes a base path ending in an
     // unsigned integer number and loads that file plus N more, where the
     // final number in the path is replaced by the previous number plus one
     // each time.
```
bal-e: Just wondering: does this comment need to move somewhere else now?

src/persistence/restore.rs
```diff
@@ -40,61 +40,72 @@ pub fn restore_loaded(
...
+        state.storage.diffs.clear();
     }
 
-    info!("Restoring loaded zone from persisted data");
-    state.storage.diffs.clear();
+    let state = zone.read();
+    let mut paths_iter = state.persistence.loaded_diff_paths.iter();
+    let Some(snapshot_path) = paths_iter.next() else {
+        return io::Result::Ok(false);
+    };
```
bal-e, line 50: I have a few questions about this. It seems that this first element of `loaded_diff_paths` has a different meaning than all the rest. Should it be stored in a separate field in `PersistenceState`? Secondly, what does it mean if we hit this `else` case -- that the zone has never been persisted before? Should we log something here?

src/persistence/restore.rs
```diff
@@ -40,61 +40,72 @@ pub fn restore_loaded(
...
-    }
-
-    let mut source = loaded_source.to_path_buf();
     let mut all_serials = vec![];
+    let mut diffs_to_store: Vec<Arc<DiffData>> = vec![];
 
-    for i in 1..count {
+    for diff_path in paths_iter {
+        trace!(
+            "Loading and applying loaded diff form '{}'",
```
bal-e: typo: "form"

src/persistence/restore.rs
```diff
@@ -40,32 +40,38 @@ pub fn restore_loaded(
...
     // unsigned integer number and loads that file plus N more, where the
     // final number in the path is replaced by the previous number plus one
     // each time.
-    let loaded_source = center
-        .config
-        .zone_state_dir
-        .join(format!("{}.loaded.0", zone.name));
-    let count = state.persisted_loaded_diff_paths.len();
     let mut buf = Vec::<u8>::new();
-    drop(state);
```
bal-e, line 68: Why was `drop(state)` removed? I can't tell whether it was used further, but it would be nice to drop it before doing the expensive restoration work (then re-lock the zone state afterwards if needed).

bal-e: It looks like `snapshot_paths` and I guess `paths_iter` are borrowed from `state` so it needs to be retained; but those could be cloned out before dropping `state`. I worry locking `state` for too long could block Cascade's responsiveness at startup.

bal-e (commented):
Note: some of my comments might concern how some existing code works, even though the PR only made trivial changes to them. I probably just missed these when I reviewed the initial zone persistence PR.

src/persistence/restore.rs
ximon18: Indeed it is not really good to leave that there. I made some changes, please take a look. To some extent these changes may be obsoleted by the work in the purging PR.

src/persistence/restore.rs
ximon18: Given that it is only called from `start_restore()` which should only happen once on startup I assume it is actually redundant. Tests still pass with it removed so I'm just going to remove it.

src/persistence/persist.rs
ximon18: Here we are storing a diff to be served in response to future IXFR requests. By definition an IXFR diff has a removed SOA and an added SOA that should differ to each other. That is what we check for here.

src/persistence/restore.rs
ximon18: Yes, the 0th entry is a snapshot. I have a few times considered splitting out snapshots from diffs but was not convinced that it actually makes sense as a snapshot is also a diff, just a diff compared to the empty state, and so does not have a different meaning than the rest, though with the zone builder/replacer code it must use different code to restore an initial diff (aka snapshot) than a subsequent diff.

The else case just means there is nothing to restore but that can be normal, e.g. the zone was added but loaded review was not approved so nothing was ever persisted. That isn't a wrong case that needs to be logged.

src/persistence/restore.rs
ximon18: I think the drop was removed just because it wasn't needed anymore. Only a read lock is taken so operations that query the zone state can still succeed, and no write operations should happen on the zone while it is restoring.

bal-e: There are many cases where the zone state is locked mutably, and they don't always match with high-level expectations of what a "write operation on a zone" should be. I'm worried this will lead to more unnecessary delays when users try to interact with Cascade (e.g. #785).

ximon18: Hmm, I hadn't considered it that way. Let me see if I can remove it. The zone state machinery should prevent any unexpected attempts to change the state anyway as it is in state restoring which is not the right state for other actions to occur on the zone.

ximon18: Please re-check it now.

bal-e (approved):
Let's go!!

Issue #514, opened 2026-03-10
Inconsistent CLI output.


ximon18:
Some commands give quite verbose output:
```
$ cascade policy reload
Policies reloaded:
  - default unchanged
  - nsec    added

$ cascade status
Signing queue:
  The signing queue is currently empty.
```

Some commands just state what they did.
```
$ cascade zone add ...
Added zone example
```

Others give an indication of success.
```
$ cascade zone reload example
Success: Sent zone reload command for example
```

And some don't say much at all if there isn't anything to report.
```
$ cascade zone list

$ cascade policy list

```

This is all a bit inconsistent.

PR #807, opened 2026-06-12
Print a message in list commands when no items can be shown

tertsdiepraam:
I chose to use `eprintln` because I don't want this to be filtered out with tracing levels. I think we should probably make it so that the CLIs error printing does not rely on tracing, that feels kind of wrong.

This PR is related to https://github.com/NLnetLabs/cascade/issues/514

### DIFF
--- a/crates/cli/src/commands/policy.rs
+++ b/crates/cli/src/commands/policy.rs
@@ -8,7 +8,7 @@ use crate::{
         SignerSerialPolicyInfo,
     },
     client::CascadeApiClient,
-    println,
+    eprintln, println,
 };
 
 #[derive(Clone, Debug, clap::Args)]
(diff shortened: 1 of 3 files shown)


ximon18:
I think I agree about printing vs tracing in the client. The only case I can imagine for logs/tracing is HTTP connectivity / response parsing issues, but I guess that should just bubble up into a user visible printed error.

crates/cli/src/commands/policy.rs
```diff
@@ -38,6 +38,10 @@ impl Policy {
             PolicyCommand::List => {
                 let res: PolicyListResult = client.get_json("policy/").await?;
 
+                if res.policies.is_empty() {
```
ximon18, line 41: This is inconsistent with `tsig list` which can show "none" for related policies and/or zones. Should we change this to "none" or make `tsig list` say more than "none", assuming you agree that this is inconsistent? I'm inclined to say we should use "none" everywhere.

tertsdiepraam: It's slightly inconsistent, but the `none` is shown as `zones: none`, so the context is clear there. Also this gets printed to stderr, while the `none`s get printed to stdout.

ximon18: Okay, I'm not convinced but let's not block this PR on this point, I'll resolve this conversation.

ximon18 (approved):

PR #810, opened 2026-06-12
Return some proper JSON for the health api endpoint

### LINKED ISSUE #375: cascaded `health` endpoint returns `null`
I read [in the source](https://github.com/NLnetLabs/cascade/blob/44b3b80c6887a6dd9baf4b7e4cbe121a88821b38/src/units/http_server.rs#L145) that 

> if this endpoint responds, the daemon is considered healthy

but wouldn't emiting something a bit more positive look a bit more healthy? ;-)

```console
$ curl -sSf http://localhost:4539/health | jq .
null
```

Maybe a `true` for healthy, or even 

```json
{ "healthy" : true }
```

tertsdiepraam:
Closes #375

Instead of returning `null`/`()` we now return `{ healthy: true }`. This makes it a bit clearer what the response means and allows us to extend it with more information in the future (e.g. maybe check for dnst keyset and stuff like that).

### DIFF
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -71,7 +71,7 @@ impl Command {
                     println!("Ok");
                 } else {
                     // This path is unreachable in practice at the moment.
-                    println!("Something's wrong!")
+                    println!("The Cascade daemon is reachable but reports itself as unhealthy.");
                 }
                 Ok(())
             }
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -66,8 +66,13 @@ impl Command {
         match self {
             Self::Debug(cmd) => cmd.execute(client).await,
             Self::Health => {
-                client.get_json::<()>("health").await?;
-                println!("Ok");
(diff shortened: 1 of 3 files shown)


crates/cli/src/commands/mod.rs
```diff
@@ -66,8 +66,13 @@ impl Command {
...
             Self::Debug(cmd) => cmd.execute(client).await,
             Self::Health => {
-                client.get_json::<()>("health").await?;
-                println!("Ok");
+                let health = client.get_json::<cascade_api::Health>("health").await?;
+                if health.healthy {
+                    println!("Ok");
+                } else {
+                    // This path is unreachable in practice at the moment.
+                    println!("Something's wrong!")
```
ximon18: Will we regret this message later if the branch does become reachable and we forget to update this part? We know at least that the issue isn't with the client, and that the server is up and reachable and was able to understand our request and return a response, so we can be more specific and say at least "The Cascade daemon is reachable but reports itself as unhealthy" or something?

tertsdiepraam: Yeah that makes sense!

ximon18 (approved):

PR #815, opened 2026-06-15
add basic info command to get daemon version

### LINKED ISSUE #803: `cascade --version` prints client's version even when `--server` is specified
Using

```console
$ cascade -s 192.168.1.140:4539 -V
```

I was hoping to get the server's version, i.e. the Cascaded version on `192.168.1.140:4539`, but I get the locally installed version of the host running the `cascade` client.

The `health` subcommand doesn't give any useful information about that either:

```
$ cascade -s 192.168.1.140:4539 health
Ok
```

I think `-V` is useful, but not really in combination with `-s`, and maybe `health` can issue a bit of JSON along the lines of

```json
{
  "healthy": true,
  "version": "0.1.0-beta2-dev",
  "rev": "db315a7ee7"
}
```

Related to #337

tertsdiepraam:
Closes https://github.com/NLnetLabs/cascade/issues/803.

Add an `info` command that prints:
```
Cascade daemon version: 0.1.0-beta3-dev at c9a03b3a11-dirty
```

It uses the same string as `--version` but for the server instead of the client.

Some background on internal discussion:
- `health` should be used for load balancers and monitoring, not for getting the version.
- `status` doesn't feel right either, since that describes more what Cascade is doing.
- Therefore, a new command feels appropriate.

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -764,6 +764,11 @@ pub struct Health {
 }
 
 #[derive(Deserialize, Serialize, Debug, Clone)]
+pub struct Info {
+    pub version: String,
+}
+
+#[derive(Deserialize, Serialize, Debug, Clone)]
 pub struct ServerStatusResult {
     pub halted_zones: Vec<(ZoneName, String)>,
     pub signing_queue: Vec<SigningQueueReport>,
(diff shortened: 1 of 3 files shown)


ximon18:
An alternative way to do this would be to report both the client **and** the server version (if available) from `--version`.

ximon18 (approved):

PR #817, opened 2026-06-15
Add `CASCADE_DAEMON` env var for the CLI

### LINKED ISSUE #801: Can `IP:PORT` for `cascade -s` be specified in an environment variable?
The `cascade` utility has an option `-s` (or `--server`) which expects `<IP:PORT>`. I do not see we can specify that in an environment variable or configuration file.

I've not found anything in the docs nor in `strings cascade` but might well have missed it.

I think it would be a good idea to have, something along the lines of

```sh
$ export CASCADE_DAEMON="192.0.2.42:4539"
```

(I explicitly didn't use `CASCADE_SERVER` since that is one of the variables passed from cascaded to the review scripts and it might cause confusion.)

tertsdiepraam:
Closes #801

### DIFF
--- a/crates/cli/src/args.rs
+++ b/crates/cli/src/args.rs
@@ -16,6 +16,7 @@ pub struct Args {
         short = 's',
         long = "server",
         value_name = "IP:PORT",
+        env = "CASCADE_DAEMON",
         default_value = "127.0.0.1:4539",
         global = true
     )]
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -39,7 +39,7 @@ path = "../api"
 # The CLI delegates to Clap to handle basic concerns like argument parsing.
 [dependencies.clap]
 workspace = true
-features = ["cargo", "derive", "wrap_help"]
+features = ["cargo", "derive", "env", "wrap_help"]
 
 # TODO: Evaluate this choice.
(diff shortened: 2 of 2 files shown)


crates/cli/src/args.rs
```diff
@@ -9,30 +9,14 @@ use super::client::CascadeApiClient;
...
-        display_order = 100,
-    )]
-    pub version: (),
-
     /// The cascade server instance to connect to
     #[arg(
         short = 's',
         long = "server",
         value_name = "IP:PORT",
+        env = "CASCADE_DAEMON",
```
ximon18, line 19: Given that the command line argument name is `server` wouldn't `CASCADE_SERVER` be a better name here, and also allow us to name other env vars later (if any) also after their respective command line argument names?

ximon18: Ah I see why `CASCADE_SERVER` wasn't used, as mentioned on #801.

ximon18 (approved):

crates/cli/src/args.rs
tertsdiepraam: JP suggested `CASCADE_DAEMON` because `CASCADE_SERVER` is also set by the review script, which made sense to me. Happy to be persuaded to use `CASCADE_SERVER` though.

PR #820, opened 2026-06-15
Support importing keys by private key path as well as public key path (fixes #717).

### LINKED ISSUE #717: Importing key with `zone add` requests `--private-key` which doesn't exist

The [documentation](https://github.com/NLnetLabs/cascade/blob/main/doc/manual/source/key-management.rst) specifies on importing public/private key files:

> It is sufficient to give the name of the file that holds the public key if the filename ends in .key and the filename of the private key is the same except that it ends in .private. If this is not the case then the private key filename must be specified separately

but this doesn't appear to be the case:

```console
$ dnssec-keygen -a13 e1.aa.    # generate key pair (BIND util)
$ ls -l Ke*
-rw-r--r-- 1 jpm jpm 330 Jun  8 10:20 Ke1.aa.+013+27233.key
-rw------- 1 jpm jpm 187 Jun  8 10:20 Ke1.aa.+013+27233.private


$ cascade zone add --source ../e1.aa --policy 13 --import-csk-file Ke1.aa.+013+27233.private e1.aa
2026-06-08T10:24:27.358124Z ERROR cascade: Failed to add zone: Zone registration failed: Keyset command '/home/jpm/.cargo/bin/dnst keyset -c /home/jpm/cascade/keys/e1.aa.cfg import csk file Ke1.aa.+013+27233.private' returned non-zero exit code: exit status: 1 [stdout=, stderr=[/home/jpm/.cargo/bin/dnst] ERROR: public key Ke1.aa.+013+27233.private should end in .key, use --private-key to specify a private key separately


$ cascade zone add --source ../e1.aa --policy 13 --import-csk-file Ke1.aa.+013+27233.key --private-key Ke1.aa.+013+27233.private  e1.aa
error: unexpected argument '--private-key' found

  tip: to pass '--private-key' as a value, use '-- --private-key'

Usage: cascade zone add --source <SOURCE> --policy <POLICY> --import-csk-file <I
MPORT_CSK_FILE> <NAME>

For more information, try '--help'.


$ cascade zone add --source ../e1.aa --policy 13 --import-csk-file Ke1.aa.+013+27233.key e1.aa
Zone e1.aa scheduled for loading, use 'cascade zone status e1.aa' to see the status.
```

The mentioned option`use --private-key to specify a private key separately` doesn't exist in zone add.

In this example, specifying the public key sufficed as the private portion is in `*.private` (which agrees with that part of the documentaton)

Related to #713

ximon18:
Determines the path to the "other" half of the key and passes both public and private paths to `dnst keyset` explicitly, rather than just the public path as the latter can cause `dnst keyset` to produce a confusing error that the user specify `--private-key`, an argument not offered by `cascade zone add`.

Provide your description here

---

### DIFF
--- a/crates/cli/src/commands/zone.rs
+++ b/crates/cli/src/commands/zone.rs
@@ -1064,9 +1064,9 @@ fn key_file_imports(
     Ok(key_imports)
 }
 
-// This is not done as impl TryFrom<Utf8PathBuf> for FileKeyImport as
-// Utf8PathBuf is not defined in this crate and Rust doesn't allow impl for a
-// type defined in another crate.
+// This is not done as impl TryFrom<Utf8PathBuf> for FileKeyImport as neither
+// Utf8PathBuf nor FileKeyImport are not defined in this crate and Rust
+// doesn't allow impl for a type defined in another crate.
 fn expand_key_path(key_file_path: Utf8PathBuf, key_type: KeyType) -> Result<FileKeyImport, String> {
     // Is the given path to a .key file or a .private file? We need both
     // so generate the other one from the one given. Note that we don't
--- a/crates/cli/src/commands/zone.rs
+++ b/crates/cli/src/commands/zone.rs
@@ -192,24 +192,9 @@ impl Zone {
                 import_csk_kmip,
             } => {
                 let import_public_key = import_public_key.into_iter().map(KeyImport::PublicKey);
-                let import_ksk_file = import_ksk_file.into_iter().map(|p| {
-                    KeyImport::File(FileKeyImport {
-                        key_type: KeyType::Ksk,
-                        path: p,
-                    })
-                });
-                let import_csk_file = import_csk_file.into_iter().map(|p| {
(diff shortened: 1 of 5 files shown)


crates/cli/src/commands/zone.rs
```diff
@@ -1068,6 +1053,50 @@ fn format_duration(duration: Duration) -> String {
...
+) -> Result<Vec<KeyImport>, String> {
+    let mut key_imports = Vec::with_capacity(key_paths.len());
+    for key_path in key_paths {
+        key_imports.push(KeyImport::File(expand_key_path(key_path, key_type)?));
+    }
+    Ok(key_imports)
+}
+
+// This is not done as impl TryFrom<Utf8PathBuf> for FileKeyImport as
+// Utf8PathBuf is not defined in this crate and Rust doesn't allow impl for a
```
Philip-NLnetLabs: That should say that neither Utf8PathBuf nor FileKeyImport is defined in this crate.

Philip-NLnetLabs (approved):

Issue #786, opened 2026-06-10
thread 'cascade-worker' (104843) panicked at src/zone/machine.rs


jpmens:
I didn't do anything, officer, I swear! I was adding a text to an issue! ;-)

`cascade 0.1.0-beta2-dev at 10ffc35931`

```console

$ /home/jpm/.cargo/bin/cascaded -c config.toml --state state.db -l file:/tmp/cascaded.log

thread 'cascade-worker' (104843) panicked at src/zone/machine.rs:153:13:
The storage was in the passive state but the state machine wasn't in the waiting state
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

at this point I was looking at some `zone status` and `zone history` as well as `tsig list`:

```
$ RUST_BACKTRACE=1 /home/jpm/.cargo/bin/cascaded -c config.toml --state state.db -l file:/tmp/cascaded.log
thread 'cascade-worker' (105070) panicked at src/zone/machine.rs:153:13:
The storage was in the passive state but the state machine wasn't in the waiting state
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: cascaded::zone::machine::<impl cascaded::zone::ZoneHandle>::try_start_resign 
   3: cascaded::signer::zone::SignerZoneHandle::enqueue_resign
   4: cascaded::units::key_manager::KeyManager::run::{{closure}}
   5: tokio::runtime::task::harness::Harness<T,S>::poll
   6: tokio::runtime::scheduler::multi_thread::worker::Context::run_task
   7: tokio::runtime::scheduler::multi_thread::worker::Context::run
   8: tokio::runtime::context::runtime::enter_runtime
   9: tokio::runtime::scheduler::multi_thread::worker::run
  10: tokio::runtime::task::core::Core<T,S>::poll
  11: tokio::runtime::task::harness::Harness<T,S>::poll
  12: tokio::runtime::blocking::pool::Inner::run
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

```
$ RUST_BACKTRACE=full /home/jpm/.cargo/bin/cascaded -c config.toml --state state.db -l file:/tmp/cascaded.log > /tmp/full 2>&1

$ cat /tmp/full

thread 'cascade-worker' (105276) panicked at src/zone/machine.rs:153:13:
The storage was in the passive state but the state machine wasn't in the waiting state
stack backtrace:
[...]

PR #821, opened 2026-06-15
[zone/machine] Check more state before starting a new operation

bal-e:
If the zone is in `SigningFailed` state due to a re-signing failure, the zone data storage state machine is `Passive`. This breaks an assumption made in `try_start_{load,resign}()`. This commit adds a check that the zone is in `Waiting` state before trying the zone data storage state machine.

Should help with #786.

---

### DIFF
--- a/src/zone/machine.rs
+++ b/src/zone/machine.rs
@@ -405,11 +405,16 @@ impl<'a> ZoneHandle<'a> {
             }
             ZoneStateMachine::SigningFailed(signing_failed) => {
                 let waiting = signing_failed.reset();
+                transition.move_to(ZoneStateMachine::Waiting(waiting));
 
                 // TODO: This should be handled by 'Instances'.
                 self.state.next_min_expiration = None;
 
-                transition.move_to(ZoneStateMachine::Waiting(waiting));
+                // The signing operation has already been abandoned, so the zone
+                // data storage is already passive. Its call to `on_passive()`
+                // was ignored because the zone state machine was busy at the
+                // time. Call it again now.
+                self.storage().on_passive();
             }
             _ => {
                 transition.move_to(state);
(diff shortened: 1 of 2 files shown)


src/zone/machine.rs
```diff
@@ -401,11 +405,16 @@ impl<'a> ZoneHandle<'a> {
...
 
                 // TODO: This should be handled by 'Instances'.
                 self.state.next_min_expiration = None;
 
-                transition.move_to(ZoneStateMachine::Waiting(waiting));
+                // The signing operation has already been abandoned, so the zone
+                // data storage is already passive. Its call to `on_passive()`
+                // was ignored because the zone state machine was busy at the
+                // time. Call it again now.
+                self.storage().on_passive();
```
tertsdiepraam, line 417: I love that I wanted to suggest adding a `Waiting` check to `on_passive` and then scroll and yup, there it is.

src/zone/storage.rs
```diff
@@ -860,14 +862,20 @@ impl StorageZoneHandle<'_> {
...
+
         if self.zone().loader().start_pending() {
             // The zone is no longer passive.
             return;
         }
 
         if self.zone().signer().start_pending() {
             // The zone is no longer passive.
-            // return;
+            //return;
```
tertsdiepraam, line 878: Ok 😄

tertsdiepraam (approved):

tertsdiepraam:
Oh, note that GH has linked the issue even though you wrote "might". If you don't want to close the issue, unlink that first.

src/zone/storage.rs
bal-e: lol, sorry, just something I had meant to do a while back.

bal-e:
> Oh, note that GH has linked the issue even though you wrote "might". If you don't want to close the issue, unlink that first.

thanks for the heads up!

PR #824, opened 2026-06-17
Improve incremental rrsigs

Philip-NLnetLabs:
Reduce memory consumption for incremental signing by keeping RRSIG in new base.
---


crates/zonedata/src/lib.rs
```diff
@@ -244,6 +261,15 @@ impl DerefMut for RegularRecord {
...
 
+impl Hash for RegularRecord {
+    fn hash<H>(&self, h: &mut H)
+    where
+        H: Hasher,
+    {
+        self.0.hash(h);
+    }
+}
+
```
bal-e: I think a simple `#[derive(Hash)]` will do? Or am I missing something?

src/signer/incremental.rs
```diff
@@ -2279,6 +2284,379 @@ impl IncrementalSigningState {
...
+        }
+    }
+
+    fn add_existing_record(&mut self, record: &'zd RegularRecord) {
+        let NewRecordData::Rrsig(rrsig) = record.data() else {
+            panic!("ZoneRecordData::Rrsig expected");
+        };
+        let key = (record.owner(), rrsig.type_covered());
+        self.old_rrsigs.entry(key).or_default().push(record);
+    }
```
bal-e, line 2312: Nice, I like this!

src/signer/incremental.rs
```diff
@@ -2279,6 +2284,379 @@ impl IncrementalSigningState {
...
+        // First check the changes map.
+        match self.changes.entry(buf_key) {
+            hash_map::Entry::Occupied(mut entry) => {
+                let change = entry.get_mut();
+                match change {
+                    RrsigChange::Delete { .. } => {
+                        // Create a new empty change, either insert of
+                        // modify depending on the existing RRSIGS.
+                        let key = (record.owner(), rrsig.type_covered());
+                        let new_change = if let Some(rrsigs) = self.old_rrsigs.get(&key) {
```
bal-e: Doesn't the `change` you just matched on already provide these RRSIGs? That way you don't need the extra hashmap lookup, nor deal with the possibility of `None`.

This could lead to ownership issues (i.e. you need to move out `change` but it's a mutable reference); we can take a look together sometime.

src/signer/incremental.rs
```diff
@@ -2279,6 +2284,379 @@ impl IncrementalSigningState {
...
+                        };
+                        *change = new_change;
+                    }
+                    RrsigChange::Modified { new, .. } | RrsigChange::Insert { new, .. } => {
+                        new.push(record);
+                    }
+                }
+            }
+            hash_map::Entry::Vacant(entry) => {
+                // Whether to new change is Modified or Insert depends on
```
bal-e: typo: "to"?

src/signer/incremental.rs
```diff
@@ -2279,6 +2284,379 @@ impl IncrementalSigningState {
...
+                } else {
+                    // Nothing yet. Create an Insert.
+                    RrsigChange::Insert { new: vec![record] }
+                };
+                entry.insert(change);
+            }
+        }
+    }
+
+    fn insert_new_records(&mut self, records: Vec<RegularRecord>) {
```
bal-e: I guess these records are supposed to be part of the same RRset (and cover the same RTYPE)? Maybe that could be documented here, and a `debug_assert!` added to verify the condition (rough guess: `debug_assert!(records.iter().all(|r| r.owner() == &buf_key.0 && matches!(r.data(), NewRecordData::Rrsig(rrsig) if rrsig.type_covered() == buf_key.1))`).

src/signer/incremental.rs
```diff
@@ -2279,6 +2284,379 @@ impl IncrementalSigningState {
...
+                                old: rrsigs.to_vec(),
+                                new: records,
+                            }
+                        } else {
+                            todo!();
+                        };
+                        *change = new_change;
+                    }
+                    RrsigChange::Modified { new, .. } | RrsigChange::Insert { new, .. } => {
+                        *new = records;
```
bal-e, line 2386: This behavior suggests that `records` is _replacing_ the entire set of RRSIGs covering a particular RRset. Maybe the function name could reflect that?

src/signer/incremental.rs
```diff
@@ -2279,6 +2284,379 @@ impl IncrementalSigningState {
...
+                                old: rrsigs.to_vec(),
+                                new: vec![record],
+                            }
+                        } else {
+                            todo!();
+                        };
+                        *change = new_change;
+                    }
+                    RrsigChange::Modified { new, .. } | RrsigChange::Insert { new, .. } => {
+                        new.push(record);
```
bal-e, line 2334: Based on this, I guess the `new` list of records is unordered? Do you want to document that in the definition of `RrsigChange`?

src/signer/incremental.rs
```diff
@@ -2279,6 +2284,379 @@ impl IncrementalSigningState {
...
+                RrsigChange::Insert { .. } => 1,
+            })
+            .sum();
+
+        let ilen = (len as isize) + changes;
+        if ilen >= 0 {
+            ilen as usize
+        } else {
+            unreachable!()
+        }
```
bal-e: FYI, there is a `usize::checked_add_signed()` method that does this for you! You could replace this code with `len.checked_add_signed(changes).unwrap()` :)

src/signer/incremental.rs
```diff
@@ -2279,6 +2284,379 @@ impl IncrementalSigningState {
...
+                entry.insert(new_change);
+            }
+        }
+    }
+
+    fn values(&self) -> RrsigsValuesIter<'_> {
+        RrsigsValuesIter::new(self.old_rrsigs.iter(), &self.changes)
+    }
+
+    fn len(&self) -> usize {
```
bal-e: My understanding is, this method computes the number of signed RRsets in the new instance of the zone. I'm a little worried `len()` might get used accidentally in the wrong context. Could you document these semantics and/or make the name more specific?

src/signer/incremental.rs
```diff
@@ -2279,6 +2284,379 @@ impl IncrementalSigningState {
...
+}
+
+type RrsigIterItem<'a> = ((&'a RevName, &'a NewRtype), Vec<&'a RegularRecord>);
+impl<'a> IntoIterator for &'a Rrsigs<'a> {
+    type Item = RrsigIterItem<'a>;
+    type IntoIter = RrsigIter<'a>;
+    fn into_iter(self) -> <Self as IntoIterator>::IntoIter {
+        RrsigIter::new(self.old_rrsigs.iter(), &self.changes)
+    }
+}
```
bal-e, line 2505: Nice, this makes things super convenient!

src/signer/incremental.rs
```diff
@@ -2279,6 +2284,379 @@ impl IncrementalSigningState {
...
+            result
+        }
+    }
+
+    fn iter(&self) -> RrsigIter<'_> {
+        RrsigIter::new(self.old_rrsigs.iter(), &self.changes)
+    }
+}
+
+type RrsigIterItem<'a> = ((&'a RevName, &'a NewRtype), Vec<&'a RegularRecord>);
```
bal-e: I see `(&'a RevName, &'a NewRtype)` is being used in some places, in other places `(&'a RevName, NewRtype)`. Is there a chance these could be more consistent (maybe changing everything into the latter)? Then you could also add a type alias for it.

src/signer/incremental.rs
```diff
@@ -2279,6 +2284,379 @@ impl IncrementalSigningState {
     }
 }
 
+struct Rrsigs<'zd> {
+    // Store the existing RRSIG records. The incremental signer doesn't
+    // need them, but we need them to inform the zone store which records
+    // have been removed.
+    old_rrsigs: HashMap<(&'zd RevName, NewRtype), Vec<&'zd RegularRecord>>,
+
+    changes: HashMap<(RevNameBuf, NewRtype), RrsigChange<'zd>>,
```
bal-e: FYI `Box<RevName>` would save memory over `RevNameBuf`. You can copy `&RevName` into a `Box<RevName>` using `revname.unsized_copy_into()`.

src/signer/incremental.rs
```diff
@@ -2279,6 +2284,379 @@ impl IncrementalSigningState {
...
+                        RrsigChange::Modified { new, .. } | RrsigChange::Insert { new, .. } => {
+                            let new: Vec<&RegularRecord> = new.iter().collect();
+                            return Some(((name.as_ref(), rtype), new));
+                        }
+                    }
+                }
+
+                // End of iterator.
+                break;
+            }
```
bal-e: BTW you can use `while let Some(((name, rtype), change)) = changes_iter.next() { ... }` to simplify this pattern.

src/signer/incremental.rs
```diff
@@ -2279,6 +2284,379 @@ impl IncrementalSigningState {
...
+            loop {
+                let next = changes_iter.next();
+                if let Some(((name, rtype), change)) = next {
+                    match change {
+                        RrsigChange::Delete { .. } => {
+                            // Nothing here.
+                            continue;
+                        }
+                        RrsigChange::Modified { new, .. } | RrsigChange::Insert { new, .. } => {
+                            let new: Vec<&RegularRecord> = new.iter().collect();
```
bal-e: It's unfortunate that you need to create a whole `Vec`, but I can see how many problems with references that would cause. Oh well.

src/signer/incremental.rs
bal-e: Likewise for the loop before this.

bal-e (commented):
The need for so many conversions between old and new base is unfortunate but expected.

I really like `Rrsigs`, it helps clarify my understanding of the incremental signer logic a lot. I'm excited to see these changes propagate to NSECs and beyond.

Note that there are some `todo!()`s in the code still.

src/signer/incremental.rs
Philip-NLnetLabs: In theory we could have an enum that either takes a few RegularRecords or a Vec. Not sure if it is worth the trouble. Certainly not for the the new/modified signatures.

src/signer/incremental.rs
Philip-NLnetLabs: After changing the loop, clippy found a way to turn it into a for loop.

src/signer/incremental.rs
Philip-NLnetLabs: done

src/signer/incremental.rs
Philip-NLnetLabs: Done

src/signer/incremental.rs
Philip-NLnetLabs: len -> signed_rrset_count

src/signer/incremental.rs
Philip-NLnetLabs: Nice.

src/signer/incremental.rs
Philip-NLnetLabs: I renamed the method to replace_with_new_records.

src/signer/incremental.rs
Philip-NLnetLabs: Done

src/signer/incremental.rs
Philip-NLnetLabs: Done

src/signer/incremental.rs
Philip-NLnetLabs: Nice. This was code from an early version where Delete didn't have old yet. Now it is much cleaner.

crates/zonedata/src/lib.rs
Philip-NLnetLabs: Completely forgot about that possibility. And clippy doesn't seem to check for this either.

Philip-NLnetLabs:
I fixed to todos. The conversion between old and new base should mostly go away when everything is stored as new base. Though crypto is only in old base.

src/signer/incremental.rs
bal-e: I think it's very hard to do better than the current code, so I don't think anything needs to be changed.

src/signer/incremental.rs
bal-e: awesome!

bal-e (approved):

Issue #764, opened 2026-06-09
Move important logic from `zone/storage.rs` to `zone/machine.rs`


bal-e:
Before the pipeline state machine was introduced, the storage state machine incurred a lot of important logic for progressing Cascade. We should take a good look at the logic in `zone/storage.rs` and decide what needs to move to `zone/machine.rs`.

Issue #766, opened 2026-06-09
Heavily simplify `zonedata`


bal-e:
The `cascade-zonedata` crate is quite complex, but this complexity allows it to offer one feature: wait-free access to the published instance of the zone. I thought this would be important for zone serving. But due to Cascade's position as a hidden primary, @Philip-NLnetLabs mentioned that we don't need this feature; it is okay for Cascade to block for a little while when serving XFRs. Thus, a lot of the complexity in `cascade-zonedata` is unnecessary; we can simplify it, reducing code, removing `unsafe` blocks, and making it more maintainable.

Proposal 1 for the new API: track the published instance of the zone as `RwLock<ZoneData>` and the upcoming one similarly (or maybe pass it around directly; not sure). The upcoming instance can be stored as a diff against the published instance; this diff can be applied transparently by the zone server when serving XFRs. Once an upcoming instance is approved, the published instance is locked and the diff is applied to it; this may take a short while.

Proposal 2 for the new API: track the published instance of the zone akin to `ArcSwap<ZoneData>` (although an `RwLock` would be fine, I'd rather avoid the dependency). Store the upcoming instance as a new `ZoneData`, which can be served directly by the zone server. Once the upcoming instance is approved, just replace the old instance.

In either case, the zone data storage state machine can go, as can various transitions.

PR #841, opened 2026-06-23
Move pipeline logic out of `zone/storage.rs`

bal-e:
Before `zone/machine.rs` was introduced, `zone/storage.rs` served as Cascade's de facto state machine, and accumulated important logic and control flow. This PR moves it into `zone/machine.rs` so that it is in the right place. These changes should not affect Cascade's behavior whatsoever.

This PR moves the management of the `*_soa` fields to `zone/machine.rs`. These will be replaced by #617, which now depends on this PR.

Solves #764. There is still important work happening in `zone/storage.rs` that should happen elsewhere (specifically communication with review servers), but that needs to be factored out separately.

---

### DIFF
--- a/crates/zonedata/src/builder.rs
+++ b/crates/zonedata/src/builder.rs
@@ -19,6 +19,7 @@ use crate::{
 /// [`LoadedZoneBuilder`] is used to load new instances of a zone from external
 /// sources (e.g. zonefiles and DNS servers). It offers read-only access to the
 /// current loaded instance of the zone, to support incremental loading.
+#[must_use = "Dropping this will lock up the zone data storage"]
 pub struct LoadedZoneBuilder {
     /// The underlying data.
     data: Arc<Data>,
@@ -285,6 +286,7 @@ impl fmt::Debug for LoadedZoneBuilder {
 /// re-sign existing signed instances. It offers read-only access to the current
 /// (loaded and signed) instance of the zone, and (if any) the new loaded
 /// instance of the zone, to support incremental signing.
+#[must_use = "Dropping this will lock up the zone data storage"]
 pub struct SignedZoneBuilder {
     /// The underlying data.
     data: Arc<Data>,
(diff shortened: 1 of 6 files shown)


src/persistence/zone.rs
```diff
@@ -109,7 +110,17 @@ impl ZonePersistenceHandle<'_> {
...
+                let (loaded_reviewer, signed_reviewer, viewer) =
+                    handle.storage().finish_signed_restoration(restored);
+
+                // Register the zone against the zone servers.
+                LoadedReviewServer::add_zone(handle.center, handle.zone.clone(), loaded_reviewer);
+                SignedReviewServer::add_zone(handle.center, handle.zone.clone(), signed_reviewer);
+                PublicationServer::add_zone(handle.center, handle.zone.clone(), viewer);
+
+                // Send a notification that the state machine is now passive.
+                handle.storage().on_passive();
```
tertsdiepraam, line 122: I wonder whether this should become `on_ready` or something on the zone state machine. Not for now though, this is fine.

crates/zonedata/src/builder.rs
```diff
@@ -732,21 +738,138 @@ impl fmt::Debug for SignedZoneBuilder {
...
+impl LoadedZoneBuilt {
+    /// The new (loaded) instance.
+    ///
+    /// If a new instance instance has been built (with [`Self::replace()`] or
+    /// [`Self::patch()`]), it can be accessed here. Note that empty instances
+    /// (as built by [`Self::clear()`]) cannot be accessed.
+    pub fn next(&self) -> Option<LoadedZoneReader<'_>> {
+        // SAFETY: As per the caller, 'loaded[index]' will not be
+        // accessed elsewhere for the lifetime of 'self', and so is sound to
+        // access immutably.
```
tertsdiepraam, line 761: I think your habit of writing "as per the caller" is a bit to be honest. The caller in my mind is the function, but it's more a guarantee of the type here. It also doesn't really explain how the type enforces this requirement.

In this case, it might be better to say something like "Since LoadedZoneBuilt can only be constructed if ..."?

Anyway, probably not worth updating all the occurrences of this. Just wanted to note it going forward. And maybe it's just something for me to get used to.

tertsdiepraam (approved):

tertsdiepraam:
Looks good. CI only fails because of the docs, so once that's fixed this can be merged.

crates/zonedata/src/builder.rs
bal-e: You're totally right, I don't like these comments either. The current situation has arisen because I copied the safety comments from the constructors and never got back around to fixing them. It's a bit of a mess, to be honest ... I think the next time I really tackle this code will be #766.

PR #853, opened 2026-07-01
Add some documentation about how key management works between Cascade, dnst and Cascade-HSM-Bridge.

ximon18:
I had to do a bit of a deep dive to remind myself exactly what is handled where and made some notes. I think the notes are useful, but don't know where to keep them. I decided to add them to the docs at the top of the key manager unit and discovered there are no docs there, so I added some.

### DIFF
--- a/src/units/key_manager.rs
+++ b/src/units/key_manager.rs
@@ -116,16 +116,17 @@
 //            KMIP_Activate
 //   3. Relabel to: <prefix>-<(partial) zone name>-<key tag>-<key type>.
 //        if supports_relabeling:
-//          if !conn.rename_key():
-//            supports_relabeling = false
-//              format_key_label()
-//              conn.rename_key(public_key_id, new_name):
-//                KMIP_ModifyAttribute(public_key_id, Attribute::Name(new_name)):
-//              if failed:
-//                supports_relabelling = false
-//              else:
-//                conn.rename_key(private_key_id, new_name):
-//                  KMIP_ModifyAttribute(private_key_id, Attribute::Name(new_name))
+//          # note: format_key_label() falls back to a hex string if it
+//          # cannot fit the values in the max label len allowed.
+//          pub_label = format_key_label(prefix, zone name, key tag, key type, "-pub", max_label_bytes)
+//          pri_label = format_key_label(prefix, zone name, key tag, key type, "-pri", max_label_bytes)
+//          conn.rename_key(public_key_id, pub_label):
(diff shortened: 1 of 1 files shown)


src/units/key_manager.rs
```diff
@@ -1,3 +1,157 @@
...
+//! confusing for Cascade operators. If the output of `dnst` is not suitable
+//! for consumption by Cascade operators it must be interpreted/massaged by
+//! Cascade. This is not ideal.
+//!
+//! Examples of this can be seen in `on_get_key()`, `on_remove_key()`,
+//! `on_roll_key()` and `on_status()`.
+//!
+//! # `dnst` controlling Cascade
+//!
+//! At moments scheduled by `dnst` via "cron" timestamps `KeyManager` checks
```
Philip-NLnetLabs: I think this is wrong or at least not how it should work. At times indicated by the cron_next field, Cascade invokes dnst with the "cron" subcommand. Independent of the cron_next field, each "tick", KeyManager checks if the keyset state file has changed, and if so, invokes the signer to (incrementally) re-sign the zone. The signer checks whether the keyset state file contains any changes that requires signing the zone.

src/units/key_manager.rs
```diff
@@ -1,3 +1,155 @@
...
+//             PublicKey::for_key_id_and_dnssec_algorithm():
+//              client.get_key():
+//                KMIP_Get
+//          if !activate_on_create (i.e. always):
+//            KMIP_Activate
+//   3. Relabel to: <prefix>-<(partial) zone name>-<key tag>-<key type>.
+//        if supports_relabeling:
+//          if !conn.rename_key():
+//            supports_relabeling = false
+//              format_key_label()
```
ximon18: Hmm, the indentation appears to have been messed up here, I'll check it.

src/units/key_manager.rs
ximon18: I made some edits, please re-review. Btw, regarding the re-check after `dnst keyset cron` of the `dnst` managed per-zone state file to see if it has been modified and scheduling of re-signing: is that check redundant given that at the next "tick" the change to the file will be detected and re-signing will then be scheduled anyway?

Philip-NLnetLabs (approved):

src/units/key_manager.rs
Philip-NLnetLabs: Ideally, tick gets replaced by a timer per zone.

ximon18: Like so?

PR #861, opened 2026-07-06
Refactor metrics to be more centralized in metrics.rs

tertsdiepraam:
This is just a small refactor to move where the metrics are defined. I think we want the metrics to be more centrally defined rather than by each component. Metrics are now in the `Center` and there is also a clone of those in each zone, so that operations on a zone are easily accessible.

---


ximon18 (approved):
@tertsdiepraam, this is a work of beauty, I love it. ❤️

PR #866, opened 2026-07-06
Initial version of memory reduction for unsigned data.

Philip-NLnetLabs:
Provide your description here

---


bal-e (commented):
It's hard to review this PR because a lot of the diff is related to the old base / new base conversion, not to the unsigned data work. Is it possible to separate the old base / new base efforts into a distinct PR?

Philip-NLnetLabs:
That would be a ton of work. The main difference is where there was load_unsigned_zone before there is now load_unsigned_diffs. And the new Data type makes sure that new data is stored as diffs.

src/signer/incremental.rs
```diff
@@ -1594,142 +1596,134 @@ impl<'a> IncrementalSigningState<'a> {
...
+                    } else {
+                        let added = HashSet::new();
+                        let mut removed = HashSet::new();
+                        removed.insert(rtype);
+                        self.changes.insert(key.0.clone(), (added, removed));
+                    }
+                }
+                DataChange::Modified { new, .. } => {
+                    let key = (key.0.as_ref(), key.1);
+                    if self.rrsigs.remove(&key).is_some() {
```
bal-e, line 1552: Is there a case where `self.rrsigs.remove(&key)` will return `None`? Can you add a comment in the code to explain why that might happen and why nothing needs to be done in that case? Or, if it is impossible, can you add an `unwrap()` or an `unreachable!()`?

src/signer/incremental.rs
```diff
@@ -2202,19 +2258,425 @@ impl<'a> IncrementalSigningState<'a> {
...
+        let box_owner = record.owner().unsized_copy_into();
+        let key = (box_owner, record.data().rtype());
+        match self.old_data.entry(key) {
+            btree_map::Entry::Vacant(entry) => {
+                entry.insert(vec![record]);
+            }
+            btree_map::Entry::Occupied(mut entry) => {
+                entry.get_mut().push(record);
+            }
+        }
```
bal-e: Minor: I think this can be simplified to `self.old_data.entry(key).or_default().push(record);`.

src/signer/incremental.rs
```diff
@@ -2202,19 +2258,425 @@ impl<'a> IncrementalSigningState<'a> {
...
+                    .expect("position should exist");
+                new.remove(index);
+                let change = if new.is_empty() {
+                    DataChange::Removed { old: old.clone() }
+                } else {
+                    DataChange::Modified {
+                        old: old.clone(),
+                        new,
+                    }
+                };
```
bal-e, line 2278: Are the `.clone()` calls necessary? IIRC Clippy has a lint for this, we should check whether it is enabled.

bal-e (approved):
Looks good! Great job!

src/signer/incremental.rs
Philip-NLnetLabs: Yes: "expected `Vec<&RegularRecord>`, found `&Vec<&RegularRecord>`"

bal-e: Oh, my bad, I thought it was a freshly built `Vec`.

src/signer/incremental.rs
Philip-NLnetLabs: I added the following:
```
                    // Check if RRSIGs exist for the RRset. RRSIGs do not
                    // exist for glue, or in general, all RRsets below a
                    // delegation. Only sign the modified RRset if there are
                    // existing RRSIGs.

```

src/signer/incremental.rs
Philip-NLnetLabs: Done

PR #873, opened 2026-07-07
Introduce `Instances`

bal-e:
This introduces a (relatively) simple `Instances` type, tracked in `ZoneState`, which holds information about the current and upcoming instances of the zone. It is a successor to #617, which itself succeeded #528. I will follow this up with a separate PR which makes existing code read from `Instances`, instead of relying on other state (e.g. `next_min_expiration`).

---


tertsdiepraam (commented):
I'm not sure about this PR. I don't mean that negatively, but there's very little in this PR yet selling me on the idea. We could just accept this and then see where this takes us. Maybe you could explain where these instances will be used?

One concern I have is that this is yet another state machine. Could you explain why we wouldn't for example encode parts of this in the state machine (especially the upcoming instance). Does it need to be globally accessible?

The implementation, as far as I can judge it in this state, is good though. Looks well thought out!

bal-e:
#877 shows the usages of `Instances`. While it would be nice to embed into `ZoneStateMachine`, the same fields would have to show up in many stages (e.g. the upcoming unsigned instance would need to be stored in `LoadedReview`, `Signing`, `SignedReview`, and all the hard-halt states), and I think that would be too much boilerplate.

tertsdiepraam (approved):
We'll give this a shot. I'd love to understand more of the vision but this is probably just one of those things that we just need to take further and see where it goes.

What I'm missing I think is some identifier for an instance to keep track of it through history. I thought that was one of the main advantages, to be able to talk about the lifecycle of a certain instance. But that can be added in another PR.

PR #863, opened 2026-07-06
Use `Scheduler` for re-signing timers

bal-e:
This PR changes the zone signer to use shared scheduler infrastructure (provided by `scheduler.rs`, already used by the zone loader for zone refresh timers) for managing re-signing. It also removes the `resign_busy` field from `Center` in favor of explicit data in `ZoneState`.

Solves #740. Solves #730 (avoids panicking if a zone has no policy).

---

- If you are changing Rust code or integration tests (`Cargo.*`, `crates/`, `etc/`, `integration-tests/`, `src/`):
  - [x] Did you run the integration tests with `act` through the `act-wrapper` (as described in `TESTING.md`)?

PR #877, opened 2026-07-07
Use instances

bal-e:
This PR makes Cascade use `Instances` (as introduced by #873) and replaces a lot of workarounds we had introduced in lieu of it.

TODO: I would also like to replace the current use of `min_expiration` / `next_min_expiration` and store the relevant data in `Instances`, but #863 should be merged first.

---

### DIFF
--- a/integration-tests/tests/review-signed-zone/action.yml
+++ b/integration-tests/tests/review-signed-zone/action.yml
@@ -112,7 +112,9 @@ runs:
       env:
         EXPECTED_SERIAL: ${{ steps.expected-serial.outputs.serial }}
       run: |
-        cascade zone approve --signed example.test ${EXPECTED_SERIAL}
+        # `EXPECTED_SERIAL` was used when the zone was first signed; it has
+        # since been re-signed, so the serial number has increased.
+        cascade zone approve --signed example.test $((EXPECTED_SERIAL + 1))
         
     - name: Querying the zone at the publication server should succeed
       run: |
--- a/src/signer/zone.rs
+++ b/src/signer/zone.rs
@@ -191,23 +191,15 @@ impl SignerZoneHandle<'_> {
         // needs re-signing (i.e. that the signing keys used for building that
         // instance are different from the latest ones).
         //
-        // TODO: Stop relying on `{next_,}min_expiration` to tell us about the
-        // published and upcoming instances of the zone.
-        //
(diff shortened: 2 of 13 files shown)


integration-tests/tests/review-signed-zone/action.yml
```diff
@@ -112,7 +112,7 @@ runs:
       env:
         EXPECTED_SERIAL: ${{ steps.expected-serial.outputs.serial }}
       run: |
-        cascade zone approve --signed example.test ${EXPECTED_SERIAL}
+        cascade zone approve --signed example.test $((EXPECTED_SERIAL + 1))
```
Philip-NLnetLabs, line 117: Why is this? Is EXPECTED_SERIAL wrong?

src/signer/zone.rs
```diff
@@ -191,23 +191,15 @@ impl SignerZoneHandle<'_> {
...
-        // an upcoming signed instance does, we should try re-signing it too.
-        // But before re-signing is actually initiated, we will verify that
-        // the upcoming instance was accepted and published (and ignore the
-        // re-signing request otherwise).
-        if self
-            .state
-            .min_expiration
-            .or(self.state.next_min_expiration)
-            .is_none()
+        // an upcoming signed instance does (or if the zone is being restored
```
Philip-NLnetLabs, line 196: "an upcoming signed instance does"

I guess something is missing here.

Philip-NLnetLabs (approved):
Looks good to me. Two minor comments. Hard to be sure that the logic is correct but it looks fine.

integration-tests/tests/review-signed-zone/action.yml
bal-e: Yes, the zone gets re-signed and the signer makes sure it does not use the same serial twice.

src/signer/zone.rs
bal-e: Right, I'll fix the comment.

bal-e: Hm, seems the comment is fine, it's just hard to tell because of the removals GitHub shows above it.

integration-tests/tests/review-signed-zone/action.yml
bal-e: Added a comment.

ximon18: Why did it work before then?

PR #882, opened 2026-07-08
Report signing step to zone status

tertsdiepraam:
Does a few things:

- Report signing steps to the zone status.
- This required adding a `step` to the `SigningStatus`, which kind of overlaps with the `current_action` but I left that in, because that still shows more granular information that might be helpful for debugging or something.
- There was some info in the status that was shown as "not yet known" even though it was, because the HTTP server only included the signing report on signing review, which is now fixed.
- We now also trigger `SigningStatus::InProgress` in incremental signing, which turned out to not happen yet.
- We now show the signed serial in the zone status because we now take it from the signing status and not from the `signed_review_soa` on the storage because the signer knows the serial earlier.
- We now also show the signing strategy (full or incremental).
- Finally, some unused metrics were removed. They were also not created for incremental signing, so this is not a very big deal.

```
zone:   se
policy: default
source: /home/terts/data/zones/se.zone.stripped.txt

review
  loaded: off
  signed: off

last published
  <no versions published yet>

status: signing
  ✔ load (serial: 1776081352)
  ○ review loaded zone (disabled)
  ● sign (serial: 2026071300)
  |   loaded serial: 1776081352
  |   signed serial: 2026071300
  |   strategy: sign the entire zone
  |   step 1/5: collecting records
  |   start time: 2026-07-13T14:18:15 (9s ago)
  |
  ○ review signed zone (disabled)
  ○ publish
```

---

### DIFF
--- a/src/signer/keys.rs
+++ b/src/signer/keys.rs
@@ -66,8 +66,6 @@ impl ZoneSigningKeys {
         keyset_state: &KeySetState,
         status: &RwLock<SigningStatusPerZone>,
     ) -> Result<Self, Box<LoadError>> {
-        status.write().unwrap().current_action = "Loading signing keys".to_string();
-
         let mut list = Vec::new();
 
         for (pub_key_name, key_info) in keyset_state.keyset.keys() {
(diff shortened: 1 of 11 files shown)


ximon18:
I haven't read the changes yet but I like the CLI output updates shown in the PR description!

src/signer/keys.rs
```diff
@@ -282,9 +274,6 @@ impl KeyPair {
         let kmip_conn_pool = match kmip_servers.entry(priv_key_url.server_id().to_string()) {
             std::collections::hash_map::Entry::Occupied(e) => e.into_mut(),
             std::collections::hash_map::Entry::Vacant(e) => {
-                status.write().unwrap().current_action =
-                    format!("Connecting to KMIP server '{}'", priv_key_url.server_id());
```
bal-e, line 286: @ximon18 had asked to retain these in case the KMIP connection takes very long, or hangs outright. It would help users debug the root cause.

bal-e (changes requested):
Looks great overall!

Issue #871, opened 2026-07-07
Panic when removing a zone during restore


Philip-NLnetLabs:
When removing a zone while Cascade is restoring the zone, cascaded panics:
```
2026-07-07T09:23:02.522845Z ERROR cascaded::daemon: thread 'cascade-worker' (ProcessId(923524), ThreadId(30)) panicked at src/server/service.rs:850:53: the zone exists
```

Code:
```
    846     pub fn remove_zone(&self, zone: &Arc<Zone>) {
    847         let mut state = self.state.write().unwrap();
    848         let name = RevNameBuf::parse_bytes(zone.name.as_slice()).unwrap();      
    849         let existed = state.zones.remove(&*name);
    850         let ServedZone { handle, viewer } = existed.expect("the zone exists");
    851         assert!(
    852             Arc::ptr_eq(&handle, zone),
    853             "distinct 'Arc<Zone>'s had the same name"
    854         );
    855         let _ = viewer;
    856     }
    857 }
```

bal-e:
I looked into this quickly; a simple fix would be to check whether the zone is mid-restoration, then to avoid calling `*Server::remove_zone()`. But this doesn't solve everything; if the zone is mid-restoration, it will call `*Server::add_zone()` when it finishes, and this will never be removed (a waste of memory + possibly panicking if a zone of the same name is added later). I think what we need is some kind of `mid_removal` flag on `ZoneState` (or maybe a `ZoneStateMachine::Removing` state?). Then the restoration code can check this and avoid calling `*Server::add_zone()` too.

Philip-NLnetLabs:
Maybe quick fix is to return some kind of error when trying to remove zone while cascade is restoring it. Otherwise we could get inconsistent answers from other CLI commands.

bal-e:
Fixed by #886 ... I'm going to make a new issue to consider whether we should allow removing a zone during restoration at all.

PR #886, opened 2026-07-10
Prohibit removing a zone while it is being restored from disk

bal-e:
A quick fix for #871, as suggested by @Philip-NLnetLabs.

---


bal-e:
@Philip-NLnetLabs can you check whether you still hit the crash?

ximon18 (approved):
LGTM. Tested manually and zone was able to be removed while restoring and `zone status` worked fine afterward and the cascade log file looked fine. Re-adding the zone also worked fine.

However, the persisted data files remained behind when the zone was removed. I assume this is actually a separate existing bug. If one re-adds the zone it also still has the old persisted data files.  I will raise this as a separate bug.

PR #899, opened 2026-07-13
Trim the dependency tree

bal-e:
A clean `cargo check` on `main` pulls in 279 crates, fills 660MiB in `target`, and takes ~25 seconds.

A clean `cargo check` with this PR pulls in 251 crates (-28), fills 597MiB in `target` (-63MiB), and takes ~23 seconds.

- Disabled IDN support for URLs, saving a good number of dependencies.
- `futures` is an umbrella project pulling in many crates, I limited it to `futures-util` which provides what we need (and is already being pulled in by `domain`).
- `cascade-cfg` unnecessarily required `clap` features, I disabled all of them.
- Removed the `crossbeam-utils` dependency, which was unused.

Further savings can be obtained removing `syslog` and `time` from `daemonbase` (we already do our own logging, and we use `jiff` for the rest of Cascade). Left for future work.

---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -170,6 +170,13 @@ hostname = "0.4.1"
 # organization.
 prometheus-client = "0.24.0"
 
+# `idna`, which provides internationalized domain names, is pulled in by `url`.
+# It adds a significant number of dependencies. But Cascade only uses pure-ASCII
+# URLs, so IDN support is unnecessary. `url` does not offer a way to disable
+# `idna` directly; instead, the documented (albeit hacky) method to disable IDN
+# support is to force the version of the `idna_adapter` dependency.
+idna_adapter = { version = "=1.0.0" }
+
 cascade-api = { path = "crates/api" }
 cascade-cfg = { path = "crates/cfg" }
 cascade-zonedata = { path = "crates/zonedata" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -158,10 +158,11 @@ tracing-subscriber.workspace = true
 lazy_static = { version = "1.5" }
 
 # support-color checks various environment variables and terminal settings for
-# whether ANSI escape codes should be emitted. It is also used by `anstream`
-# (and therefore `clap`).
+# whether ANSI escape codes should be emitted.
 #
-# TODO: `anstream` now uses `anstyle-query` instead of `supports-color`.
+# TODO: This dependency was chosen because `anstream` (which we depend on
+# elsewhere) used it, but `anstream` now uses `anstyle-query`. We now have two
+# crates in our dependency graph that have the same functionality.
 supports-color = "3.0.2"
 
 # `hostname` is a very popular crate to get the hostname of the current process.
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -116,14 +116,15 @@ axum               = { version = "0.8.4", default-features = false, features = [
...
 daemonbase         = { version = "0.1.5", features = ["tokio"] }
-futures            = "0.3.17"
-futures-util       = "0.3"
 serde              = { workspace = true, features = ["derive", "rc"] }
 rayon = "1.10.0"
 serde_json         = "1.0"
 serde_with         = "3"
 url                = { version = "2.4", features = ["serde"] }
 
+# `futures` provides some async functionality not available in `std` yet.
```
ximon18: I'm confused by this comment. You removed `futures` then refer to it? If it provides some async stuff but we don't use it, why mention it?

bal-e: `futures-util` is a part of the `futures` umbrella of crates, so the reference made sense to me (since below this comment is a dependency on `futures-util`). I can clarify it.

bal-e:
Note: updated to trim some more deps.

Cargo.toml
bal-e: I've added more detail to the comment.

Cargo.toml
```diff
@@ -157,6 +160,8 @@ lazy_static = { version = "1.5" }
 # support-color checks various environment variables and terminal settings for
 # whether ANSI escape codes should be emitted. It is also used by `anstream`
 # (and therefore `clap`).
+#
+# TODO: `anstream` now uses `anstyle-query` instead of `supports-color`.
```
ximon18: It's unclear to me from this comment what action is being proposed here. Currently for me it adds no value, only confusion.

ximon18 (approved):
LGTM although I haven't tested the changes.

Cargo.toml
bal-e: The "it is also used by `anstream`" note just above it is incorrect, the TODO is here to mark that we should re-evaluate the dependency.

ximon18: Ok. That isn't clear to me from the text.

Cargo.toml
```diff
@@ -167,6 +172,13 @@ hostname = "0.4.1"
 # organization.
 prometheus-client = "0.24.0"
 
+# `idna`, which provides internationalized domain names, is pulled in by `url`.
+# It adds a significant number of dependencies. But Cascade only uses pure-ASCII
+# URLs, so IDN support is unnecessary. `url` does not offer a way to disable
+# `idna` directly; instead, the documented (albeit hacky) method to disable IDN
+# support is to force the version of the `idna_adapter` dependency.
+idna_adapter = { version = "=1.0.0" }
```
tertsdiepraam, line 182: And we're sure this won't break later? Where is this documented?

bal-e: https://crates.io/crates/idna_adapter

It's certainly very weird, but this appears to be a stable solution.

Cargo.toml
bal-e: Made the comment more explicit.

Cargo.toml
bal-e: I've added the URL to the comment.

PR #900, opened 2026-07-14
Persist maintenance mode and fix an issue with it

### LINKED ISSUE #816: `cascade zone maintenance enable` is forgotten on process restart.
After restart `cascade zone status` lists the zone as idle again rather than in maintenance mode.

This, to my mind, seems to go against the intent of the operator.

tertsdiepraam:
Closes #816


ximon18:
Was the bug reproducible? I'm wondering if we can extend the system tests to ensure we don't reintroduce it later.

tertsdiepraam:
Yes, it's just
1. Add a zone, starting a new load
2. Put it into maintenance mode before it's done.

However, I can't run the tests, so I can't really implement it.

bal-e:
Integration tests pass.

bal-e (approved):

PR #901, opened 2026-07-14
Add persisting states to the state machine and zone status

### LINKED ISSUE #867: `zone status` doesn't really work for large zones / insufficient indication of what is happening.
Often Cascade is doing something that is not reflected in zone status.

Reported by @Philip-NLnetLabs.

tertsdiepraam:
Looks like this in zone status:
```
status: persisting loaded zone
  ✔ load (serial: 1776081352)
  ○ review loaded zone (disabled)
  ● persist loaded zone
  |
  ○ sign
  ○ review signed zone (disabled)
  ○ persist signed zone
  ○ publish
```

I think this closes #867, at least as far as I can tell from the description.

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -496,10 +496,12 @@ pub enum Progress {
     Loading,
     LoadedReview,
     HaltLoaded,
+    PersistingLoaded,
     Signing,
     SigningFailed,
     SignedReview,
     HaltSigned,
+    PersistingSigned,
 }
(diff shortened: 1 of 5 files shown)


bal-e (approved):
Wow, that looked surprisingly straightforward!

bal-e:
The `upstream-tsig` integration test is failing. Investigating.

bal-e:
Apparently flaky.

PR #906, opened 2026-07-15
`zone edit` command to change zone source and policy

### LINKED ISSUE #846: Add support for changing the source of a zone.
Currently, the source of a zone is fixed when the zone is added. Add support in Cascade and Cascaded to change the source of a zone.

### LINKED ISSUE #868: A zone cannot switch / change to a different profile / policy
We should support changing the policy that a zone is currently using.

Reported by @Philip-NLnetLabs.

### LINKED ISSUE #903: How to change the policy of a zone?
I want to migrate a zone to another policy but I find no way to apply a new policy. The obvious one fails;

```
 % cascade  zone add --source /etc/cascade/zones/cours-dns.fr --policy ovh-no-ed25519 cours-dns.fr 
2026-07-15T07:08:00.002914Z ERROR cascade: Failed to add zone: a zone of this name already exists
```

Is the only way to remove the zone and re-add it?

tertsdiepraam:
Allows for the following controls:
```
cascade zone edit se --source ~/data/zones/se.zone.txt
cascade zone edit se --policy my-policy
```
It's also possible to pass both.

The output could be improved if for example the new policy is the same policy that is already in use.

Closes #903
Closes #868
Closes #846


src/units/http_server.rs
```diff
@@ -300,6 +302,134 @@ impl HttpServer {
...
+                .transpose()?;
+
+            // Now we start making actual changes.
+            // Anything below this point should be infallible.
+
+            let mut changes = Vec::new();
+
+            if let Some(policy) = policy {
+                let mut handle = zone.write_handle(center);
+                let old = handle.state.policy.replace(policy.latest.clone());
```
bal-e: I think you also need to un-link the zone from the policy here...

tertsdiepraam: Sure, how?

tertsdiepraam: I think I got it now, please take a look.

bal-e:
I'll continue this work.

bal-e:
Note: the CLI command should wait until all policy changes (including keyset stuff; note there are some `tokio::spawn()`s to watch out for) to finish before it returns.

bal-e:
Integration tests pass.

PR #907, opened 2026-07-15
Fix: use id's to sort persisted diffs instead of file paths

### LINKED ISSUE #908: Incorrect PersistedDiffFileInfo sort order
The order of `PersistedDiffFileInfo ` items in the `diff_infos` connection of `PersistedDiffManager` is incorrect.

It should order items in insertion order. However it orders by path and not according to the order the parts were generated using the unique integer identifier as the sort is alphabetical.

Previously I resisted using a `Vec`-like type as earlier diff purging implementations lacked serial number relation metadata and used unclear and error prone `Vec` indexing, and I wanted an ordered "set" type whereas `Vec` is neither ordered nor a set.

However, the `BTreeMap` ordering is not insertion order, and the current version of the code indexes into it only by path, lookup by serial number requires a search as using a `Vec` would require, while a `Vec` would be order by insertion.

So perhaps it is best to switch to a `Vec`-like type.

tertsdiepraam:
This was something I missed during the purging review. I think this is the simplest fix for it. I'm intentionally keeping this very small and self-contained to get the fix in as quickly as possible.

The issue is that the paths get sorted lexicographically so the path `se.loaded.10` is sorted before `se.loaded.2`. To fix this, we store the id as a `usize` separately and implement `Ord` by comparing just that field.

Closes #908

### DIFF
--- a/src/persistence/zone.rs
+++ b/src/persistence/zone.rs
@@ -656,7 +656,12 @@ impl PersistedDiffManager {
                 self.next_uniqifier
             ))
             .into_std_path_buf();
-        let file_info = PersistedDiffFileInfo::new(path.clone(), loaded_serial, signed_serial);
+        let file_info = PersistedDiffFileInfo::new(
+            self.next_uniqifier,
+            path.clone(),
+            loaded_serial,
+            signed_serial,
+        );
 
         trace!(
             "Pushing diff with loaded serial {loaded_serial:?} and signed serial {signed_serial:?} for zone '{}' with path '{}'",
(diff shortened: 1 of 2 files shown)


src/persistence/zone.rs
```diff
@@ -741,6 +746,8 @@ impl PersistedDiffManager {
 /// Information about a single persisted zone data file.
 #[derive(Clone, Debug)]
 pub struct PersistedDiffFileInfo {
+    id: usize,
```
ximon18, line 749:
```suggestion
    /// Field to order by so that entries in the tree can be iterated over in insertion order.
    id: usize,
```

ximon18 (approved):
During review of PR #657 I was actually against adding an `id` field to this type, ironically, as it added no value, but now it is being added as a work around for a huge oversight on my part... very appreciated @tertsdiepraam.

bal-e:
The `persist-then-reject` integration test seems to fail. Investigating.

bal-e:
Never mind, it's flaky (time-sensitive).

PR #912, opened 2026-07-16
Hint at the existence of --detailed

maertsen:
I rediscovered the existence of cascade zone status --detailed, so I figured it could figure a bit more prominently in the quick start.


bal-e (approved):
Thanks!

PR #913, opened 2026-07-16
Rename cascaded.service to cascade.service

### LINKED ISSUE #895: Consider renaming cascade**d**.service to cascade.service?
This is feedback from dogfooding Cascade internally, going only by public documentation.

The thing that tripped me up twice, last week (when `systemctl enable cascade.service`), and just now once again (`journalctl -u cascade.service`), is that the systemd service is called cascade**d**.service (matching the daemon process), and not cascade.service (like the name of the project). I understand why it’s named that way, but somehow it goes against what I’m used to. 

I think that's primarily to do with how common systemd services I use follow the <projectname>.service scheme. One example is ssh.service starting ssh**d**. I'm sure there are counterexamples, just reporting my experience.

I am curious about the preferences of others, but I would be in favor of renaming the .service file to cascade.service.

maertsen:
In response to the feedback collected in #895, rename the systemd service files, and their mentions in the docs.

---

I did not yet, let's see if I can do so.

Would you like me to add a commit for a Changelog entry, or is that too likely to cause merge conflicts?


bal-e (approved):
Nice!

I don't think the integration tests rely on the systemd service, so you don't have to run them. I'll add a Changelog entry after the PR is merged :)

Issue #436, opened 2026-01-27
Event Tracing and Logging


bal-e:
One of Cascade's key principles is _transparency_ -- giving the user precise information about what happened, when it happened, and why it happened. A feature that has been considered since its inception is _event tracing_ -- tracking the causality of high-level events (e.g. "received a `zone refresh` command from the API" -> "the zone is being refreshed" -> "a new instance of the zone was loaded" -> "the zone is being signed"). We want to collect this information within Cascade and offer human-facing views of it (e.g. answering queries like "why did event X happen"). We also want these high-level events to be associated with low-level logs, e.g. debug logging from a zone refresh operation.

At the moment, some high-level events are collected by "zone history". They do not contain causality information. Low-level logs are collected by `tracing` (which also collects from `log`). We would like to integrate these two systems together and collect causality information (at least for high-level logs).

@tertsdiepraam and I discussed what such a system should look like. We realized that all our high-level events are associated with state transitions of high-level components (i.e. the zone loader, signer, server, and key manager), and each of these components will eventually have explicit state machines in `ZoneState`. We plan to store some kind of "high-level logger" within / next to these state machines. Many states (e.g. "refreshing a zone") are directly associated with particular `async fn`s / Tokio tasks -- we can wrap these in `tracing::instrument`, which will collect low-level logs for us and associate them with the right tasks. We can implement a custom `tracing` subscriber that collects those low-level logs, figures out which high-level events they are a part of, and stores them in some kind of internal database (while also logging them as usual).

An aside: I am of the personal opinion that log levels should be hierarchical (i.e. if function `foo` is outputting "debug" logs, and it encounters an error, it should output a "debug error" log message; and this should extend structurally to the subroutines it calls), strongly typed (i.e. expressed as an `enum`, similar to the conventions for errors in Rust today) and explicitly part of library APIs (including SemVer). Some of these ideas may feature in the implementation, at least in the high-level event system.

PR #914, opened 2026-07-16
Propagate policy reload warnings to the CLI

### LINKED ISSUE #462: When `policy reload` ignores a file due to it being misnamed, the warning is not visible to the CLI user.
```
$ target/release/cascade policy reload
Policies reloaded:

```

But in the logs I see this:

```
2026-02-17T12:07:07.489877Z  WARN cascade::policy: Ignoring potential policy '/tmp/cascade/policies/default'; policies must end in '.toml'
```

tertsdiepraam:
Closes #462.

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -844,6 +844,7 @@ impl Display for PolicyReloadError {
 #[derive(Deserialize, Serialize, Debug, Clone)]
 pub struct PolicyChanges {
     pub changes: Vec<(String, PolicyChange)>,
+    pub warnings: Vec<String>,
 }
 
 #[derive(Deserialize, Serialize, Debug, Clone)]
--- a/crates/cli/src/commands/policy.rs
+++ b/crates/cli/src/commands/policy.rs
@@ -109,7 +109,7 @@ impl Policy {
                 }
 
                 for w in res.warnings {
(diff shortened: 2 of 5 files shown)


bal-e (approved):
Nice! A few thoughts:

1. If a policy reload is initiated from the CLI and there are warnings, users might expect to be able to find the same warnings in the cascade daemon logs. Would it make sense to log to the `warnings` vec as well as to `tracing`?

2. In the long term, I'd like to leverage `tracing` to extract this information for us (#436). The same mechanism could be used to build the zone history implicitly from logs, instead of manual `record_zone_event()` calls.

These are not blockers for merging.

tertsdiepraam:
> If a policy reload is initiated from the CLI and there are warnings, users might expect to be able to find the same warnings in the cascade daemon logs. Would it make sense to log to the warnings vec as well as to tracing?

That was the first version I did. It felt unnecessary. I think these are first and foremost immediate errors for the user to see when reloading. But maybe if this becomes the convention then we could do that.

> In the long term, I'd like to leverage tracing to extract this information for us (https://github.com/NLnetLabs/cascade/issues/436). The same mechanism could be used to build the zone history implicitly from logs, instead of manual record_zone_event() calls.

Agreed, would be cool.

PR #920, opened 2026-07-27
Make signing functions unit-testable

bal-e:
This should make it easier for us to introduce more complex or thorough tests for signing. We can copy the existing `incremental-signing` integration tests into unit tests, and possibly remove some (not all) of the tested cases from the integration tests.

---


Philip-NLnetLabs (approved):
Looks good to me.

PR #932, opened 2026-07-30
Fix display of Ed25519 when writing policy.

### LINKED ISSUE #930: Inconsistent algorithm display name.
From: https://github.com/NLnetLabs/cascade/blob/d72715da6e8c606fecf0dfe9c8e839597bfded1a/src/policy/file/v1.rs#L588

```rust
impl Display for KeyGenerationParametersSpec {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        f.write_str(match self {
            Self::RsaSha256(2048) => "RSASHA256",
            Self::RsaSha512(2048) => "RSASHA512",
            Self::EcdsaP256Sha256 => "ECDSAP256SHA256",
            Self::EcdsaP384Sha384 => "ECDSAP384SHA384",
            Self::Ed25519 => "Ed25519",
            Self::Ed448 => "ED448",

            Self::RsaSha256(bits) => return write!(f, "RSASHA256:{bits}"),
            Self::RsaSha512(bits) => return write!(f, "RSASHA512:{bits}"),
        })
    }
}
```

Note the difference between ED25519 and ED448.

I tripped up on this while fixing https://github.com/NLnetLabs/ods2cascade/issues/62 as I decided to test every algorithm number in input sample policy files for `ods2cascade` and expected the wrong algorithm name string in the generated Cascade policy file.

I assume this should be all uppercase as these are algorithm mnemonics that are being output.

Philip-NLnetLabs:
Fixes #930

### DIFF
--- a/src/policy/file/v1.rs
+++ b/src/policy/file/v1.rs
@@ -585,7 +585,7 @@ impl Display for KeyGenerationParametersSpec {
             Self::RsaSha512(2048) => "RSASHA512",
             Self::EcdsaP256Sha256 => "ECDSAP256SHA256",
             Self::EcdsaP384Sha384 => "ECDSAP384SHA384",
-            Self::Ed25519 => "Ed25519",
+            Self::Ed25519 => "ED25519",
             Self::Ed448 => "ED448",
 
             Self::RsaSha256(bits) => return write!(f, "RSASHA256:{bits}"),


ximon18 (approved):
I applied this patch to a local copy of cascade v0.1.0-beta4 and used it to build ods2cascade. This fixed the test that currently fails due to this issue. This PR won't fix the actual ods2cascade failing test as ods2cascade cannot use v0.1.0-beta5 which this PR builds upon, due to the lack of lib support in v0.1.0-beta5, but this does show that this PR has the desired effect.

Issue #951, opened 2026-08-10
Error message 'a patchset could not be applied' doesn't give enough information.


ximon18:
This error message is caused by one of many different error returns from `crates/zonedata/src/writer.rs` that return `PatchError::Inconsistency`, but what kind of inconsistency?

In particular we have seen it from `next_patchset()` which has multiple such return paths, but from the error alone it is not possible to know which was triggered. Was it an inconsistent attempt to remove a SOA which does not exist? Was it an attempt to add the same record twice, or an attempt to add the same record twice? And which records were the cause of the problem?

One cannot simply add trace level logging as the zonedata crate has no support for logging or tracing, though `cargo add tracing` is one way to make that possible. An alternate approach would be to add a reason string to the `Inconsistency` enum variant and let the caller do the logging.

PR #952, opened 2026-08-10
Include the correct loaded SOA in the compacted zone file. (fixes #948)

### LINKED ISSUE #948: Wrong serial requested from upstream via IXFR
I'm seeing this in a test setup after running for a while then restarting Cascade.

In this particular case the Cascade instance is not able to receive NOTIFY messages from the upstream so refreshes according to the SOA REFRESH period.

Both upstream and Cascade are using YYYYMMDDNN serial policy, but in the time between refreshes the upstream advances N more than once so the Cascade serial increases less often than the upstream serial.

I suspect this doesn’t show up in our tests because they receive every upstream update and keep in sync with the upstream serial number thereby masking a problem that occurs when they are not in sync.

Perhaps also important to note is that pruning of diffs had occurred as there were more diffs at some point than the max allowed (5).

ximon18:
When compacting the zone new loaded and signed snapshots of the zone are written to disk. The code was wrongly including the signed SOA in the loaded snapshot. Once restored the next IXFR request to the upstream correctly used the SOA from the loaded zone but that SOA was actually from the signed copy of the zone.

Normally this would cause an AXFR and everything would continue working.

However, if the upstream and Cascade were both using `data-counter` serial policy, AND the upstream were queried less frequently for updates than new updates were available (e.g. because the upstream was unable to send NOTIFY messages to Cascade so it only fetched when the SOA REFRESH timer expired) then this could cause the requested IXFR serial to refer to a valid diff available at the upstream but for an older serial than the most recent one that Cascade had actually received. This would result in error message 'a patchset could not be applied' which is not very helpful (#951 tracks this issue).

E.g. if the upstream created zone versions with serial numbers 2026081000, 2026081001, 2026081002, 202608101003 then Cascade may have fetched 2026081000 and 2026081002, and produced signed zone serials 2026081000 and 2026081001. If compaction occurred the 2026081001 SOA would wrongly be written to the compacted loaded zone snapshot. On restart of Cascade that would be restored then an IXFR to the upstream with serial 2026081001 would be sent instead of the correct value 2026081002. The resulting received diff for 2026081001->2026081002 would not cleanly apply to the current loaded version of the zone in memory in Cascade.

---

  - [x ] Did you run the integration tests with `act` through the `act-wrapper` (as described in `TESTING.md`)?

### DIFF
--- a/src/persistence/zone.rs
+++ b/src/persistence/zone.rs
@@ -387,7 +387,7 @@ impl PersistenceState {
             persist_to_file_from_parts(
                 loaded_snapshot_path,
                 None,
-                reader.soa().clone(),
+                reader.loaded().soa().clone(),
                 [].iter(),
                 reader.loaded_records(),
             );


bal-e (approved):

ximon18:
I'll merge this as it has been approved, but  I think I should also extend the tests to make sure this kind of error doesn't go undetected. I'll add an issue as a reminder to do that.

ximon18:
> I'll merge this as it has been approved, but I think I should also extend the tests to make sure this kind of error doesn't go undetected. I'll add an issue as a reminder to do that.

See https://github.com/NLnetLabs/cascade/issues/954.

PR #963, opened 2026-08-12
Debug command to expose the last and next moments that XFR will be attempted.

ximon18:
This is something that keeps bugging me: I can't tell from Cascade when it is going to try an upstream XFR next, and whether that will be a normal attempt or a retry.

I don't think this is something a user should ever need to normally care about, so I wondered about exposing it via a debug command, hence this PR.

I have not put any effort into styling the output at this point as I want to know first if anyone else sees value in this.

I would also like to extend this to show the set of available IXFR diffs in memory that can be served to downstream clients.

Example output:

```
$ ./cascade debug xfr-status example.com
zone: example.com
refresh status: refreshing: last=2026-08-12T10:22:58Z, next=2026-08-12T11:22:58Z
```

Another way to approach this could be to expose this via metrics, e.g. last refresh time and whether or not the last refresh was successful or not.

It could also just be shown as part of `zone status` but we're trying to keep that fairly minimal. It could also be added to `zone status --detailed` but I think we should keep that to things operators really need to know, and this feels more like diagnostics in case of an unexpected problem or an aid when debugging/testing.

Yet another idea: perhaps this should be considered part of the missing information we want to show about "what is going to happen next" for a zone?

---

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -481,6 +481,27 @@ pub struct ZoneStatus {
 }
 
 #[derive(Deserialize, Serialize, Debug, Clone)]
+pub struct ZoneXfrStatus {
+    pub name: ZoneName,
+    pub refresh_timer_state: RefreshTimerState,
+}
+
+#[derive(Deserialize, Serialize, Debug, Clone)]
+pub enum RefreshTimerState {
+    Disabled,
+
+    Refresh {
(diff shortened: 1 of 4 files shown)


Philip-NLnetLabs:
Nice. I think there should be an entry in the documentation. We experiment with these features and later see if it should move to zone status or not.

Issue #316, opened 2025-11-03
Bug: policy can be loaded that refers to a non-existent HSM


bal-e:
At the moment, the policy loader doesn't look up specified HSMs, and so doesn't notice if they don't exist.  This requires a fairly simple change so that `PolicyVersion` stores a direct link to an HSM -- but this also relates to #303.

PR #964, opened 2026-08-12
Policy hsm improvements

### LINKED ISSUE #685: `policy reload` should fail if the referenced HSM does not exist.
If `hsm-server-id` in a policy file refers to a HSM which has NOT been added via `cascade hsm add` then `policy reload` should fail. Currently it does not.

### LINKED ISSUE #727: Changing policy field `hsm-server-id` has no effect on existing zones using the policy.
When the `cascade zone add --policy P` command is executed and policy P has `hsm-server-id` set to a HSM that was previously registered via command `cascade hsm add`, then Cascade executes a number of `dnst keyset` commands including `dnst keyset kmip add-server` like so:

```
2026-06-09T08:27:36.287497Z DEBUG cascaded::units::key_manager: Executing keyset command /tmp/cascade-dnst keyset -c /tmp/cascade/keys/arya.com.cfg kmip add-server cascade-hsm-bridge 127.0.0.1 --port 5696 --connect-timeout 3s --read-timeout 30s --write-timeout 3s --max-response-bytes 8192 --key-label-max-bytes 32 --insecure --credential-store /tmp/cascade/kmip/credentials.db
```

However, if the policy had an empty or missing `hsm-server-id` setting but later that `hsm-server-id` is set and the `cascade policy reload` is command is executed, while various `dnst keyset` commands are again executed, the `kmip add-server` command is not one of them.

Philip-NLnetLabs:
Fixes #727 
Fixes #685 
---

### DIFF
--- a/src/hsm/mod.rs
+++ b/src/hsm/mod.rs
@@ -25,6 +25,10 @@ impl HsmStore {
     }
 
     // TODO: Methods to get, modify, add, remove HSMs?
+
+    pub fn server_exists(&self, hsm_server_id: &str) -> bool {
+        self.map.contains_key(hsm_server_id)
+    }
 }
 
 //----------- Hsm --------------------------------------------------------------
--- a/src/main.rs
+++ b/src/main.rs
@@ -242,6 +242,7 @@ fn main() -> ExitCode {
                 &mut state.policies,
                 &config,
                 &state.tsig_store,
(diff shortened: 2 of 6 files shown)


bal-e:
Fixes #316.

bal-e (approved):
Looks good!

PR #971, opened 2026-08-24
Introduce an access control service layer.

### LINKED ISSUE #763: Manually handle TSIG within the zone server
Rather than relying on `domain`'s TSIG middleware layer, manually handle all TSIG concerns. This would allow us to tackle #607 and #608. We need to make sure the functionality used by the middle layer is public in `domain` so we can use it. This would add a significant amount of code, but it would mostly be copied from `domain`. And this would let us gradually convert it to new-base.

Philip-NLnetLabs:
Fixes #763 
---

### DIFF
--- a/src/units/zone_server.rs
+++ b/src/units/zone_server.rs
@@ -653,7 +653,8 @@ where
 
         // Find the right zone.
         #[allow(clippy::mutable_key_type)]
-        let zones = &self.center.state.lock().unwrap().zones;
+        let zones = { &self.center.state.lock().unwrap().zones };
+
         let Some(zone) = zones.get(&qname).map(|z| &z.0) else {
             // No such zone could be found.
             let rcode = if opcode == Opcode::QUERY && (qtype == Rtype::AXFR || qtype == Rtype::IXFR)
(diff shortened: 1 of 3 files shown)


src/units/zone_server.rs
```diff
@@ -535,3 +549,361 @@ impl Notifiable for LoaderNotifier {
...
+            }
+        };
+        let qname = question.qname();
+        let qname: Name<Bytes> = qname.to_name();
+        let qtype = question.qtype();
+
+        // Find the right zone.
+        #[allow(clippy::mutable_key_type)]
+        let zones = &self.center.state.lock().unwrap().zones;
+        let Some(zone) = zones.get(&qname).map(|z| &z.0) else {
```
bal-e: This will hold on to a lock of the global state for the duration of this function (but not the full duration of the XFR). To make sure this doesn't silently introduce hangs in the future, could you put the code locking `center.state`, looking up the zone, and copying out `Arc<Zone>` in a `{}` block?

bal-e (approved):
It's a lot of code, but it looks good to me :)

PR #980, opened 2026-08-27
docs: Create test zone in quick start

thgoebel:
Instruct users how to create test zone in the quick start. This makes it easier to follow the quick start if you don't have a zone file lying around.

The HSM guides (SoftHSM, SmartCard-HSM, NetHSM) all have a "Sign a Test Zone" section. This shows that there is a user need for this piece of information. We might as well move this up to the generic quick start guide.

---

_None of the below apply_

### DIFF
--- a/doc/manual/source/quick-start.rst
+++ b/doc/manual/source/quick-start.rst
@@ -192,6 +192,21 @@ default policy directory is not writable by the current user.
    cascade template policy | sudo tee /etc/cascade/policies/default.toml
    cascade policy reload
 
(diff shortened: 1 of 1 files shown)


bal-e (approved):
Thanks, this is a great idea!

P.S. I now see that different parts of the documentation use different example domains, e.g. the NetHSM manual uses `example.net`. We could standardize all of these to use `example.org` since IIRC that is "officially" a test domain. But that's not important for this PR.

PR #979, opened 2026-08-27
docs: Improve dev experience

thgoebel:
Improve dev experience:

- Hot reloading with sphinx-autobuild
- More make targets for common actions
- Document main make targets in README for new contributors

If this change is acceptable, I can make a similar PR for the docs of cascade-hsm-bridge.

Also, consider updating/regenerating the requirements.txt and uv.lock. I deliberately omitted this from the PR, so that a maintainer can do it (as a more trusted source of the lockfile update).

---

_None of the below apply_

- If you are changing Rust code or integration tests (`Cargo.*`, `crates/`, `etc/`, `integration-tests/`, `src/`):
  - [ ] Did you run the integration tests with `act` through the `act-wrapper` (as described in `TESTING.md`)?

- If you are adding/deleting man pages:
  - [ ] Did you update the `man_pages` config in `doc/manual/source/conf.py`?
  - [ ] Did you update the packaged man pages in the `Cargo.toml`?
  - [ ] Did you commit the freshly built man pages?

- If you are modifying man pages:
  - [ ] Did you commit the updated built man pages?

PR #984, opened 2026-08-27
docs: Add CI to check that man pages are committed

thgoebel:
This PR simplifies the pull request template by automating the "man page modified" check. (The "man page added/removed" case I didn't cover, because I don't want to mess with shell magic right now.)

This PR depends on #979 because it uses "make man", and thus relies on "uv run" being in the Makefile. Hence the Makefile is also in the list of changed paths, to catch Makefile edits that might break this Action.

---

### DIFF
--- a/.github/workflows/check-man-pages.yml
+++ b/.github/workflows/check-man-pages.yml
@@ -1,0 +1,30 @@
+name: check man pages
+
+on:
+  pull_request:
+    paths:
+      - 'doc/manual/source/man/**'
+      - 'doc/manual/Makefile'
+
+jobs:
+  check-man-pages:
+    name: Check man pages are up-to-date
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
+
+      - name: Build man pages
+        working-directory: doc/manual/
+        run: make man
+
+      - name: Fail if build output differs
+        run: |
+          if ! git diff --quiet -- doc/manual/build/man; then
+            echo "::error::doc/manual/build/man is not up-to-date! Run 'make man' and commit the changed man pages."
+            git diff -- doc/manual/build/man
+            exit 1
+          fi
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -9,6 +9,3 @@ Provide your description here
   - [ ] Did you update the `man_pages` config in `doc/manual/source/conf.py`?
   - [ ] Did you update the packaged man pages in the `Cargo.toml`?
   - [ ] Did you commit the freshly built man pages?
-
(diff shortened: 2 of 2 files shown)


thgoebel:
You can see an example here: https://github.com/thgoebel/cascade/actions/runs/33070959056/job/98512775383

I also pushed a dummy commit that I will drop before we merge this. Once a maintainer approves the new workflow, we will see if its runs correctly in a pull_request context.

.github/workflows/check-man-pages.yml
```diff
@@ -0,0 +1,30 @@
...
+      - 'doc/manual/source/man/**'
+      - 'doc/manual/Makefile'
+
+jobs:
+  check-man-pages:
+    name: Check man pages are up-to-date
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
bal-e: Could you bump this to `@v7`?

.github/workflows/check-man-pages.yml
```diff
@@ -0,0 +1,30 @@
...
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
```
bal-e, line 18: I would prefer to pull in as little new code as possible, even though you've taken the nice step of locking the commit. Would `apt install uv` work? Even better: maybe a recent(-ish) version of `uv` is pre-installed on the CI runner, and this step can be omitted?

.github/workflows/check-man-pages.yml
```diff
@@ -0,0 +1,30 @@
+name: man pages
```
bal-e: Could you change this to "check man pages" to match the file name?

bal-e (changes requested):
This looks great! It's nice to trim down the number of simple, repetitive actions involved in contribution.

I have a few comments regarding the CI workflow, let me know what you think.

Also, I guess you'll have to rebase the PR on `main` now that #979 is merged.

.github/workflows/check-man-pages.yml
thgoebel: I didn't see uv being packaged in apt, but there is an [unofficial snap](https://snapcraft.io/astral-uv) by a Canonical employee.
That said, imho using https://github.com/astral-sh/setup-uv is fine, since it is maintained by Astral, the authors of uv.

bal-e: Oh, that's surprising. I assumed `uv` would be available everywhere. In that case, the current code looks good :)

bal-e:
Everything looks good @thgoebel. I looked at the CI check and noticed that some man pages beyond your test commit were outdated; I've updated them on `main`. Can you remove the test commit? GitHub will then run CI against the updated `main`, and we should see everything pass. I'll then approve and merge :)

thgoebel:
Done

bal-e (approved):

bal-e:
Thank you once again @thgoebel :)

Issue #982, opened 2026-08-27
Support KMIP client/server certificate authentication


thgoebel:
## Current behaviour

`cascade hsm add` has the [following options](https://cascade.docs.nlnetlabs.nl/en/latest/man/cascade-hsm.html#client-certificate-authentication): 

- `--client-cert <CLIENT_CERT_PATH>`
- `--client-key <CLIENT_KEY_PATH>`
- `--server-cert <SERVER_CERT_PATH>`
- `--ca-cert <CA_CERT_PATH>`

However, these options are stubs and don't do anything yet. They are all ignored: https://github.com/NLnetLabs/cascade/blob/v0.1.0-beta6/src/units/http_server.rs#L1547

## Expected behaviour

Implement storing and using these four values.

## Context

I was trying to test the [Securosys KMIP Server](https://docs.securosys.com/kmip/overview) for Primus HSM with Cascade. It only supports mTLS authentication, not username/password. It took some debugging into why the PEM files I provided weren't working, only to find out that Cascade's client cert CLI options don't do anything yet.

Would it be worth to add warnings to [the docs](https://cascade.docs.nlnetlabs.nl/en/latest/man/cascade-hsm.html#client-certificate-authentication) that these options are currently stubs?

## Version

- cascade v0.1.0-beta6

ximon18:
I suppose yes in the short term adding a warning could be a good option, though not ideal as the underlying dependency crates do already support certificate/key based authentication so it's just a matter of missing "wiring".

However, one of the reasons for not having this support yet is that it would need testing and for that we need something to test against. We in theory have access to a Kryptus kNet cloud HSM but haven't had time to pursue that yet, other than that we've only tested with the Fortanix free trial which expired, though we might be able to arrange a new one.

@thgoebel: If you have access to a KMIP capable HSM which requires these settings would you be able to help us test this support for it? In theory it should also be possible to do a sanity check using https://github.com/ximon18/keyls but that hasn't been updated in a long time and may not even currently build, but I've been meaning to return to that along with updating all of the KMIP related crates and projects we have.

[...]

PR #985, opened 2026-08-27
docs: Add warnings that KMIP client/server cert arguments are stubs

thgoebel:
See https://github.com/NLnetLabs/cascade/issues/982

---

### DIFF
--- a/doc/manual/build/man/cascade-hsm.1
+++ b/doc/manual/build/man/cascade-hsm.1
@@ -140,6 +140,14 @@ password set here will be used as the PKCS#11 PIN to login with.
 .B \-\-client\-cert <CLIENT_CERT_PATH>
 Optional path to a TLS certificate to authenticate to the KMIP server
 with. The file will be read and sent to the server.
+.sp
+\fBWARNING:\fP
+.INDENT 7.0
+.INDENT 3.5
+This argument \X'tty: link https://github.com/NLnetLabs/cascade/issues/982'\fI\%is not yet used\fP\X'tty: link'
+by Cascade.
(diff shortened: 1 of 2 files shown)


doc/manual/build/man/cascade-hsm.1
```diff
@@ -140,6 +140,14 @@ password set here will be used as the PKCS#11 PIN to login with.
 .B \-\-client\-cert <CLIENT_CERT_PATH>
 Optional path to a TLS certificate to authenticate to the KMIP server
 with. The file will be read and sent to the server.
+.sp
+\fBWARNING:\fP
+.INDENT 7.0
+.INDENT 3.5
+This argument \X'tty: link https://github.com/NLnetLabs/cascade/issues/982'\fI\%is currently\fP\X'tty: link'
+a stub and is not yet used by Cascade.
```
ximon18: I'm not sure that "stub" is something that will have meaning for an end user. If we just instead say here "is not yet used by Cascade" that would also deliver the same message but without using the term "stub". What do you think?

ximon18:
I think ideally we would also make the CLI actually reject attempts to use these arguments but that might actually not be much less work than just wiring them up to work. However https://github.com/ximon18/keyls/issues/13 shows that there is work to be done somewhere in our KMIP handling code and we'd also need to verify that it actually works so I think just documenting the lack is at least better than nothing.

bal-e (approved):
Thanks @thgoebel! @ximon18 is working on adding the missing functionality, so this warning will (hopefully) not be needed for long. But it's good to have until the new functionality is ready :)

